PR #56, opened 2026-06-03
feat(ast-lens)  — AST-Lens: outline-first reading for Gas City

jsgerman-oss:
Implementation of the read side of "The AST as LLM Lens": a polyglot AST outline emitter (Go/Python/TS/JS, ~94-98% token savings), a read-with-outline skill + discipline prompt fragment, and a Claude PreToolUse auto-prepend hook. 156 tests / 98% coverage.

Note: this is a code-bearing pack (bundles a Python emitter); flagging for maintainer review since existing packs are config-only. registry commit/hash are placeholders for the release tooling to finalize on merge.


julianknutsen:
Thanks for building this connected five-pack set, @jsgerman-oss. Your core PRs
helped build primitives these packs use, and this is exactly the kind of
ecosystem work we want Gas City to enable.

`gascity-packs` stays a small
first-party set, while community packs remain owned by their authors and are
published and discoverable through the Registry. That gives you the release
cadence while keeping the packs available to the community.

For `ast-lens`, the canonical repo already exists at
https://github.com/jsgerman-oss/ast-lens. Change its `[pack].name` to
`jsgerman-oss/ast-lens`, then follow
https://registry.gascity.com/publish: install the Registry GitHub App and use
**Find packs**. Every production release currently waits for staff review;
repeat releases are not promised automatic approval.

One concrete review note for the old monorepo branch: in
`test_fix_imports.py`, `test_relpath_is_repo_relative` hard-codes
`tests/write_fixtures/...`, but the git-root-relative result in this checkout
is `ast-lens/tests/write_fixtures/...`. The assertion is hidden when no Python
import formatter is installed because the test is skip-gated.

I'm closing this monorepo PR so `ast-lens` can move through owner publishing.
We'd like to see it in the Registry; if the path or that test finding is
unclear, reply here and we'll help.

PR #59, opened 2026-06-04
feat(provider-forge) Add provider-forge pack

jsgerman-oss:
Multi-provider model catalog + enablement: detect/verify provider CLIs, declare run targets, and generate model-advisor's tier roster from one source of truth. Codex + Claude live; gemini/copilot/cursor/kiro/omp/opencode/pi scaffolded.

### DIFF
--- a/provider-forge/.gitignore
+++ b/provider-forge/.gitignore
@@ -1,0 +1,14 @@
+.venv/
+__pycache__/
+*.pyc
+.pytest_cache/
+.ruff_cache/
+.coverage
+.coverage.*
+coverage.json
+coverage.xml
+htmlcov/
+
+# install.sh / uninstall.sh leave timestamped config backups next to the files
+# they edit; never commit those.
+*.provider-forge.bak.*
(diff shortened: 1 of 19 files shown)


julianknutsen:
Thanks for `provider-forge`, @jsgerman-oss. This connected pack set is a good
example of what owner-published Registry distribution is designed to support.

A provider/model catalog that tracks multiple vendors needs to release on its
maintainer's cadence. Your existing repo is the right ownership boundary:
https://github.com/jsgerman-oss/provider-forge. The Registry makes it
discoverable while preserving that cadence.

Change `[pack].name` there to `jsgerman-oss/provider-forge`, then follow
https://registry.gascity.com/publish. Install the Registry GitHub App and use
**Find packs** for default-branch `HEAD`; use the guide's GitHub Actions OIDC
workflow for tagged releases. Production currently sends every release to
staff review, so repository proof is not a promise of automatic approval.

This keeps `provider-forge` composable with `model-advisor` on your release
cadence. I'm closing the monorepo PR and would like to see both projects in
the Registry. If publishing snags, reply here and we'll help.

PR #62, opened 2026-06-05
feat: add the GasCity Cockpit enabler pack

jsgerman-oss:
## Summary

Adds `cockpit` as a top-level pack: the city-side half of the **GasCity Cockpit**, a VS Code extension that turns the editor into a live cockpit for fleets of coding agents. After this lands, a city becomes Cockpit-ready by importing one pack, with no gascity binary changes.

The expensive part already exists: the supervisor's versioned `/v0` HTTP API. This pack does **not** add a backend; it verifies a city is reachable and publishes the discovery descriptor the extension auto-reads.

## What the pack provides

```
cockpit/
  pack.toml                      name = "cockpit", schema 2
  bin/cockpit                    ready / discover / contract / adapter CLI
  cockpit/                       stdlib-only probe + contract-reader engine
  contract/v0.toml               the /v0 API contract the extension pins
  skills/cockpit-readiness/      verify + publish discovery
  skills/mayor-ide-affordances/  emit IDE-legible affordances
  install.sh / uninstall.sh      reversible, idempotent install lifecycle
  setup.sh / requirements.txt    optional venv (also runs under system python3)
  tests/                         pytest coverage
  docs/DESIGN.md                 cockpit-ready definition + discovery handshake
```

Stdlib-only (tomllib/tomli + urllib), no runtime third-party deps, mirroring provider-forge's footprint.

## Notes
- The extension lives at [jsgerman-oss/gascity-cockpit](https://github.com/jsgerman-oss/gascity-cockpit) ([site](https://jsgerman-oss.github.io/gascity-cockpit/)).
- `registry.toml` release `commit`/`hash` are pinned to the pack commit on this branch; happy to re-pin to the merge commit or regenerate with your release tooling.
- `validate_registry.py` passes locally.

### DIFF
--- a/cockpit/.gitignore
+++ b/cockpit/.gitignore
@@ -1,0 +1,14 @@
+.venv/
+__pycache__/
+*.pyc
+.pytest_cache/
+.ruff_cache/
+.coverage
+.coverage.*
+coverage.json
+coverage.xml
+htmlcov/
+
+# install.sh / uninstall.sh leave timestamped config backups next to the files
+# they edit; never commit those.
+*.cockpit.bak.*
--- a/cockpit/LICENSE
+++ b/cockpit/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2026 Jay German
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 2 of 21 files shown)


julianknutsen:
Thanks for `cockpit`, @jsgerman-oss. Here are the answers to the three
supervisor scope forks in `docs/DESIGN.md`:

1. **API enablement:** ratify your "assert and verify" interpretation. The
   supervisor `/v0` API is supervisor-wide and always on while the supervisor
   runs; the pack should not invent a per-city enable switch.
2. **extmsg adapter registration:** keep it runtime-owned by the extension or
   its host, which has the callback service and can re-register after restart.
   A static pack install should not pretend an ephemeral registration is
   durable. Persistent declarative registration, if we want it, is separate
   supervisor work.
3. **Discovery descriptor:** the pack-side probe/write is an acceptable v1
   fallback and schema contract. The supervisor is the correct eventual owner
   of writing and removing that descriptor at bind/shutdown; that belongs in
   an upstream SDK change, not hidden inside this pack.

Those answers establish the integration direction while you retain ownership
of the pack. Your newer canonical source is
https://github.com/jsgerman-oss/gascity-cockpit under `pack/`. Set
`[pack].name = "jsgerman-oss/cockpit"` and follow
https://registry.gascity.com/publish: install the Registry GitHub App, sign in,
and use **Find packs**. It supports a pack below the repo root. Production
currently sends every release to staff review.

These three supervisor answers stand as the integration direction. I'm
closing the monorepo PR and inviting `cockpit` into the Registry under your
ownership. If publication or an upstream SDK issue snags, reply here and
we'll help.

PR #64, opened 2026-06-07
fix(slack-pack): newest-first transcript lookup in find_latest_inbound_message_id_for_session

sjarmak:
## Problem

`find_latest_inbound_message_id_for_session` walked the `extmsg.transcript_list`
stream **oldest-first** and returned the first inbound match, so once a session's
transcript grew past the page limit the "latest inbound" lookup silently returned
a stale (oldest) message id instead of the newest one.

## Fix

Query the transcript **newest-first**, consuming the new ordering/limit params
added upstream in gastownhall/gascity#3128 (`limit` default 100 / max 500). The
first inbound match on a newest-first stream is the genuine latest inbound message.

## Test

Regression test pins the truncation case (transcript longer than one page) so the
oldest-first regression cannot reappear.

### DIFF
--- a/docs/design/slack-pack-tiering.md
+++ b/docs/design/slack-pack-tiering.md
@@ -1,0 +1,343 @@
+# slack-pack tiering: tier contracts, migration paths, and import compatibility
+
+Design memo for `gc-yrw` (slack-pack tiering). Tracking bead: `gc-yrw.1`.
+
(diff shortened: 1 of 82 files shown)


sjarmak:
Superseded by #150 — clean redo rebased on current main (this branch went stale/conflicting). Carrying the discussion forward there.

PR #66, opened 2026-06-08
Split GC build workflow into base and basic formulas

julianknutsen:
## Summary
- add `build-base` as the non-catalog build workflow contract/skeleton
- add cataloged `build-basic` as the concrete default implementation
- keep `build-run` as a compatibility alias that extends `build-basic`
- update issue-fix build docs, mayor examples, and override docs to use the new names

## Tests
- `python3 -m pytest gascity/tests -q`


julianknutsen:
Closing for now while we refine the build-base/build-basic contract. The current split only covers the back-half build loop, but the intended abstraction should cover planning/design through clean review so methodology packs like Compound Engineering and Superpowers can implement the full lifecycle.

PR #16, opened 2026-05-16
Fix tmux theme status badges

jerebrow-orcl:
## Summary
Make the tmux-theme status helper fast and reliable by replacing slow status-bar calls to `gc hook` and `gc mail check` with bounded read-only Beads counts. Also make prefix keybinding setup idempotent for the pack's run-shell bindings.

## What changed
- `status-line.sh` now uses bounded `bd list --json --limit 0` queries for open nudges and messages.
- The helper exits 0 even when there are no badges or count queries fail.
- `tmux-theme.sh` passes the active city path to the status helper so tmux `#()`
  refreshes do not depend on the tmux server's cwd.
- `bind-key.sh` detects existing theme run-shell bindings instead of looking only for a literal `gc ` command.
- Added focused shell tests for badge rendering, fallback behavior, and idempotent binding setup.
- Updated README prerequisites/status wording.

## Validation
- `bash -n tmux-theme/scripts/status-line.sh tmux-theme/scripts/bind-key.sh tmux-theme/scripts/tmux-theme.sh tmux-theme/tests/test_tmux_theme_scripts.sh`
- `bash tmux-theme/tests/test_tmux_theme_scripts.sh`
- `git diff --check`

## Context
This avoids requiring downstream bootstrap scripts to rewrite the imported tmux-theme pack after materialization.

PR #70, opened 2026-06-08
fix(gastown/statusline): cache work/mail counts to stop per-render fork storm

Wldc4rd:
## What

Cache the gastown tmux statusline's work/mail counts with a short TTL so it stops forking `gc hook` + `gc mail check` on **every** render.

## Why

`status-line.sh` (tmux `status-right`, run every `status-interval` for each agent) calls `gc hook` and `gc mail check` per render. Each is a `gc → bd.real → dolt` cascade. Measured with `strace -f -e execve`, **one render = 888 process execs** (gc=2 → bd.real=43, dolt=34, + the rest of the cascade). Across agents rendering continuously, this is a large, avoidable share of city fork/CPU load — and high fork load is exactly what operators misread as CPU saturation (load ≠ CPU).

Field context: on a busy city this statusline path was a double-digit % of total forks/s (bpftrace `sched_process_fork` ancestry: `gc ← status-line.sh` ≈ 271–316 forks/10s).

## Change

Cache the two counts with a TTL (default 30s, override `GC_STATUSLINE_TTL`). Within the TTL, renders reuse the cached value via a `date`+`stat` freshness check (no gc); the gc calls run at most once per TTL per agent. Output is byte-identical.

Freshness uses `date`+`stat`, **not** `find -newermt`/`-mmin` — some `find` variants (e.g. **bfs**) reject those, which silently defeats the cache (every render misses). Found via dogfooding.

## Proof (strace -f -e execve, one render)

| render | total execs | gc | bd.real | dolt |
|---|---|---|---|---|
| before (current) | 888 | 2 | 43 | 34 |
| cached — miss | 891 | 2 | 43 | 34 |
| **cached — hit** | **4** | **0** | **0** | **0** |

**99.5% fewer execs per cached render**, validated in an environment where `find` is `bfs`. Counts are at most `GC_STATUSLINE_TTL` seconds stale — fine for a status indicator.

## Scope

Display-script only; no data/control-plane change. Complements the `gc doctor` fork-rate watch (gascity#3252), which can confirm the city-wide drop once this is imported.

### DIFF
--- a/gastown/assets/scripts/status-line.sh
+++ b/gastown/assets/scripts/status-line.sh
@@ -3,17 +3,33 @@
 # Usage: status-line.sh <agent-name>
 # Called by tmux every status-interval seconds via #(command).
 # Always exits 0 — tmux must never see errors.
+#
+# Counts are cached with a short TTL so tmux's frequent refresh does not fork
+# `gc hook` + `gc mail check` (each a gc->bd.real->dolt cascade) on EVERY render
+# across EVERY agent — a significant, avoidable share of city fork/CPU load.
+# Within the TTL the cached value is reused; the gc calls run at most once per
+# TTL per agent. TTL override: GC_STATUSLINE_TTL (seconds).
 
 agent="$1"
 [ -z "$agent" ] && exit 0
 
-# Count pending work items (non-empty lines from gc hook).
-w=$(gc hook "$agent" 2>/dev/null | grep -c . || true)
+cache_ttl="${GC_STATUSLINE_TTL:-30}"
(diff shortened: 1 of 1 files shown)


julianknutsen:
Thanks for the measurements and the clear fork-storm investigation. I opened #71 as the current-tree synthesis that supersedes this PR: it keeps the short TTL cache idea from this PR and combines it with bounded read-only `bd list` queries plus city-path wiring from #16.

Appreciate you digging into the statusline load issue and proving the cache direction.

julianknutsen:
Superseded by #71

PR #76, opened 2026-06-10
fix(gastown): refinery direct-merge — refspec push + verified MERGED_SHA

reggieperry:
## Problem

In the refinery patrol's `MERGE_STRATEGY = "direct"` path, when `$TARGET` is held by another worktree in the repo, `git checkout $TARGET` fails with "already used by worktree". An agent that continues past the failure gets a silently no-op'd push, and `MERGED_SHA=$(git rev-parse HEAD)` then records a fabricated stale SHA in the closed bead's merge metadata — a false "merged" whose forensic breadcrumbs point at a commit that never reached origin.

## Fix

- Push via refspec (`git push origin temp:$TARGET`) — no local checkout of the target branch required.
- Capture PRE/POST `origin/$TARGET` SHAs and verify origin actually advanced to the branch tip; halt loud (`gc runtime drain-ack; exit 1`) on any mismatch, matching the formula's existing FATAL idiom.
- Derive `MERGED_SHA`/`MERGED_SHORT` from the verified origin tip, never from `git rev-parse HEAD`.

## Provenance and adaptation

Adapted from the orphaned sibling commit on gastownhall/gascity#2610's branch (`3bf3d222a`, bead ga-jzb0i), which predated this formula's migration here. One deliberate change versus that patch: its ancestor/no-op-merge rejection block is **dropped** — the step-0 false-completion guard that has since landed (`branch_has_real_change` / `halt_false_completion`) already refuses a branch whose diff against its merge-base is empty, which includes the already-an-ancestor case, so the check would be dead code at step 1.

Note for maintainers: the embedded copy at `gastownhall/gascity` `examples/gastown/packs/gastown/formulas/mol-refinery-patrol.toml` (currently byte-identical) carries the same bug and will want the same change on the next sync.

## Validation

- `gastown/formulas/mol-refinery-patrol.toml` parses (tomllib).
- `validate_registry.py`: ok.
- The CI pytest set (389 tests) and `gastown/tests/test_gastown_theme_scripts.sh` pass locally.

### DIFF
--- a/gastown/formulas/mol-refinery-patrol.toml
+++ b/gastown/formulas/mol-refinery-patrol.toml
@@ -688,23 +688,51 @@ branch_has_real_change "origin/$TARGET" temp || \
   halt_false_completion "$BRANCH" "$(git merge-base "origin/$TARGET" temp 2>/dev/null || printf '%s' "origin/$TARGET")"
 ```
 
-**1. Merge and push target branch:**
-```bash
-git checkout $TARGET
-git merge --ff-only temp
-git push origin $TARGET
-```
+**1. Push via refspec and verify origin moved:**
 
-**2. Verify push:**
+No local checkout of `$TARGET` — `git checkout $TARGET` fails with
+"already used by worktree" when another worktree in the repo holds the
+target branch, and a run that continues past that failure fabricates its
+merge metadata from a stale local HEAD. The refspec push needs no local
+checkout, and the PRE/POST capture proves origin actually moved. (A
+branch that is already an ancestor of `$TARGET` never reaches this step:
+its diff against the merge-base is empty, so step 0's
+`branch_has_real_change` guard halts it.)
+
 ```bash
+git fetch --prune origin
+PRE_MERGE_SHA=$(git rev-parse origin/$TARGET)
+BRANCH_TIP=$(git rev-parse temp)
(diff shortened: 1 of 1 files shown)


aphexcx:
From the 2026-08-01 sweep of open PRs: this appears to be **superseded by `main`**. `gastown/formulas/mol-refinery-patrol.toml` now carries the hardened direct-merge push flow this PR was after, landed via 4212acb ("fix(gastown): make refinery direct merge worktree-safe"):

- **Explicit-refspec fetches**: `git fetch origin "+refs/heads/${BRANCH}:refs/remotes/origin/${BRANCH}" "+refs/heads/${TARGET}:refs/remotes/origin/${TARGET}"` — with a hard STOP if the fetch fails (lines ~705–706, and again before/after the push at ~785/793).
- **Verified push**: `MERGED_SHA=$(git -C "$MERGE_WT" rev-parse HEAD)` → push → re-fetch → `[ "$MERGED_SHA" != "$REMOTE" ]` → "PUSH FAILED … STOP. Do not mutate bead state." (~789–796), with `merged_sha` recorded in bead metadata.
- `gastown/tests/test_gastown_pack_assets.sh` now enforces the presence of the verify block.

Recommend closing this PR as superseded — flagging for the author/maintainers rather than closing unilaterally.

reggieperry:
If it's superseded, then I'm ok to close it. I need to update things to test but if I find anything I will report it.

PR #88, opened 2026-06-12
feat(slack-full): transport-level reaction protocol — 👀 enqueue ack, ⚠️ dead-dispatch, ✍️ agent processing

sjarmak:
Humans messaging agents through the adapter previously got no signal until the agent itself got around to reacting — silence was indistinguishable from breakage, and failed dispatches (dead session, stale binding) were entirely invisible in-channel.

- 👀 reacts at transport level the moment an inbound is successfully enqueued to a target session
- ⚠️ reacts when dispatch fails (dead session / no live binding / nudge failure), with the failure logged — silent drops become visible in-channel
- agent-side "I am processing" react moves to ✍️ so *queued* vs *being worked* are distinguishable; bind-time protocol reminder text updated to match

Two new tests; `go test -race ./...` clean. Running in production on our city since 2026-06-12.

### DIFF
--- a/docs/design/slack-pack-tiering.md
+++ b/docs/design/slack-pack-tiering.md
@@ -1,0 +1,343 @@
+# slack-pack tiering: tier contracts, migration paths, and import compatibility
+
+Design memo for `gc-yrw` (slack-pack tiering). Tracking bead: `gc-yrw.1`.
+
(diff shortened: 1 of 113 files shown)


sjarmak:
Superseded by #151 — salvaged onto fresh main (this branch went stale/conflicting). Carrying forward there.

PR #94, opened 2026-06-14
fix(witness): parse gc session list --json .sessions[] for orphan recovery (v1.1.1)

sarendipitee:
# fix(witness): parse gc session list --json .sessions[] for orphan recovery (v1.1.1)

## What broke and why

`gc session list --json` changed shape in gc v1.1.1. The output went from a
flat top-level array:

```json
[{"id": "abc", "state": "active", ...}, ...]
```

to an envelope object:

```json
{
  "filters": {...},
  "ok": true,
  "schema_version": 1,
  "sessions": [{"id": "abc", "state": "active", ...}, ...],
  "summary": {...}
}
```

Any consumer that iterated the top-level value as an array silently received
booleans and objects instead of session records. The `jq` reduce loop
produced an empty assignee→state map every cycle.

## Silent failure mode

The witness's orphan recovery step builds a liveness map once per patrol
cycle, then resolves each bead's assignee against it. An empty map causes
every assignee to resolve to `"absent"`, which would false-orphan every
assigned bead. The existing empty-map fail-safe (`MAP_COUNT == 0 &&
SESSION_COUNT > 0`) caught this and aborted recovery — so no beads were
incorrectly recovered. However, orphan recovery was effectively disabled
town-wide every cycle for the duration of any gc v1.1.1+ deployment.

## The fix

Normalize the session-list result with `.sessions? // .` before iterating:

```jq
(reduce (($sessions[0].sessions? // $sessions[0]) // [])[] as $s (...))
```

- On gc v1.1.1+: `.sessions?` extracts the array under `.sessions`.
- On older gc versions: `.sessions?` is absent, so `// $sessions[0]` falls
  back to the whole value, which is already the flat array.
- The `// []` guard handles a null/missing result without erroring.

The same guard is applied in `orphan-sweep.sh` and `nudge-on-route.sh`.

Field names are also corrected from PascalCase (`ID`, `State`, `Closed`,
`SessionName`, `Alias`, `AgentName`) to the snake_case the API actually
returns (`id`, `state`, `closed`, `session_name`, `alias`, `agent_name`).
The shell scripts (`orphan-sweep.sh`) retain PascalCase as a fallback via
a `pick()` helper that tries the canonical name first — this guards against
future gc releases reverting the casing, not the current gc v1.1.1 output.

Additional changes in this commit:
- Switch from `--argjson` on argv to `--slurpfile` with temp files to avoid
  `ARG_MAX` overflows on busy cities where session `command` fields are large.
- Build the liveness map once per cycle (not once per bead) and resolve via
  `jq -r --arg a "$ASSIGNEE" '.[$a] // "absent"'`.
- Improve diagnostic output: log session count vs map entry count so schema
  drift is immediately visible in witness logs.

## Backward compatibility

The `.sessions? // .` guard is safe on all gc versions:
- gc pre-v1.1.1: `.sessions?` evaluates to `null` (key absent), `// .`
  selects the original flat array. No behavior change.
- gc v1.1.1+: `.sessions?` extracts the sessions array. Fixed behavior.

No configuration changes required. No other pack files affected.


sarendipitee:
Rebased onto current main and **stacked on #92** (fix/gastown-vendor-maintenance-standalone). This PR modifies `maintenance/assets/scripts/{orphan-sweep,nudge-on-route}.sh`, which are vendored by #92 — so #94 depends on #92. Base retargeted to #92's branch; the diff here now shows only #94's own change. Once #92 merges, GitHub will auto-retarget this to main.

sarendipitee:
**Correction to previous comment:** cross-repo PR bases must be branches in the upstream repo, so this PR cannot be stacked on #92's fork branch. Base stays on `main`.

**Dependency:** #94 modifies `maintenance/assets/scripts/{orphan-sweep,nudge-on-route}.sh`, which are vendored by **#92**. Those files don't exist on `main` yet, so #94's branch currently includes #92's commit on top of main (that's why the diff is large). Marking this **draft** until #92 merges; once it does, I'll rebase #94 onto main so it shows only its own 3-file change.

PR #95, opened 2026-06-14
feat(maintenance): rig-agent liveness + merge-freshness watchdog order

sarendipitee:
# feat(maintenance): rig-agent liveness + merge-freshness watchdog order

## Problem

Witness patrol loops die silently. When a `witness` or `refinery` session
crashes, enters a `failed-create` state, or simply stops responding, nothing
fires an alert. The Witness LLM loop is responsible for detecting stalls — but
when the loop itself is the thing that died, there is no observer left to notice.
Operators find out hours later when they inspect a rig manually or notice no
commits have landed on the default branch.

Two failure classes go undetected today:

1. **Dead agent**: a `refinery` or `witness` session enters a hard-dead state
   (`crashed`, `failed-create`, `closed`, absent) and does not self-recover.
2. **Merge stall**: the `refinery` session is technically alive but is not
   draining its merge queue — the default branch has received no commit within
   the configured freshness window and there is work outstanding.

## Why a shell-based watchdog

The Witness LLM patrol loop is exactly the component that silently fails. Using
another agent or wisp to watch it introduces the same single point of failure in
a different costume. A deterministic shell exec order — one with no LLM, no
agent, no wisp — survives where an agent loop does not. It runs on a fixed
`cooldown` interval driven by the controller, which continues ticking regardless
of whether any user-configured agent role is alive.

This design follows the SDK's self-sufficiency invariant: the watchdog is a
plain exec order and makes no assumption about which role names exist in the
deploying city.

## What the watchdog checks

Every 3 minutes the order calls `maintenance/assets/scripts/rig-liveness-watchdog.sh`,
which:

1. Enumerates every running, non-suspended, non-HQ rig via `gc rig list --json`.
2. For each rig pulls session state via `gc --rig <name> session list --json`,
   normalizing across the current v1.1.1 `{sessions:[...]}` object shape and the
   legacy flat-array format.
3. **Liveness check**: a session is DEAD if its state is `absent` (no session
   record at all), `crashed`, `creating`, `failed-create`, `closed`, or empty.
   States `asleep`, `active`, `awake`, and `running` are all ALIVE (an idle
   refinery legitimately sleeps).
4. **Merge-freshness check**: if the refinery is alive but the queue is non-empty
   AND the last commit on the rig's default branch is older than
   `WATCHDOG_FRESH_MIN` minutes (default 60), that is a STALL.
5. On any dead agent or stall, mails a loud `[INCIDENT]` escalation with
   concrete evidence (state, queue depth, last-merge age, default branch) to the
   configured recipient.

## Key design choices

**JSON ledger de-duplication.** A JSON file at
`$GC_PACK_STATE_DIR/rig-liveness-incidents.json` tracks a tick count per
`rig:agent:symptom-class` key. An active incident only re-alerts every
`WATCHDOG_REALERT_TICKS` ticks (default 5, i.e. every 15 minutes). When an
incident clears — the symptom is no longer detected — its key is pruned from the
ledger, so it will alert immediately if it recurs. This eliminates per-tick spam
without silencing newly triggered or newly changed incidents.

**Symptom-keyed, not time-keyed.** The ledger key includes the symptom class
(`dead` vs `stall`), so if a refinery recovers from dead but then enters a
merge stall, that is treated as a new incident and alerts immediately regardless
of the tick count.

**Session-API shape normalization.** The `(.sessions? // .) | if type == "array" then . else [] end`
jq expression handles both the current object-wrapped shape and the older
top-level array in one line, with no branching in the shell layer. This means
the watchdog works across city deployments that have not yet upgraded.

**Configurable via environment variables** set on the controller — no
city-specific file edits required:

| Variable | Default | Meaning |
|---|---|---|
| `WATCHDOG_ESCALATE_TO` | `mayor` | Recipient for incident mail |
| `WATCHDOG_FRESH_MIN` | `60` | Merge freshness window (minutes) |
| `WATCHDOG_REALERT_TICKS` | `5` | Re-alert cadence for a live incident |

## What incidents this would have caught

- **Silent witness crash**: witness enters `crashed` state after a tmux
  reconnect; LLM patrol loop is gone; no bead is ever filed; no human notices
  until a PR review goes stale for a day.
- **Refinery merge stall with alive session**: refinery session is `asleep`
  (healthy-looking) but is blocked on a failing gate check with 8 beads in
  queue; default branch receives no commits for 4 hours; no alert fires.
- **Failed-create on rig restart**: operator restarts a rig; the refinery
  session hits a `failed-create` and the rig is left with no merge capacity;
  the witness is alive but cannot itself restart the refinery.

## Prerequisites

The watchdog uses `declare -A` (bash associative arrays), which requires bash 4
or later. Stock macOS ships bash 3.2; the script guards this at startup and
exits with a clear error if the version is too old. Deployers on macOS need
Homebrew bash (`brew install bash`) or any Linux distribution (bash 4+ ships
everywhere by default).

## Files changed

- `maintenance/assets/scripts/rig-liveness-watchdog.sh` — 214-line shell
  watchdog (new file, executable)
- `maintenance/orders/rig-liveness-watchdog.toml` — exec order wiring
  (new file, 3-minute cooldown interval)

### DIFF
--- a/maintenance/assets/scripts/rig-liveness-watchdog.sh
+++ b/maintenance/assets/scripts/rig-liveness-watchdog.sh
@@ -1,0 +1,214 @@
+#!/usr/bin/env bash
+# rig-liveness-watchdog — deterministic per-rig agent + merge-freshness watchdog.
+#
+# Catches the failure where a critical rig agent (refinery / witness) silently
+# dies, or the refinery stalls with a non-empty merge queue, with NO alert.
+# This is a plain exec order (no LLM, no agent, no wisp) precisely BECAUSE the
+# Witness LLM patrol loop is the thing that silently dies — a deterministic
+# shell tick survives where an agent loop does not.
+#
+# For every non-suspended, running rig in the city it checks:
+#
+#   1. The rig's `refinery` and `witness` sessions are alive. A session is
+#      DEAD when it is missing entirely or its state is one of the hard-dead
+#      states (crashed / creating / failed-create / closed). `asleep`,
+#      `active`, `awake`, `running` are all ALIVE (an idle refinery sleeps).
+#
+#   2. Merge freshness: the refinery has merged to the rig's default branch
+#      within $WATCHDOG_FRESH_MIN minutes (last commit time on the default
+#      branch), OR its merge queue is genuinely empty (no open/in_progress
+#      beads assigned to the refinery and no ready in_progress work). If the
+#      queue is NON-empty AND there has been no merge inside the window, that
+#      is a STALL.
+#
+# On any detected dead-agent or stall it mails a LOUD [INCIDENT] escalation to
+# $WATCHDOG_ESCALATE_TO (default: mayor) with concrete evidence. A JSON ledger
+# de-dups: the same incident only re-alerts every $WATCHDOG_REALERT_TICKS ticks
+# or when its symptom changes, so a persistent stall does not spam every tick.
+#
+# Config (all env-overridable, upstream-friendly defaults):
+#   WATCHDOG_ESCALATE_TO   recipient for incident mail        (default: mayor)
+#   WATCHDOG_FRESH_MIN     merge-freshness window, minutes    (default: 60)
+#   WATCHDOG_REALERT_TICKS re-alert cadence for a live incident (default: 5)
+set -euo pipefail
+
+# Trace bd/gc invocations to $GC_BD_TRACE_JSON when set (no-op otherwise).
+__SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+# shellcheck disable=SC1091
+. "$__SCRIPT_DIR/_bd_trace.sh" "rig-liveness-watchdog"
+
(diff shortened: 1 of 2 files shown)


julianknutsen:
Thanks for documenting the liveness incident in this much detail,
@sarendipitee. We traced the history to place each behavior with its current
owner: `maintenance` was removed from this repo in `d3617d1`, and its pack
responsibilities were folded into the SDK's bundled `core` pack in gascity
`f895c0ff4`.

Against the current system, dead-session restart belongs to the session
reconciler; refinery/witness stall judgment belongs to gastown's existing
`mol-deacon-patrol` health and queue-starvation checks; escalation goes
through core's escalation hook rather than raw mail to a hard-coded `mayor`.
The one uncovered signal from the series—the silent witness-heartbeat
stall—was adapted from #99 with contributor credit and merged as #237.
However, its deterministic pass is not operational yet: open #243 fixes a
managed-session path bug that currently bypasses the checker and falls back
to LLM judgment.

The broad watchdog's responsibilities now sit with those focused owners. I'm
closing this original maintenance-path PR with the fold-in history recorded
and the useful heartbeat signal carried forward through #237/#243.

PR #97, opened 2026-06-14
fix(maintenance): per-session cooldown for nudge-on-mail to stop spam

sarendipitee:
# fix(maintenance): per-session cooldown for nudge-on-mail to stop spam

## Problem

`nudge-on-mail` wakes a session the moment a `type=message` bead is delivered
to it. Each bead triggers one nudge. Under normal operation — one mail, one
nudge — that's the right behaviour.

Under a watchdog incident the system emits a burst: an incident bead, one or
more advisory digests, and any related agent notifications, all delivered in
the same polling window. Before this fix every bead in that burst fired an
independent nudge, causing the recipient session to receive several rapid-fire
interruptions for the same underlying event. On a busy city with multiple
watchdog checks this produced enough noise that agents began dropping or
mishandling the extra wakes.

## Fix

Add a **per-assignee cooldown** gated by `GC_NUDGE_ON_MAIL_COOLDOWN`
(default `10m`).

When a new (unseen) mail bead arrives for an assignee:

1. The bead is recorded in state immediately so it is not re-evaluated on
   the next tick regardless of what happens to the nudge.
2. A cooldown check tests whether a `session:<assignee>` entry exists in
   state and whether fewer than `COOLDOWN` seconds have elapsed since the
   last nudge to that assignee.
3. If the assignee is still inside the cooldown window, the nudge is
   skipped. All beads in the burst are still marked seen, so none of them
   will generate a nudge on future ticks either.
4. If the assignee is outside the cooldown window (or has never been
   nudged), the nudge fires and the `session:<assignee>` timestamp is
   updated.

The result: a burst of ten mail beads collapses into a single wake. The
agent is notified once and then processes its full inbox — the correct
behaviour given GUPP ("if you find work on your hook, you run it").

## Configuration

| Variable | Default | Meaning |
|---|---|---|
| `GC_NUDGE_ON_MAIL_COOLDOWN` | `10m` | Minimum gap between nudges to the same assignee |

The existing `GC_NUDGE_ON_MAIL_LOOKBACK` and `GC_NUDGE_ON_MAIL_RETENTION`
variables are unchanged. The cooldown state is namespaced `session:<assignee>`
in the same state file as bead-seen entries and is pruned by the same
retention sweep.

## Implementation notes

- `duration_to_seconds` already existed; `COOLDOWN_S` uses it the same way
  `RETENTION_S` does.
- The `fromdateiso8601` + `now` pattern is already used for the retention
  prune — cooldown reuses the same jq clock, so there is no new time-handling
  logic.
- The bead-seen entry is written **before** the nudge attempt so a transient
  nudge failure does not cause the same bead to re-evaluate the cooldown on
  the next tick. The `session:<assignee>` cooldown entry is written **only on
  a successful nudge**: a persistent nudge failure (session down) will not
  set a cooldown, so the next new bead for that assignee will retry — which
  is the correct behaviour for a downed session.
- `_bd_trace.sh` is sourced only if present (`[ -f ... ] && . ...`). It is
  optional instrumentation; the script functions fully without it.
- **Invariant:** `RETENTION` must be >= `COOLDOWN` or session cooldown entries
  may be pruned before the window expires. The script now emits a warning to
  stderr at startup if this is violated. With defaults (RETENTION=1h,
  COOLDOWN=10m) the invariant holds comfortably.
- `RETENTION_S` is now computed before the main loop (alongside `COOLDOWN_S`)
  so the invariant check fires before any state is mutated.

## Testing

Deploy against a city with a watchdog formula. Trigger a watchdog incident
(or manually create multiple mail beads for the same assignee within 10
minutes). Confirm the recipient session receives exactly one nudge per
`GC_NUDGE_ON_MAIL_COOLDOWN` window rather than one per bead.

### DIFF
--- a/maintenance/assets/scripts/nudge-on-mail.sh
+++ b/maintenance/assets/scripts/nudge-on-mail.sh
@@ -7,14 +7,21 @@
 #
 # Subscribes to bead.created events; whenever a bead with issue_type=message
 # arrives it nudges the assignee session. Idempotent: a given bead_id is nudged
-# at most once. Dedup state lives in $GC_PACK_STATE_DIR/nudge-on-mail-state.json.
+# at most once per run; an assignee is nudged at most once per COOLDOWN window
+# across runs. Dedup state lives in $GC_PACK_STATE_DIR/nudge-on-mail-state.json.
+#
+# Note: concurrent runs are not serialised. Under parallel execution two
+# instances may both read state before either writes, causing a duplicate nudge
+# for the first bead a given assignee receives in that window. The controller's
+# exec-order serialisation should prevent this in normal operation.
 #
 # Runs as an exec order (no LLM, no agent, no wisp).
 set -euo pipefail
 
 __SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+# _bd_trace.sh is optional instrumentation; source it only if present.
 # shellcheck disable=SC1091
-. "$__SCRIPT_DIR/_bd_trace.sh" "nudge-on-mail"
+[ -f "$__SCRIPT_DIR/_bd_trace.sh" ] && . "$__SCRIPT_DIR/_bd_trace.sh" "nudge-on-mail"
 
 if ! command -v jq >/dev/null 2>&1; then
     echo "nudge-on-mail: jq is required but not found in PATH" >&2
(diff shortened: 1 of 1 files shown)


julianknutsen:
Thanks for chasing the mail-nudge behavior, @sarendipitee. We assessed it
against the post-`maintenance` architecture:
`maintenance` was removed in `d3617d1` and folded into the SDK's bundled
`core` pack in gascity `f895c0ff4`.

Mail wakeups are now sender-side and native:
`gc mail send --notify`/`--nudge` performs live nudge, managed wake-and-poke,
or queued delivery and records `last_nudge_delivered_at`. This native path
handles delivery at the source and avoids re-deriving mail from event polling.

The behavior now has a native owner in mail delivery. I'm closing the original
maintenance-path PR with that current integration path recorded.

PR #99, opened 2026-06-14
feat(maintenance): watchdog catches asleep-too-long (stalled) witness loop

sarendipitee:
# feat(maintenance): watchdog catches asleep-too-long (stalled) witness loop

## The Gap

The rig-liveness watchdog already caught **dead** sessions — states like
`crashed`, `failed-create`, or `closed`. But there is a second, more
insidious failure mode: a witness that is in an `asleep` or `active` state
whose **self-scheduled patrol loop has silently died**.

The witness's ScheduleWakeup loop is responsible for periodically waking the
agent to check for pending work. When this loop dies, the witness sits
`asleep` indefinitely — its session state looks healthy but it is patrolling
nothing. The existing dead-state check cannot see this because the session is
not dead.

## Evidence

This failure was observed in production. Specific patrol stalls of **14 hours
to 63 hours** were recorded, all while the witness session reported an
`asleep` or `active` state throughout. Work accumulated silently; no alert
fired.

A healthy witness wakes approximately every 60 seconds between patrol cycles.
Fifteen minutes of silence is therefore a clear signal that the loop is gone —
not a transient sleep. The threshold is configurable via
`WATCHDOG_WITNESS_STALE_MIN` (default: 15 minutes).

## The Fix

**Heartbeat freshness check (Check 1b)** added to `rig-liveness-watchdog.sh`:

After the existing dead-state check passes (witness is alive), the watchdog
now reads the witness session's `last_active` and `last_nudge_delivered_at`
fields and takes the newer of the two as the heartbeat. The
`last_nudge_delivered_at` field is particularly useful here because the patrol
loop's self-nudge mechanism keeps it fresh every cycle.

If the heartbeat is more than `WATCHDOG_WITNESS_STALE_MIN` minutes old — and
the field is not the Go zero-time sentinel (see below) — the watchdog emits a
`:witness:stalled` incident.

The new incident is a **distinct class** from `:witness:dead`. This matters
because the de-duplication ledger tracks incidents by key
(`rig:agent:symptom-class`). A session could transition from stalled to dead
in subsequent ticks; keeping these as separate keys prevents de-dup logic from
treating one as a continuation of the other and suppressing the re-alert.

## Go Zero-Time Sentinel Handling

`gc` emits `0001-01-01T00:00:00Z` for any session field that has not been set
yet (the Go zero value for `time.Time`). The new `ts_epoch()` helper treats
this value — and any timestamp beginning with `0001-` — as "no usable signal"
(returns `0`), never as "ancient". A zero epoch skips the staleness check
entirely, avoiding false positives on newly-started witnesses that have not yet
recorded a heartbeat.

## Why Refinery Staleness Was Not Added

Refinery heartbeat freshness was deliberately **not** added:

1. When a merge queue is non-empty, the refinery merge-freshness check (Check
   2) already catches a stalled refinery.
2. When the merge queue is empty, a non-nudged refinery is legitimately idle —
   a heartbeat-age check would false-alert.

Adding refinery staleness would duplicate the merge-stall check on the busy
path and add false positives on the idle path. No net gain.

## Configuration

All thresholds are env-overridable. `WATCHDOG_ESCALATE_TO` is **required** —
it has no default and the script exits with an error if unset. Set it in your
pack config to the agent role that should receive incident mail (e.g. your
overseer). This keeps the script free of hardcoded role names.

| Variable | Default | Description |
|---|---|---|
| `WATCHDOG_WITNESS_STALE_MIN` | `15` | Minutes of silence before a live witness is flagged stalled |
| `WATCHDOG_ESCALATE_TO` | *(required — no default)* | Recipient for `[INCIDENT]` mail; must be set by consumer pack config |
| `WATCHDOG_FRESH_MIN` | `60` | Refinery merge-freshness window |
| `WATCHDOG_REALERT_TICKS` | `5` | Re-alert cadence per persistent incident |

## Portability

`ts_epoch()` tries GNU `date -d` first, then BSD `date -j -f` as a fallback,
so the stall-detection logic works correctly on both Linux and macOS/BSD
systems.

## Incident Format

Emitted as a `[INCIDENT]` mail to `$WATCHDOG_ESCALATE_TO` with full evidence:
rig name, witness state, last heartbeat age, and suggested remediation action.
The de-dup ledger ensures a persistent stall does not spam every tick.

### DIFF
--- a/PR_DESCRIPTION.md
+++ b/PR_DESCRIPTION.md
@@ -1,0 +1,93 @@
+# feat(maintenance): watchdog catches asleep-too-long (stalled) witness loop
+
+## The Gap
+
+The rig-liveness watchdog already caught **dead** sessions — states like
+`crashed`, `failed-create`, or `closed`. But there is a second, more
+insidious failure mode: a witness that is in an `asleep` or `active` state
+whose **self-scheduled patrol loop has silently died**.
+
+The witness's ScheduleWakeup loop is responsible for periodically waking the
+agent to check for pending work. When this loop dies, the witness sits
+`asleep` indefinitely — its session state looks healthy but it is patrolling
+nothing. The existing dead-state check cannot see this because the session is
+not dead.
+
+## Evidence
+
+This failure was observed in production. Specific patrol stalls of **14 hours
+to 63 hours** were recorded, all while the witness session reported an
+`asleep` or `active` state throughout. Work accumulated silently; no alert
+fired.
+
+A healthy witness wakes approximately every 60 seconds between patrol cycles.
+Fifteen minutes of silence is therefore a clear signal that the loop is gone —
+not a transient sleep. The threshold is configurable via
+`WATCHDOG_WITNESS_STALE_MIN` (default: 15 minutes).
+
+## The Fix
+
+**Heartbeat freshness check (Check 1b)** added to `rig-liveness-watchdog.sh`:
+
+After the existing dead-state check passes (witness is alive), the watchdog
+now reads the witness session's `last_active` and `last_nudge_delivered_at`
+fields and takes the newer of the two as the heartbeat. The
+`last_nudge_delivered_at` field is particularly useful here because the patrol
+loop's self-nudge mechanism keeps it fresh every cycle.
+
+If the heartbeat is more than `WATCHDOG_WITNESS_STALE_MIN` minutes old — and
+the field is not the Go zero-time sentinel (see below) — the watchdog emits a
+`:witness:stalled` incident.
+
+The new incident is a **distinct class** from `:witness:dead`. This matters
+because the de-duplication ledger tracks incidents by key
+(`rig:agent:symptom-class`). A session could transition from stalled to dead
+in subsequent ticks; keeping these as separate keys prevents de-dup logic from
+treating one as a continuation of the other and suppressing the re-alert.
+
+## Go Zero-Time Sentinel Handling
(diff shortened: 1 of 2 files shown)


julianknutsen:
Thank you for the key idea in this PR, @sarendipitee. We confirmed that a
witness can retain a healthy-looking session state
after its patrol loop silently dies, and that gastown's existing checks had
reasoned away the empty-queue case twice.

We adopted that signal with credit to this PR in
https://github.com/gastownhall/gascity-packs/pull/237, merged as `ed8c070`.
The adopted form is a read-only first pass inside gastown's existing
`health-scan`; it measures and reports, while the existing formula keeps the
escalation decision.

One important current-status caveat: a managed-session path bug means that
the deterministic pass does not actually run today and silently falls back to
LLM judgment. Open PR #243 contains the dispatch fix. The checker and threshold
design are adopted, but the protection is not operational until that fix
merges.

The default threshold is deliberately 90 minutes rather than 15. This PR was
calibrated to a roughly 60-second self-scheduled witness loop, while
gastown's witness deliberately allows one hour of legitimate idle silence.
The 90-minute default clears that healthy window and remains far below the
reported 14–63 hour stalls. It is configurable.

`maintenance` itself was removed in `d3617d1` and folded into SDK `core` in
gascity `f895c0ff4`, so I'm closing this original PR now that its genuinely
new design has an owner and its wiring fix is tracked in #243.

PR #109, opened 2026-06-16
fix(slack-channel): retry adapter registration on city-startup race

sjarmak:
## Problem

The `slack-channel` adapter calls `log.Fatalf` and exits permanently if its registration POST to the gc API fails — including during city startup, when the gc server may not yet be accepting adapter registrations. The result is a registration *race*: a transient `404`/`5xx`/transport error at boot kills the adapter for good instead of waiting for the server to come up.

## Fix

- `postJSON` now returns a typed `*statusError{StatusCode, Status, Body}` so callers can classify failures by HTTP status (was an opaque `fmt.Errorf`).
- New `registerAdapterWithRetry`: bounded exponential backoff (1s → cap 30s) on **transient** failures — `404`, any `5xx`, and transport errors — until a **5-minute deadline** (`context.WithTimeout`), then `log.Fatalf`. Permanent failures (`400`/`401`/`403`, other non-404 `4xx`) **fail fast** on the first attempt, so a real misconfiguration surfaces immediately instead of hanging.
- The deadline context is propagated into each attempt and checked in the backoff `select`, so the loop can't spin hot and is bounded by the deadline.

## Tests

3 new tests in `gcclient_test.go` against a real `httptest.Server` with atomic call counts:
- `404 ×3 then 200` → exactly N+1 calls (retry-then-success)
- deadline expiry → fails with the expected message after ≥1 attempt
- `400` → exactly 1 call (fail-fast, non-retryable)

`go test ./slack-channel/adapter/...` and `go test -race ./...` both clean (65 tests).

4 files, +184/-3.

### DIFF
--- a/slack-channel/adapter/config.go
+++ b/slack-channel/adapter/config.go
@@ -33,6 +33,16 @@ const (
 	// cannot pin an inbound-bridge goroutine (or block startup
 	// registration) forever.
 	gcCallTimeout = 15 * time.Second
+
+	// Startup registration retry: on a supervisor/city restart the adapter
+	// can come up before the city has finished adopting sessions, so the
+	// first registration gets a 404 "city not found or not running". The
+	// adapter retries with exponential backoff (initial→max) until
+	// registerDeadline rather than exiting on the first 404 and killing all
+	// Slack comms until a manual restart; past the deadline it fails loudly.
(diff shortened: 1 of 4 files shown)


sjarmak:
@copilot review

PR #121, opened 2026-06-18
fix(gastown): patrol queries use --type=molecule not invalid --type=wisp (port of gascity#3416)

brandonmartin:
## What
The deacon + refinery patrol startup blocks and formulas query the current/assigned patrol wisp with `gc bd list … --type=wisp`. `wisp` is **not a valid bd type** — wisp roots are `issue_type=molecule` — so the query matches nothing, the `CURRENT_WISP`/`ASSIGNED_WISP` lookup always returns empty, and a **fresh patrol wisp is poured every restart** (the leak @quad341 confirmed is still shipping). This substitutes `--type=wisp → --type=molecule` on the 10 real queries (deacon + refinery prompts + formulas) plus one doc reference. The witness template was already fixed; this brings the other two patrol agents in line. The explanatory "never use `--type=wisp`" warnings are intentionally left.

## Provenance
Ports the query fix from `gastownhall/gascity#3416`, which @quad341 reviewed and confirmed correct, then closed because #3335 moved the gastown pack out of gascity into this repo. Re-rooted to this repo's `gastown/…` paths, authored by us.

## Note on scope — reconcile-to-exactly-one
gascity#3416 also added a witness-style **reconcile-to-exactly-one (burn the surplus)** startup block to deacon + refinery. This PR is just the `--type` query fix, which stops *new* leaks. The deacon/refinery currently lack the surplus-burn the witness has (`witness/prompt.template.md` Step 1), so any wisps already leaked from prior `--type=wisp` restarts won't be auto-cleaned. Happy to add that surplus-burn block in this PR if you'd like it here — say the word and I'll mirror the witness pattern into both.

### DIFF
--- a/gastown/agents/deacon/prompt.template.md
+++ b/gastown/agents/deacon/prompt.template.md
@@ -89,9 +89,9 @@ If `next-iteration` already ran, do not pour again; run `gc hook`.
 ```bash
 CURRENT_WISP=${GC_BEAD_ID:-}
 if [ -z "$CURRENT_WISP" ]; then
-  CURRENT_WISP=$(gc bd list --assignee="$GC_AGENT" --status=in_progress --type=wisp --limit=1 --json | jq -r '.[0].id // empty')
+  CURRENT_WISP=$(gc bd list --assignee="$GC_AGENT" --status=in_progress --type=molecule --limit=1 --json | jq -r '.[0].id // empty')
 fi
-ASSIGNED_WISP=$(gc bd list --assignee="$GC_AGENT" --status=open --type=wisp --limit=1 --json | jq -r '.[0].id // empty')
+ASSIGNED_WISP=$(gc bd list --assignee="$GC_AGENT" --status=open --type=molecule --limit=1 --json | jq -r '.[0].id // empty')
 if [ -n "$CURRENT_WISP" ] && [ -z "$ASSIGNED_WISP" ]; then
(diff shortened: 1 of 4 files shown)


quad341 (approved):
Thanks @brandonmartin — this is the kind of fix that's easy to undervalue and easy to get subtly wrong, and you nailed both. Porting the `--type=wisp → --type=molecule` correction from gascity#3416, re-rooting it onto this repo's `gastown/…` paths, and catching all ten real query sites plus the stale doc reference lines the deacon and refinery up exactly with the already-fixed witness. I verified against `witness/prompt.template.md` that `molecule` is the right root type and that the old `wisp` queries silently matched nothing — so this genuinely stops the per-restart wisp leak @quad341 flagged. The TOML still parses cleanly and there's no maintainer-side cleanup, so I'm merging as-is.

On your scope note: yes please — I'd love the witness-style reconcile-to-exactly-one (surplus-burn) block mirrored into deacon + refinery, since this stops *new* leaks but won't reap the ones already poured. A follow-up PR is perfect so this clean one ships now. The provenance write-up made this a pleasure to review — thank you.


---
**Reviewers**: Qwen (local) — ok · Claude (claude-opus-4-8[1m]) — ok · Codex (gpt-5.5) — ok
**Synthesis**: claude-opus-4-8[1m]


Verified locally + tests pass. Ready to merge — clicking the button is the only step left.

PR #123, opened 2026-06-19
fix(refinery): clarify mechanical conflict handling

sarendipitee:
## Summary
- clarify that refinery should resolve obvious mechanical rebase conflicts itself
- keep semantic/product decisions routed back to the polecat pool
- include conflicted paths and unresolved decisions in rejection metadata

## Rationale
Refinery agents have been overly cautious around trivial conflicts and often push work back to the polecat pool by default. That creates unnecessary churn: pool resets, repeated pickup/rebase cycles, and stale branches for conflicts the refinery can safely resolve as part of normal merge processing. This keeps the boundary explicit: mechanical integration is refinery-owned; developer judgment still goes back to polecat.

## Validation
- go test ./...

### DIFF
--- a/gastown/agents/refinery/prompt.template.md
+++ b/gastown/agents/refinery/prompt.template.md
@@ -13,10 +13,11 @@
 ## Your Role: REFINERY (Merge Queue Processor for {{ .RigName }})
 
 **CARDINAL RULE: You are a merge processor, NOT a developer.**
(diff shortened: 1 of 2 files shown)


sarendipitee:
Just wanted to add that I've been using this in my city for a while and the bit of re-wording and clarification has made a difference — I no longer see Refiners insta-rejecting every conflict without even attempting a resolution.

PR #134, opened 2026-06-23
fix(deacon): drop --type=wisp from mol-wisp scan queries

pnost:
## Problem

`mol-deacon-patrol` beads have `issue_type=molecule`, **not** `wisp`. The `--type=wisp` filter in the deacon's startup protocol and "No Idle State" fallback was silently returning 0 results.

**Consequence**: The fallback scan missed stale wisps from prior iterations, and on any crash-recovery restart that lands in the fallback path, it would find no `ASSIGNED_WISP` and pour a **duplicate successor wisp** each cycle.

**Evidence**: Reported by `gastown.deacon` on first patrol (city DB, bd 1.0.4). Stale wisp `ci-wisp-s9iq` was missed and had to be caught via an unfiltered assignee query.

## Fix

Drop `--type=wisp` from both queries on lines 92 and 94. Assignee + status filtering alone is sufficient and correct — the deacon is only ever assigned mol-deacon-patrol beads.

## Test

Deacon startup with an existing assigned mol-wisp bead correctly resolves `ASSIGNED_WISP` without the type filter.

### DIFF
--- a/gastown/agents/deacon/prompt.template.md
+++ b/gastown/agents/deacon/prompt.template.md
@@ -89,9 +89,9 @@ If `next-iteration` already ran, do not pour again; run `gc hook`.
 ```bash
 CURRENT_WISP=${GC_BEAD_ID:-}
 if [ -z "$CURRENT_WISP" ]; then
-  CURRENT_WISP=$(gc bd list --assignee="$GC_AGENT" --status=in_progress --type=wisp --limit=1 --json | jq -r '.[0].id // empty')
+  CURRENT_WISP=$(gc bd list --assignee="$GC_AGENT" --status=in_progress --limit=1 --json | jq -r '.[0].id // empty')
 fi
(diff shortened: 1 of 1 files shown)


aphexcx:
From the 2026-08-01 sweep of open PRs: this appears to be **superseded by `main`**. The same invalid `--type=wisp` queries in `gastown/agents/deacon/prompt.template.md` were fixed via #121 (1407301, "patrol queries use `--type=molecule` not invalid `--type=wisp`") — but with `--type=molecule` (the valid bd type for wisp roots) instead of dropping the type filter entirely, as this branch does. Current main's deacon prompt reads:

```bash
CURRENT_WISP=$(gc bd list --assignee="$GC_AGENT" --status=in_progress --type=molecule --limit=1 --json | jq -r '.[0].id // empty')
```

The filtered form is strictly tighter than the no-filter form here (it can't match non-molecule beads assigned to the agent), and this branch is now CONFLICTING against it. Recommend closing as superseded — flagging for the author/maintainers rather than closing unilaterally.

PR #138, opened 2026-06-24
fix(gastown): mol-refinery-patrol uses invalid --type=wisp to resolve its wisp

outdoorsea:
## Problem
`mol-refinery-patrol` resolves its current wisp with `gc bd list ... --type=wisp`, but **`wisp` is not a valid bd `issue_type`** — a wisp root is `issue_type=molecule` (`ephemeral=true`). The query errors (`invalid issue type "wisp"`), so `CURRENT_WISP` resolves empty.

The loop guard then *correctly* refuses to burn an unresolved current wisp after pouring the next iteration, producing a **transient duplicate `in_progress` wisp** every cycle. It self-heals the following cycle via reconcile-to-one; the only real leak window is a refinery dying between pour-new and burn-old. Low severity, but town-wide (every rig shares this formula) and it spams a resolve error each cycle.

## Fix
Replace `--type=wisp` → `--type=molecule` at the **5 executable wisp-resolve sites** (the `CURRENT_WISP=$(gc bd list --assignee="$GC_AGENT" --status=in_progress --type=wisp --limit=1 ...)` blocks), plus the **1 prose reference** in the patrol-summary step for consistency. 6 lines total.

## Why `--type=molecule` is correct
The sibling `gastown/formulas/mol-witness-patrol.toml` already uses and documents this exact rule:
- L517: `Wisp roots are issue_type=molecule (never --type=wisp, which is not a [valid type])`
- L520: `... --status=open --type=molecule --limit=0 ...`

This change just brings `mol-refinery-patrol` in line with that established convention.

## Impact / testing
- Pure formula-text fix; no behavior change beyond making the wisp-resolve query actually succeed (and the loop burn its current wisp).
- Projects down to each rig's `.beads/formulas/` on `gc import`.

Reported by a refinery health check (switchyard `gastown.witness`).

### DIFF
--- a/gastown/formulas/mol-refinery-patrol.toml
+++ b/gastown/formulas/mol-refinery-patrol.toml
@@ -119,7 +119,7 @@ wisp, burn the current wisp, then request a restart:
 ```bash
 CURRENT_WISP=${GC_BEAD_ID:-}
 if [ -z "$CURRENT_WISP" ]; then
-  CURRENT_WISP=$(gc bd list --assignee="$GC_AGENT" --status=in_progress --type=wisp --limit=1 --json | jq -r '.[0].id // empty')
+  CURRENT_WISP=$(gc bd list --assignee="$GC_AGENT" --status=in_progress --type=molecule --limit=1 --json | jq -r '.[0].id // empty')
 fi
 NEXT=$(gc bd mol wisp mol-refinery-patrol --root-only --var target_branch={{target_branch}} --var rig_name={{rig_name}} --var binding_prefix={{binding_prefix}} --json | jq -r '.new_epic_id // empty')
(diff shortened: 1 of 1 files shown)


aphexcx:
From the 2026-08-01 sweep of open PRs: this exact fix **already landed on `main`** via #121 (1407301, "patrol queries use `--type=molecule` not invalid `--type=wisp`"). Current main's `gastown/formulas/mol-refinery-patrol.toml` has zero `--type=wisp` occurrences and six `--type=molecule` ones — i.e. GitHub still reports this branch MERGEABLE, but merging it would be a no-op: there is no delta left to apply. Recommend closing as already-fixed — flagging for the author/maintainers rather than closing unilaterally.

PR #139, opened 2026-06-25
feat(gc-roles): add city-scope run-operator-city variant

iiirto:
## Problem

`gc-roles`' `run-operator` is `scope = "rig"`. When a city imports the roles pack at **city scope**, `expandCityImportedAgentsForRigs` expands it to one instance **per rig** (`<rig>/run-operator`) — there is no city-scope instance for a `[[named_session]]` to bind to. A city-scope `[[named_session]] template = "gc.run-operator"` therefore fails:

```
named_session "gc.run-operator": backing template "gc.run-operator" not found
after pack expansion; named session disabled until its template resolves
```

This blocks **city-scope generative pipelines** — formulas whose steps run at city root (e.g. a PRD-decomposition / cross-rig planning formula that force-loads contracts by absolute `$GC_CITY_PATH`). They currently have no stock deterministic-plumbing worker and must hand-roll a `scope = "city"` agent that just re-implements the GC Role Worker soul.

There's no config-only fix: scope is keyed on the agent's own `scope` field at pack **load**, while `ApplyPatches` (which *does* have a `Scope` override) runs later at **compose** — by then the only instances are per-rig, so a patch can't collapse them into one city instance.

## Change

Add `run-operator-city` — the city-scope sibling of `run-operator`:

- `agent.toml`: identical to `run-operator` except `scope = "city"` (keeps `fallback = true`, the universal role-pack convention).
- `prompt.template.md`: byte-identical GC Role Worker soul (matches the inline convention every role agent in this pack already uses).

Because `expandCityImportedAgentsForRigs` **skips** `scope = "city"` agents, `run-operator-city` stays a single city-root instance. The rig-scoped `run-operator` is **untouched**, so existing rig pipelines are unaffected.

## Validation

Imported the roles pack at city scope into a real city and pinned a `[[named_session]] template = "gc.run-operator-city" scope = "city"`. `gc config show`:

- The "backing template not found" warning is **gone** — the named session binds.
- Resolved config shows a single `name = "run-operator-city"`, empty `dir`, `scope = "city"` (one city-root instance).

## Open question for maintainers

Is a dedicated city-scope variant the shape you'd prefer, or would you rather make `run-operator`'s scope overridable at the import boundary (a binary change so one role can serve both scopes)? Happy to adjust.

### DIFF
--- a/gascity/roles/agents/run-operator-city/agent.toml
+++ b/gascity/roles/agents/run-operator-city/agent.toml
@@ -1,0 +1,3 @@
+description = "City-scope run operator for deterministic setup, validation, gates, metadata, and finalization at city root"
+scope = "city"
+fallback = true
--- a/gascity/roles/agents/run-operator-city/prompt.template.md
+++ b/gascity/roles/agents/run-operator-city/prompt.template.md
@@ -1,0 +1,237 @@
+# GC Role Worker
+
+You are `{{ .AgentName }}`, a Gas City `graph.v2` role worker for template
+`{{ .TemplateName }}`.
+
+## Core Rule
+
+You work only the routed bead assigned to this live session. Do not use
+`bd mol current` to infer workflow position. Do not assume a parent bead or
+root bead describes your work. The workflow graph advances through explicit
+ready beads, and you execute the ready bead claimed by this session.
+
+## Startup Claim Protocol
+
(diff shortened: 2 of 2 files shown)


iiirto:
Closing — going with an atlas-local city-scope agent instead of an upstream pack change.

PR #156, opened 2026-07-04
fix(slack-full): redact url_private token from slackDownloadToFile error logs

sjarmak:
Slack file-download errors could leak a pre-signed URL and its query token into logs. `slackDownloadToFile` wrote the raw `*url.Error` (URL plus token) on the transport path and an unredacted URL on the 4xx path.

## Change
Strip the `url_private` query token from `slackDownloadToFile` error output before it reaches `log.Printf` or the receipt error body, so download failures no longer expose the signed URL.

## Tests
Covered by the slack-full error-redaction tests. Gates green.


sjarmak:
Superseded by #162 — same fix, rebased onto current upstream main with a broader scope (redacts all remaining leak sites via a shared helper, this branch's stale base makes it unmergeable as-is). Closing rather than rebasing further; #162 covers everything this PR did plus the getUploadURL site.

PR #158, opened 2026-07-04
profiler pack v0: retroactive formula-run profiling (collect / report / compare / analyze)

csells:
Implements the pack-v0 scope of the profiling proposal (gastownhall/gascity#3925), motivated by the measured findings in gastownhall/gascity#3924.

## What's in the pack

- **`gc profile collect <root>`** — assembles a durable capture under `.gc/runtime/profiles/<root>/` from sources that already survive: the run's bead closure (linkage metadata + dependency edges to fixpoint), city session/nudge beads, the run's slice of the durable event log, transcripts (session_key lookup with a content-match fallback), usage facts, formula provenance, and git evidence. Best-effort throughout; every miss is disclosed in `manifest.json` (`gc.profile.capture.v1`).
- **`gc profile report <root> [--json|--html]`** — formula-agnostic timing analysis: step spans (ready → started → closed, dispatch waits computed from blocker-close times), session lanes by role, latency+cost totals, and findings tagged by fixable layer (`formula` / `config` / `platform`). `--html` emits a self-contained page.
- **`gc profile compare <a> <b>`** — totals and per-step wait/active deltas between two captures.
- **`profile-analyze` formula + `gc.profile-analyst` role** — the judgment layer: reads a capture and writes `analysis.md` beside it (why the slow steps were slow, model-vs-tool-vs-dispatch attribution from transcripts, suggested formula edits). Mechanical report stays deterministic; analysis is labeled as analysis.
- **`transcript-archive` order** (on `session.stopped`) — preserves transcripts, the one volatile source retroactive profiling depends on.
- **`bench-nullop` formula** — no-op step chain + fan-out to measure pure orchestration overhead (noise floor for `compare`).

## Validation

Tested against a fully hand-collected reference capture of a real 3 h 35 m `build-from-requirements` run:

| Check | Result |
|---|---|
| Bead closure | 110/110 exact |
| Session beads | 36/36 exact |
| Event slice | 2,095/2,095 (identical predicate vs reference) |
| Transcripts | 32/36 (the 4 misses are marker-less short sessions the manual pass couldn't attribute either) |
| Report totals | exact: 215.32 m wall, 60.13 m dispatch wait over 40 spans, 569.63 m cumulative session time |
| HTML report | rendered and visually verified |

Two collector bugs were found and fixed *by* this validation: `gc.input_convoy_id` is repointed mid-run (the launch convoy survives only in `gc.var.convoy_id`), and session beads carry no `session_key`, so transcript resolution needs the content-match fallback.

## Known v0 limits (also in README)

Events are read from the active city log only (rotated archives not yet scanned); only the Claude Code transcript layout is resolved; reconciler trace slices are not captured; cost rollup depends on the city's usage sink being enabled (missing dimension is disclosed in the manifest).


sjarmak:
Thanks for this, @csells. The capture pipeline is a genuinely useful addition and the design is close. We ran a full review (3 independent passes plus a grounded cross-check against head `917764f`) and found 6 things worth resolving before merge:

1. **Standalone install:** `bench-nullop`'s 7 steps all call `gc.run-operator`, defined only in the gascity pack, but `profiler/pack.toml` has no `[imports]` and the README doesn't list gascity as a prerequisite, so the pack won't run on its own. Sibling packs (bmad, compound-engineering) import `../gascity` for this.
2. **Analysis in the report layer:** `profile_report.py` `analyze()` (L97-120) does keyword matching on step titles, hardcoded thresholds, and prescriptive recommendations inside the deterministic report code. The README assigns that reasoning to the `profile-analyze` model role, so it likely belongs there rather than in `report.py`.
3. **Path handling (2 spots):** `args.root/capture/a/b` drive `makedirs`+writes before the store is validated, and `session_key` from bead metadata is concatenated into the transcript path unsanitized, which can escape `~/.claude/projects/`. The github/discord intake code has allowlist sanitizers to reuse.
4. **`--html` crashes on in-flight roots:** `window.end = root.closed_at` is `None` for an unclosed root (`report.py:124/243`, TypeError), and `collect` has no None fallback on `created_at`, which contradicts the "never fatal" docstring. Text and `--json` are fine.
5. README advertises a token/cost rollup `report.py` doesn't compute (no `usage.jsonl` consumer).
6. No tests yet on ~1180 lines; `analyze()` and `closure()` are pure functions and cheap to cover.

Happy to paste the detailed follow-up notes inline for any of these. @julianknutsen, flagging for how you'd like to steer the disposition (formal request-changes vs. letting csells iterate on the above). Glad to help however's useful.

julianknutsen:
Status update, @csells: we adopted `profiler` as a first-party pack in
https://github.com/gastownhall/gascity-packs/pull/238, merged as `6373985`.
Your three commits were cherry-picked with authorship intact, and the
`profiler/` tree was verified byte-identical to this PR's head before the
adoption fixes.

The adoption addressed the standalone gascity import, path traversal,
in-flight report failures, the missing usage/cost rollup, and the absence of
tests. It also fixed several additional integration defects found while
exercising the pack with the real `gc`, including the unreachable analyst
role and nondeterministic capture output. The deterministic-report versus
model-analysis boundary remains an explicit follow-up design question rather
than being silently changed.

Thank you for the original capture pipeline and for validating it against a
real 3.5-hour run—that work is now in the first-party tree. I'm closing this
original PR as adopted and superseded by #238.

PR #164, opened 2026-07-06
fix(gascity): make build-artifact validator robust to stripped sys.path and shallow vendoring

richardkiene:
Fixes two field-reported robustness bugs in `gascity/assets/scripts/validate_build_artifact.py`.

## Bug 1: PyYAML unavailability was deferred to a hard failure, not handled

The script caught `ImportError` at import time and set `yaml = None`, but `parse_front_matter` and `load_schema` then unconditionally raised `PyYAML is required...` — no recovery attempt.

Observed in the field: a check-runner had PyYAML installed only under the invoking user's `~/.local/.../site-packages`, and the launcher excluded user site-packages (reproduced exactly with `PYTHONNOUSERSITE=1`). Every artifact check in that environment failed regardless of artifact validity.

**Fix:** on `ImportError`, retry the import after explicitly appending `site.getusersitepackages()` and `site.getsitepackages()` to `sys.path` (user site first). If yaml is still unavailable, the validation error now names the site-packages roots that were checked, e.g.:

```
error: PyYAML is required to parse build artifacts; import failed even after appending
these site-packages roots to sys.path: /home/ci/.local/lib/python3.12/site-packages, ...
Install PyYAML for this interpreter or into one of those roots.
```

## Bug 2: `SCHEMA_ROOT` path math broke under partial vendoring — and `schemas/` isn't vendored there anyway

`SCHEMA_ROOT = Path(__file__).resolve().parents[2] / "schemas" / "build"` assumed the canonical pack depth. A consumer vendoring only the `scripts/` subtree (observed: `.gc/scripts/validate_build_artifact.py`, two levels deep) made `parents[2]` overshoot — and the `schemas/` directory wasn't vendored at any depth, so path-math fixes alone couldn't work.

**Fix:** the six schema definitions are embedded in the script as `EMBEDDED_SCHEMAS`, and `load_schema` is a direct dict lookup with no filesystem access. Schema lookup now works at any vendoring depth with the script as the only file present. `gascity/schemas/build/*.yaml` remain the source of truth: a new test asserts each embedded definition equals `yaml.safe_load` of its source file (and that the schema-id sets match exactly), so any unmirrored schema edit fails CI. `load_schema` no longer needs yaml at all.

No changes to `build-artifact-valid.sh` or any vendoring logic.

## Tests

Four new tests in `gascity/tests/test_validators.py`, each written first and confirmed to reproduce the original failure before the fix:

- CLI run of a shallow-vendored copy (`.gc/scripts/`, no `schemas/` anywhere) validates successfully
- CLI recovers PyYAML from user site-packages when the default `sys.path` lacks it (`python -S` + `PYTHONUSERBASE`), including asserting the recovered module's provenance
- PyYAML missing everywhere still fails cleanly (exit 1, no traceback) with the checked roots named in stderr
- Embedded schemas stay in sync with `gascity/schemas/build/*.yaml`

## Verification

- Full CI pytest invocation (`tests contributing/tests gascity/tests discord/tests github/tests slack-full/tests slack-channel/tests pr-pipeline/tests`): **734 passed, 11088 subtests**
- Gastown shell tests and root-module `go test .` (embed tests): pass
- Manual reproduction of the exact field scenario (`PYTHONNOUSERSITE=1`, PyYAML only in user site): validator now exits 0

### DIFF
--- a/gascity/assets/scripts/validate_build_artifact.py
+++ b/gascity/assets/scripts/validate_build_artifact.py
@@ -9,15 +9,156 @@ from dataclasses import dataclass
 from pathlib import Path
 from typing import Any
 
-try:
-    import yaml
-except ImportError:  # pragma: no cover
-    yaml = None
+def _site_packages_roots() -> list[str]:
+    import site
+
+    roots: list[str] = []
+    for candidate in [site.getusersitepackages(), *site.getsitepackages()]:
+        if candidate and candidate not in roots:
+            roots.append(candidate)
+    return roots
+
+
+def _import_yaml() -> tuple[Any, list[str]]:
+    """Import PyYAML, retrying with site-packages roots that launchers can
+    strip from sys.path (e.g. PYTHONNOUSERSITE=1 hiding a user-site install).
+    Returns the module (or None) and the roots appended before the retry."""
+    try:
+        import yaml as module
+    except ImportError:
+        pass
+    else:
+        return module, []
+    roots = _site_packages_roots()
+    for root in roots:
+        if root not in sys.path:
+            sys.path.append(root)
+    try:
+        import yaml as module
+    except ImportError:
+        return None, roots
+    return module, roots
+
+
+yaml, YAML_SITE_ROOTS_CHECKED = _import_yaml()
 
 
 FRONT_MATTER_RE = re.compile(r"\A---\n(?P<front>.*?)\n---(?:\n|\Z)(?P<body>.*)\Z", re.DOTALL)
-SCHEMA_ROOT = Path(__file__).resolve().parents[2] / "schemas" / "build"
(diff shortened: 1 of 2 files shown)


jomadu:
We independently hit the underlying bug this PR fixes, via a cousin path in our own fork of `build-artifact-valid.sh`: a vendored-PyYAML fallback whose candidate-path search breaks under an invocation shape that doesn't preserve the real script path. We patched around it locally by widening the candidate search, but it's a workaround on top of vendoring — the `sys.path` retry against real `site-packages` this PR takes looks like the more durable fix, since it sidesteps path-resolution fragility entirely rather than patching one more case of it.

This predates #340 by about six weeks and neither references the other, so we've left a comment there linking back here. Given the overlap, is there appetite to resolve the conflicts and revive this rather than have #340 spawn a second, parallel fix? Happy to share our repro/patch as an additional data point if it helps move it along.

PR #149, opened 2026-06-30
refactor(packs): resolve ralph check scripts via ../assets layer paths

julianknutsen:
> **Draft — depends on [gastownhall/gascity#3834](https://github.com/gastownhall/gascity/pull/3834).** Do not merge until that ships and the embedded pack pin is bumped; until then the migrated paths won't resolve at runtime.

## Summary

Migrates every ralph exec-check reference off the brittle `.gc/scripts/checks/<name>.sh` runtime path to the layer-resolved `../assets/scripts/checks/<name>.sh` form (symmetric with `description_file`). **126 files, 162/162 lines** — a pure path substitution:

- **78** formula `path =` declarations
- **77** workflow-prompt references
- **7** references in 2 guardrail tests (`gascity/tests/test_formula_assets.py`, `test_derived_pack_compatibility.py`), updated in lockstep so they now enforce the new contract

across **gascity, bmad, compound-engineering, gstack, superpowers**.

## Why

`.gc/scripts/checks/` is **not materialized** from pack `assets/` in a running city — the legacy `ResolveScripts` shim that used to populate it is retired (it's only pruned now), so today these references resolve to nothing outside the test harness (which stages the scripts by hand). PR #3834 teaches the formula compiler to resolve `../assets/...` check paths at cook time to the highest-priority formula layer that ships the script — no `.gc/` materialization, and a city/local layer can override a pack's check script via normal layer shadowing.

## Cross-pack resolution

The methodology packs reference scripts they don't ship themselves (`build-artifact-valid.sh`, `implementation-review-approved.sh`, `design-review-approved.sh` live in **gascity**). Each methodology pack imports gascity (`[imports.gc] source = "../gascity"`), so `../assets/...` resolves through the import layer stack to gascity's copy — the **same mechanism `description_file` already uses**, and the same runtime dependency these packs already had on gascity providing the script.

## ⚠️ One reviewer decision: imperative manual-run prompts

~6 prompt lines instruct the agent to run the check manually, e.g.:

> run the same validator locally from that rig root with `GC_BEAD_ID=<id> ../assets/scripts/checks/build-artifact-valid.sh`

[...]

PR #169, opened 2026-07-06
Fix Gas City formula check asset paths

duncan4123:
## Summary
- point Gas City formula check gates at pack-shipped assets instead of city-local `.gc/scripts/checks` paths
- update formula asset tests to enforce the asset-path contract
- add a regression guard against reintroducing `.gc/scripts/checks` formula paths

## Validation
- `python3 -m pytest gascity/tests/test_formula_assets.py -q`
- real-city smoke in `/data/projects/fff`: fresh `build-basic` workflow generated 7 check beads and 0 legacy `.gc/scripts/checks` paths

## Context
A real control-dispatcher run quarantined a build-basic ralph gate because the formula stamped `.gc/scripts/checks/build-artifact-valid.sh`, but the gascity pack ships the validator under `assets/scripts/checks/`. The formula compiler already resolves `../assets/...` check paths to the installed/cached pack asset, so this moves the pack formulas onto that supported path.

### DIFF
--- a/gascity/formulas/build-base.formula.toml
+++ b/gascity/formulas/build-base.formula.toml
@@ -128,7 +128,7 @@ max_attempts = 3
 
 [steps.check.check]
 mode = "exec"
-path = ".gc/scripts/checks/build-artifact-valid.sh"
+path = "../assets/scripts/checks/build-artifact-valid.sh"
 timeout = "5m"
 
 [[steps]]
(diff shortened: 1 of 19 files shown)


aphexcx:
Coordination note from the 2026-08-01 sweep of open PRs: this PR, #194, and #149 all migrate check references off `.gc/scripts/checks/...` onto layer-resolved `../assets/scripts/checks/...` paths.

This PR and #194 are both yours (@duncan4123) and cover the same gascity scope — comparing the diffs, **#194 is a strict superset of this one**: 18 of the 19 shared files carry byte-identical hunks, its `test_formula_assets.py` is the evolved version, and it adds the `gascity/REQUIREMENTS.md` path-ownership docs plus a `test_derived_pack_compatibility.py` guard.

**Recommendation: close this one in favor of #194** (please confirm it's the intended successor), with #149 following after for the remaining packs — full proposed sequencing is in the comment on #194.

PR #170, opened 2026-07-06
Add Postgres backend pack

duncan4123:
## Summary
- add the bd-gc-postgres backend pack for plugin-backed Postgres Beads/Gas City cities
- include local Postgres provisioning, backend build/install commands, and GC fastpath endpoint metadata
- install tmux navigation/status keybinding assets from the pack so postgres cities do not depend on runtime workspace paths

## Validation
- sh -n on bd-gc-postgres tmux keybinding scripts
- gc config show in /data/projects/postgres resolves the keybinding hook from bd-gc-postgres/assets/scripts
- gc status in /data/projects/postgres reports the city running with GascityBackendStore

### DIFF
--- a/bd-gc-postgres/README.md
+++ b/bd-gc-postgres/README.md
@@ -18,6 +18,10 @@ launches a trusted local plugin process over the `beads.backend.v1alpha1`
 newline-delimited JSON protocol. The plugin owns direct Postgres access and
 Beads core talks to it through the plugin adapter.
 
+If the runtime plugin binary is missing during `gc init`, the setup hook runs
+the pack build command itself before provisioning Postgres. This keeps init
+working even when the caller has not run the pack prepare step yet.
+
 ## Configuration
 
 By default, `gc init --beads-backend postgres` provisions a local Postgres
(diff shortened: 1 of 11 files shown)


julianknutsen:
Thanks for building the original PostgreSQL plugin bridge, @duncan4123. The
later native-backend path in #197 appears to have overtaken its provisioning
goal, and those backends now ship in stock `bd`.

The current product path is stock `bd` for backend support. This PR also
carries plugin-fastpath and tmux assets that depend on a `backend_plugins`
integration seam; the proposed seam in gascity#3980 remains open and
conflicting. When that contract or a successor lands, those assets have a
clear path to return as an owner-published pack.

If that SDK seam or a successor lands and `bd-gc-postgres` develops a distinct
use again, it can still be published without moving ownership to us: keep your
repo canonical, set
`[pack].name = "duncan4123/bd-gc-postgres"`, and follow
https://registry.gascity.com/publish. The recommended path is to install the
Registry GitHub App and use **Find packs**; the documented Actions OIDC
workflow is there for tagged releases.

Every production release currently waits for staff review, including repeats.
I'm closing this snapshot at the current architecture boundary, and we'd be
glad to see it return through the Registry when the SDK seam and a distinct
use line up.

PR #178, opened 2026-07-06
fix(maintenance): increase witness heartbeat staleness threshold to 30m

sarendipitee:
**Stacked on #102** — only the final commit (`fix(maintenance): increase witness heartbeat staleness threshold to 30m`) is new; the rest of the diff is the rig-liveness-watchdog series already under review in #95 / #99 / #100 / #102. Best reviewed commit-by-commit or after #102 merges (this branch will then rebase to a one-commit diff).

## Problem
The rig-liveness watchdog flags a witness as STALLED when its heartbeat is older than `WATCHDOG_WITNESS_STALE_MIN` (previously 15 minutes) and resets it. On busy rigs the witness patrol cycle itself can take 15-17 minutes because of slow orphan-recovery queries against Dolt, so a perfectly healthy witness routinely exceeded the 15m threshold.

## Root cause
The staleness threshold was tighter than the worst-case healthy patrol cycle duration, so the watchdog reset live witnesses, which restarted the patrol, which again exceeded the threshold — an infinite reset loop that also produced repeated escalations.

## Fix
Raise the default `WATCHDOG_WITNESS_STALE_MIN` from 15 to 30 minutes (script default plus docs). This is a conservative default change only; operators can still tune it via the `WATCHDOG_WITNESS_STALE_MIN` env var. Longer-term fixes (faster orphan-recovery queries or mid-cycle heartbeats in the witness patrol loop) would let this come back down.

## Testing
- `bash -n maintenance/assets/scripts/rig-liveness-watchdog.sh` passes.
- Observed in production-like rig: with the 15m default the watchdog reset a healthy witness every cycle; at 30m the false-positive resets stopped while genuinely dead witnesses (no heartbeat for >30m) are still caught.

### DIFF
--- a/PR_DESCRIPTION.md
+++ b/PR_DESCRIPTION.md
@@ -1,0 +1,132 @@
+# fix(maintenance): stop rig-liveness-watchdog false-positive [INCIDENT] spam
+
+## Problem
+
+The `rig-liveness-watchdog` exec order was flooding inboxes with bogus
+`[INCIDENT]` escalation mail on idle and suspended dev rigs. Three independent
+bugs all contributed:
+
+---
+
+### Root cause 1 — Merge-freshness measured against stale local ref
+
+**What was happening:** The stall-detection code measured "how long since the
+last merge?" against the local `default_branch` ref (`refs/heads/main`). On a
+quiet dev rig that isn't actively pulling, this local ref can lag hours behind
+`origin/main`. The watchdog would see a stale local pointer, conclude nothing
+had merged recently, and fire a stall alert even though merges were landing on
+origin with no backlog at all.
+
+**Compounding factor — branch-count as stall signal:** The old code also used
+`git for-each-ref --no-contains <default>` to count "unmerged" branches and
+treated a non-zero count as evidence of a backlog. This has a fundamental blind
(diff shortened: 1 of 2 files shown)


julianknutsen:
Thanks for surfacing the threshold concern, @sarendipitee. We adopted the
underlying silent-heartbeat signal from
#99 in merged gastown PR #237. The current design uses a different threshold
for gastown's runtime model. Its deterministic pass is not operational yet: a
managed-session path bug currently bypasses it, and open #243 is the pending
dispatch fix.

Gastown's witness deliberately permits one hour of legitimate idle silence
(`idle_timeout = "1h"` and no self-scheduled wake loop), so both 15 and 30
minutes would false-alert on healthy behavior. The adopted check defaults to
90 minutes—1.5× the legitimate window, still far below the reported 14–63
hour stalls—and is configurable.

If a current witness patrol cycle still spends 15–17 minutes in orphan
recovery, that is worth a fresh performance issue in `gastownhall/gascity`
with a current trace; it is independent of this watchdog threshold.
`maintenance` itself was removed in `d3617d1` and folded into SDK `core` in
gascity `f895c0ff4`, so I'm closing this stacked PR now that the relevant
design has an owner and the wiring defect is tracked in #243.

PR #180, opened 2026-07-07
registry: add agentops-membrane (fail-closed cross-family verification close door)

boshu2:
Adds the **agentops-membrane** pack to the registry: a fail-closed, verdict-bound close door for Gas City workflows.

What it provides:
- A `membrane-quest` formula whose `[steps.check]` routes the build diff + acceptance contract to >=2 cross-family, fresh-context reviewer lanes and fail-closes on anything but a deterministic CONFIRMED (per-round nonce anti-replay; DEGRADED-aware — transient lane loss never converts to a false refute or a silent pass).
- LAW-0-safe provider config fragments (print_args=[] for claude/antigravity), doctor checks (print-sink guard, membrane health), lane-keepalive + canary orders, quest scaffolding, and a QUICKSTART.

Source: https://github.com/boshu2/agentops/tree/main/packs/agentops-membrane (MIT). Content hash computed with `gc pack release hash` at the pinned commit. Battle-tested on live cities; the formula ships with protection against agent-minted closes of the engine gate bead (related hardening PR on the SDK side: gastownhall/gascity#3985).

### DIFF
--- a/registry.toml
+++ b/registry.toml
@@ -261,3 +261,16 @@ schema = 1
     commit = "a1490ecf90238f3af52ab2e7f24e30b5b62cbfc2"
     hash = "sha256:c06f30e61407994cef7e16aafecbc71afbbc9f8cdfd0d506146fdf95837333cf"
     description = "Initial oversight-rig release: rig-scoped project-lead + deterministic escalation machinery."
+
+[[pack]]
+  name = "agentops-membrane"
+  description = "Fail-closed, verdict-bound close door for Gas City workflows: cross-family fresh-context review quorum (deterministic finalize, per-round nonce, DEGRADED-aware), LAW-0-safe provider config, doctor checks, and quest scaffolding. From the AgentOps verification membrane."
+  source = "https://github.com/boshu2/agentops/tree/main/packs/agentops-membrane"
+  source_kind = "git"
+
+  [[pack.release]]
+    version = "0.1.0"
+    ref = "main"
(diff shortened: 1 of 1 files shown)


julianknutsen:
Thanks for this, @boshu2. Before we choose the publication route: do you still
want to publish `agentops-membrane`?

The source path recorded by this PR,
`boshu2/agentops/packs/agentops-membrane` on `main`, now returns 404. Release
commit `6e2e4c1` still contains the pack, but its canonical `main` link is
dead, so the GitHub App **Find packs** route cannot discover it at
default-branch `HEAD`.

There is also a dependency-link mismatch to resolve. This PR describes
gastownhall/gascity#3985 as engine-gate hardening, but #3985 currently
contains tmux session-teardown/PID-reuse safety work and does not mention the
membrane gate. Was the wrong SDK PR linked, or did the dependency change? If
an SDK prerequisite remains, what exact issue/commit supplies it?

If the pack is still intended, restore it in a public repo you own, set
`[pack].name = "boshu2/agentops-membrane"`, and follow
https://registry.gascity.com/publish. Install the Registry GitHub App, sign in,
and use **Find packs**; that path supports a `pack.toml` below the repository
root and scans default-branch `HEAD`. Use Actions OIDC for a tag. Production
currently sends every release to staff review; repository proof clears
ownership but is not approval.

We also need to know whether the pack is safe without that actual SDK
prerequisite or needs a documented minimum version to preserve its
fail-closed guarantee.

I'm leaving this open for your answer. If the deletion means the pack is no
longer wanted, say so and we'll close it cleanly.

PR #181, opened 2026-07-07
fix(slack-mini): persist-and-retry inbound forwards, sender display names, mrkdwn conversion, registered reply instructions

aphexcx:
Two commits hardening the slack-mini adapter, developed and live-verified on a Tier-1 city where Slack is the operator's primary mobile channel:

**1. Persist-and-retry for inbound forwards (hq-1q1, 1883248):** the adapter previously acked Slack before forwarding to gc — a gc-API timeout meant the message was lost AND fed Slack's per-app backoff. The adapter now logs receipt pre-forward, spools to disk (GC_SERVICE_STATE_ROOT/data/inbound-spool), retries 5x over ~2min, dead-letters with the original payload on exhaustion, and replays crash orphans at startup. (This path saved a real message from loss during a 3-minute gc-API stall the same night it deployed.)

**2. Reply UX fixes (hq-fh9, 34ea912):** sender display names via users.info (graceful raw-id fallback with 5m negative cache when users:read is not yet granted), Markdown→mrkdwn conversion on outbound posts, and reply instructions registered with gc so injected reminders name a verb that actually exists, with thread ts pre-filled.

Both deployed and verified end-to-end on a live city (signed synthetic inbound through funnel→adapter→gc→agent, reply posted in-thread with formatting).

### DIFF
--- a/slack-mini/CHANGELOG.md
+++ b/slack-mini/CHANGELOG.md
@@ -7,6 +7,23 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- Codex review round on this branch (1 P1 + 4 P2, all fixed): the
+  decoded mention is spooled synchronously BEFORE the events endpoint
+  200-acks Slack (the old ack → async users.info → spool order lost
+  acked mentions to a crash in the enrichment window; enrichment and
+  delivery stay async, and startup replay now finishes enrichment
+  too); spool entries are written atomically (same-dir temp file +
+  fsync + rename) so a torn write can no longer dead-letter an acked
+  message with no valid payload left; Markdown link destinations and
+  bare URLs are converted/protected BEFORE the emphasis passes (a
+  destination like `.../pkg/__init__.py` was being rewritten into a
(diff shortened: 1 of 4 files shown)


aphexcx:
Rebased onto current main (44b2eef, was 4 behind) — clean replay, adapter module tests pass locally. This is the adapter-side companion to gastownhall/gascity#4014 (registered reply instructions land there); the persist-and-retry fix stands alone and stops inbound Slack messages being lost when the gc API is briefly unavailable. Ready for review.

aphexcx:

Three codex rounds over the full diff; all P1/P2 fixed: mentions spool synchronously (atomic write + dir fsync) **before** the 200 ack with a 500-NACK when a configured spool fails; bounded live-delivery semaphore + bounded startup replay pool; coalesced users.info lookups; and a hardened GFM→mrkdwn converter (line-oriented fences incl. opener-length closers, balanced/escaped-paren link destinations, protected URLs before emphasis passes, intraword-underscore rule, label escaping, multi-backtick code spans).

Verification: gofmt/vet clean, `go test` + `-race` green.

**Blocked on:** CI workflow approval (`action_required`) and branch-protection merge — needs a gastownhall maintainer.

PR #189, opened 2026-07-09
fix(gastown): stop witness patrol self-burn in next-iteration reconciliation

bourgois:
## Summary
- \`mol-witness-patrol\`'s \`next-iteration\` step scanned \`--status=open\` wisps assigned to self to tell a genuinely leaked wisp (from a prior crashed/raced cycle) apart from the current one. But the bootstrap/resume path only ever sets \`assignee\` (never transitions status to \`in_progress\`), so the current wisp *is* open and becomes the only member of that scan.
- Net effect: \`NEXT\` resolves to the current wisp itself, the pour-new-wisp branch is skipped, and the wisp is burned with **no replacement poured** — wisp starvation, breaking the intended continuous same-turn handoff.
- Fix: replace the open-status scan with the \`CURRENT_WISP=${GC_BEAD_ID:-}\` pattern already proven in \`mol-deacon-patrol\` and \`mol-refinery-patrol\`'s own \`next-iteration\` steps — resolve the current wisp from runtime ground truth, pour+validate+assign the next wisp first, then burn the correctly-identified current one.

Companion fix for the same underlying issue: Voxist/beads#21 (\`bd update --claim\` wasn't actually idempotent for gc-orchestrated agents, which had foreclosed the alternate one-line fix of just always calling \`--claim\` on resume).

## Test plan
- [x] \`python3 -c 'tomllib.load(...)'\` — edited TOML parses, same 5-step graph intact
- [x] \`gastown/tests/test_gastown_pack_assets.sh\` — still passes
- [ ] Next witness patrol cycle after this lands: confirm \`next-iteration\` no longer self-burns (manual/runtime verification, can't unit-test a prompt formula)

### DIFF
--- a/gastown/formulas/mol-witness-patrol.toml
+++ b/gastown/formulas/mol-witness-patrol.toml
@@ -529,22 +529,30 @@ Reconcile the wisps assigned to self down to exactly one before pouring or
 burning anything. `$GC_BEAD_ID` is **not** load-bearing here — it is an
 unpopulated optimization in the live witness runtime (confirmed empty for
 this session type; see vg-5kv) and must never gate behavior or cause a
-non-zero exit when absent. The load-bearing signal is an **open-status
-scan**: a patrol wisp stays `status=open` for its entire working life (the
-bootstrap/resume path only ever sets `assignee`), so scanning `--status=open`
-wisps assigned to self reliably finds the current wisp even with
-`$GC_BEAD_ID` empty. Reduce that scan to one `ASSIGNED_WISP`, burning any
-surplus left by a prior raced cycle, then pour+assign the next wisp before
-burning the current one. The steady state (`$GC_BEAD_ID` empty, exactly one
-open wisp already assigned) must fall through to a silent no-op, never an
-error exit — a status=open scan that stops there would confuse "the current
+non-zero exit when absent — but the query below excludes it **by id**
+whenever it *is* set, so a populated `$GC_BEAD_ID` can never be mistaken for
+an already-queued replacement (review pass 4 Finding 1: checking presence of
+`$CURRENT_WISP`/`$ASSIGNED_WISP` alone, without comparing identity, silently
+reproduces this bead's original starvation bug the moment `$GC_BEAD_ID`
+becomes non-empty). The load-bearing signal is an **open-status scan,
+excluding the current wisp's own id**: a patrol wisp stays `status=open` for
+its entire working life (the bootstrap/resume path only ever sets
+`assignee`), so scanning `--status=open` wisps assigned to self — minus
+`$CURRENT_WISP` itself — reliably finds any *other* already-queued wisp even
+with `$GC_BEAD_ID` empty, and never confuses the current wisp for one.
+Reduce that scan to one `ASSIGNED_WISP`, burning any surplus left by a prior
+raced cycle, then pour+assign the next wisp before burning the current one.
+The steady state (`$GC_BEAD_ID` empty, exactly one open wisp already
+assigned) must fall through to a silent no-op, never an error exit — a scan
+that failed to exclude the current wisp by id would confuse "the current
 wisp" for "a leaked one" and burn the only wisp that exists with nothing
 poured to replace it (wisp starvation — see vg-5kv). Matches the
 mol-deacon-patrol / mol-refinery-patrol next-iteration convention.
 ```bash
 CURRENT_WISP=${GC_BEAD_ID:-}
 
-OPEN_WISPS=$(gc bd list --assignee="$GC_AGENT" --status=open --type=molecule --limit=0 --json | jq -r '.[].id')
+OPEN_WISPS=$(gc bd list --assignee="$GC_AGENT" --status=open --type=molecule --limit=0 --json \
+  | jq -r --arg self "$CURRENT_WISP" '.[] | select(.id != $self) | .id')
 ASSIGNED_WISP=$(printf '%s\n' $OPEN_WISPS | sed -n '1p')
 for extra in $(printf '%s\n' $OPEN_WISPS | sed '1d'); do
   gc bd mol burn "$extra" --force
--- a/gastown/formulas/mol-witness-patrol.toml
+++ b/gastown/formulas/mol-witness-patrol.toml
@@ -523,31 +523,65 @@ gc runtime request-restart
 This blocks forever. The controller restarts you. The next wisp
 is already assigned — the new session resumes from context.
 
-**2. Pour next iteration BEFORE burning (reconcile to exactly one):**
+**2. Resolve the current wisp, pour next iteration, then burn this wisp:**
 
-Reuse an already-queued wisp instead of pouring a duplicate. A prior cycle
-may have poured a next wisp without burning, or a restart may have raced —
-keep one open patrol wisp and burn any surplus, so wisps never accumulate.
-Wisp roots are `issue_type=molecule` (never `--type=wisp`, which is not a
-valid gc bd type and matches nothing).
+Reconcile the wisps assigned to self down to exactly one before pouring or
+burning anything. `$GC_BEAD_ID` is **not** load-bearing here — it is an
+unpopulated optimization in the live witness runtime (confirmed empty for
+this session type; see vg-5kv) and must never gate behavior or cause a
+non-zero exit when absent. The load-bearing signal is an **open-status
+scan**: a patrol wisp stays `status=open` for its entire working life (the
+bootstrap/resume path only ever sets `assignee`), so scanning `--status=open`
+wisps assigned to self reliably finds the current wisp even with
+`$GC_BEAD_ID` empty. Reduce that scan to one `ASSIGNED_WISP`, burning any
+surplus left by a prior raced cycle, then pour+assign the next wisp before
+burning the current one. The steady state (`$GC_BEAD_ID` empty, exactly one
+open wisp already assigned) must fall through to a silent no-op, never an
+error exit — a status=open scan that stops there would confuse "the current
+wisp" for "a leaked one" and burn the only wisp that exists with nothing
+poured to replace it (wisp starvation — see vg-5kv). Matches the
+mol-deacon-patrol / mol-refinery-patrol next-iteration convention.
 ```bash
+CURRENT_WISP=${GC_BEAD_ID:-}
+
 OPEN_WISPS=$(gc bd list --assignee="$GC_AGENT" --status=open --type=molecule --limit=0 --json | jq -r '.[].id')
-NEXT=$(printf '%s\n' $OPEN_WISPS | sed -n '1p')
+ASSIGNED_WISP=$(printf '%s\n' $OPEN_WISPS | sed -n '1p')
 for extra in $(printf '%s\n' $OPEN_WISPS | sed '1d'); do
   gc bd mol burn "$extra" --force
 done
-if [ -z "$NEXT" ]; then
-  NEXT=$(gc bd mol wisp mol-witness-patrol --root-only --var binding_prefix='{{binding_prefix}}' --json | jq -r '.new_epic_id')
-  gc bd update "$NEXT" --assignee="$GC_AGENT"
+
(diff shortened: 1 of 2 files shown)


bourgois:
**Reviewer verdict: sound fix, holding formal approval** (voxlingo/voxist.reviewer, vg-5kv)

Logic verified correct. The `CURRENT_WISP=${GC_BEAD_ID:-}` pattern + pour-NEXT-before-burn-CURRENT + fail-closed error handling (drain-ack + exit 1 on any failure, current wisp never burned without a confirmed successor) is byte-for-byte the same structure already in production in `mol-deacon-patrol.toml` and `mol-refinery-patrol.toml`'s own next-iteration steps — I diffed all three directly rather than taking the PR description's word for it. This directly fixes vg-5kv's root cause: the old `--status=open` scan couldn't distinguish 'this wisp' from a genuinely leaked one once the resume path left status=open, so NEXT resolved to the current wisp itself, the pour-new-wisp branch was skipped, and the only wisp got burned with no replacement. The new logic sidesteps that ambiguity entirely by using runtime ground truth instead of a status heuristic.

**Not approving yet — two things:**
1. CI (`CI` + `CodeQL`) is stuck in `action_required` on this cross-org fork PR (Voxist/gascity-packs → gastownhall/gascity-packs) and my identity has read-only access to gastownhall/gascity-packs (confirmed via `gh api repos/gastownhall/gascity-packs --jq .permissions` → push:false), so I can't clear it myself. Needs a maintainer with write access to approve the pending run before merge — right now this PR has zero automated verification, only the author's local claim.
2. The one suite in the stated test plan, `gastown/tests/test_gastown_pack_assets.sh`, doesn't actually cover `mol-witness-patrol.toml` at all (I read it end to end — it only exercises dog/shutdown-dance/refinery-direct-merge/polecat/review-leg assets). It'll pass regardless of what this diff does, so 'still passes' isn't evidence for this specific change beyond 'didn't break unrelated files.' Not a blocker, just don't read it as behavioral coverage.

Minor/non-blocking: the `CURRENT_WISP` fallback query (`gc bd list --assignee=... --status=in_progress --type=molecule`) reuses the bd-list query shape already confirmed broken for non-closed statuses (vg-3kn). This is inherited identically from the deacon/refinery formulas, not introduced here, and fails closed (refuses to burn) if it comes up empty — low real-world impact as long as $GC_BEAD_ID is reliably set for wisp-driven sessions. Worth a shared follow-up across all three patrol formulas once vg-bbl/vg-3kn lands.

Would approve once CI actually runs and passes. Full review notes: bd comment on vg-5kv (voxlingo rig).

aphexcx:
From the 2026-08-01 sweep of open PRs: this PR and #253 fix the **same witness-patrol self-burn bug** (vg-5kv) in `mol-witness-patrol.toml` with competing approaches. #253 is based on current `main` and MERGEABLE; this branch is CONFLICTING against the current file. The sweep's recommendation: **land #253 and close this one** rather than resolving this branch's conflicts.

That said, one idea here is worth keeping: this PR resolves the current wisp from `$GC_BEAD_ID` (matching the deacon/refinery next-iteration convention), which is a real hardening over #253's `THIS_WISP=<this-wisp-id>` placeholder — the latter still relies on the executing agent substituting the id correctly. Suggest porting the `$GC_BEAD_ID`-first resolution as a small follow-up on top of #253. Flagging for the author/maintainers rather than closing unilaterally.

bourgois:
## Reviewer verdict (voxlingo/voxist.reviewer) — pass 4, reviewing the 2026-08-08 rebase/correction — REWORK

No PII/GDPR/MDR surfaces here (pure gastown orchestration-formula + shell-test change, no personal-data flow) — concur with the 2026-07-28 planner's own N/A assessment.

Verified from scratch against the PR's actual current HEAD (fe03527cb95, cloned + checked out, not diff-read-only): ran the new `test_mol_witness_patrol_next_iteration.sh` for real (passes, genuinely, not just asserted), confirmed `.github/workflows/ci.yml`'s `gastown/tests/test_*.sh` glob will pick it up as a required check once Actions actually runs. Drain-ack-before-exit-1 restoration matches `mol-deacon-patrol.toml`'s live convention (diffed directly). T-007's "vg-k4h closed" claim verified true via direct `bd list` query. All of that is solid.

### Finding 1 [HIGH] — next-iteration still self-burns with zero replacement when `$GC_BEAD_ID` is populated (reproduces this bead's own root cause)

`gastown/formulas/mol-witness-patrol.toml`, step `next-iteration`, the new "2. Resolve the current wisp..." bash block.

The `if [ -n "$CURRENT_WISP" ] && [ -z "$ASSIGNED_WISP" ]` / `elif [ -n "$CURRENT_WISP" ]` pair decides whether to pour a replacement wisp based only on **presence** of `$CURRENT_WISP` and `$ASSIGNED_WISP`, never on whether they're **the same wisp**. `OPEN_WISPS`/`ASSIGNED_WISP` is computed with the exact unfiltered query from the pre-fix code — it never excludes `$CURRENT_WISP` by id — so the premise that `$GC_BEAD_ID` gives unambiguous "ground truth" isn't actually wired through into the reconciliation.

Reproduced empirically (extracted the live bash block out of the TOML at this PR's HEAD with the same tomllib approach the PR's own harness uses, ran it under the same mocked `gc`, only changing `GC_BEAD_ID` from unset to populated):

| scenario | GC_BEAD_ID | OPEN_WISPS | result |
|---|---|---|---|
| steady state, current known | `wisp-current` | `[wisp-current]` | burns `wisp-current`, **0 pours**, rc=0 |
| leaked distinct wisp, current sorts first | `wisp-current` | `[wisp-current, wisp-other]` | burns **both**, 0 pours, rc=0 |
| leaked distinct wisp, current sorts second | `wisp-current` | `[wisp-other, wisp-current]` | burns `wisp-current` **twice** (2nd call targets an already-burned id), `wisp-other` silently untouched/unpromoted, 0 pours, rc=0 |

All three exit 0 — this looks like success and would not be caught by a downstream health check, only by the same "witness has zero wisps" symptom that originally opened this bead.

This is currently masked in production only because `$GC_BEAD_ID` is confirmed always-empty for live witness sessions (this bead's own comment history across ~20 recurrences + bd memory `vg-5kv-gc-bead-id-empty-confirmed-live-witness`), so both `[ -n "$CURRENT_WISP" ]` branches are dead code today. But the PR's own added prose calls `$GC_BEAD_ID` "an unpopulated optimization" — wording that anticipates it becoming populated, not asserting it never will. And the new test file's one runner helper hardcodes `unset GC_BEAD_ID` in every case, so neither branch that would exercise this has any coverage — the exact code path that reproduces this bead's original defect ships untested.

**Suggested fix:** exclude `$CURRENT_WISP` from the `OPEN_WISPS` query by id (the way #253 already does it — see Finding 2):
```bash
OPEN_WISPS=$(gc bd list --assignee="$GC_AGENT" --status=open --type=molecule --limit=0 --json \
  | jq -r --arg self "$CURRENT_WISP" '.[] | select(.id != $self) | .id')
```
so `ASSIGNED_WISP` can never equal `CURRENT_WISP`, independent of list ordering or which branch fires. Add a 3rd test case: `GC_BEAD_ID` populated + equal to the sole open wisp → assert exactly one pour and exactly one burn (of the *old* id, never the newly-poured one).

### Finding 2 [MEDIUM] — unresolved competing PR on the identical hunk

gastownhall/gascity-packs#253 (`fix/witness-patrol-self-burn`, author atbrace, OPEN, `mergeable: MERGEABLE`) fixes the *same* reported bug on the *same lines* of the *same file*, via a simpler `THIS_WISP=<placeholder>` + `select(.id != $self)` exclude-by-id approach — which, notably, does not have Finding 1's defect, precisely because it excludes by identity rather than reasoning about presence.

A repo participant already left a comment on **this** PR (2026-08-01, aphexcx) recommending landing #253 and closing #189, while suggesting #189's `$GC_BEAD_ID`-first resolution idea be ported onto #253 as a follow-up. Today's rebase/correction pass doesn't mention #253 anywhere in its plan file, executor notes, or commits — an external review signal that hasn't been ingested or refuted (this role's re-review discipline treats that as a blocking input, not something to route around silently).

Shipping #189 without resolving this leaves two open PRs racing to edit the same hunk of the same file — whichever merges second either conflicts outright or, if force-merged, silently clobbers the other's fix. This bead has already lost real time to exactly this class of problem (see the 2026-08-01 "CONFLICTING" state after 67 unrelated commits landed on `main`).

**Suggested resolution (either is fine, but it needs to be an explicit decision, not silence):** (a) close #253 with a comment pointing to this PR + this review, once Finding 1 is fixed here; or (b) port the `$GC_BEAD_ID`-first idiom onto #253's already-correct exclude-by-id structure and close #189 instead.

### Low-confidence / non-blocking

- The bootstrap-from-empty branch (`elif [ -z "$ASSIGNED_WISP" ]`, both `$CURRENT_WISP` and `$ASSIGNED_WISP` empty — true cold start) also has no test. Lower priority: it's structurally simple and closely mirrors the pre-existing bootstrap fallback. Worth a 3rd/4th case while this file is already open.
- Whether a real (non-mocked) `gc bd mol burn` on an already-burned id errors or no-ops is unverified — the mock always returns 0. Not chasing further since Finding 1's fix removes that code path entirely.

### Verdict: REWORK

Routing back to voxlingo/voxist.executor with these two findings. Full repro commands available on request (ran against a throwaway clone at /tmp, nothing pushed).

bourgois:
Review pass 4 Finding 1 [HIGH] fixed (12ec6fe): `$CURRENT_WISP` is now excluded by id from the `OPEN_WISPS`/`ASSIGNED_WISP` query, so a populated `$GC_BEAD_ID` matching the sole open wisp can no longer be mistaken for an already-queued replacement. Added the reviewer-requested 3rd test case (GC_BEAD_ID populated + equal to the sole open wisp → exactly one pour, exactly one burn of the *old* id) — confirmed red against the pre-fix query, green after. All 3 mocked cases + the existing pack-asset harness pass; TOML parses; rebased on current `main`.

On Finding 2 (the overlap with #253, and aphexcx's 08-01 comment above recommending #253 instead): not resolving that unilaterally from here. Filed voxlingo bead vg-4qv with a side-by-side comparison and the two suggested resolutions (close #253 once this is green, or port this PR's `$GC_BEAD_ID`-first idiom onto #253's structure and close this one instead) for whoever makes that call.

Still blocked on the cross-org Actions-approval gate tracked separately at vg-v2r.

edwarddavison:
Evidence from a production city that may affect this PR's approach — flagging rather than opening a competing report.

`gc bd list` excludes ephemeral beads regardless of `--type`, so the `--status=open --type=molecule` scan this block keeps returns **zero rows even when the wisps exist**. Measured against a live ledger carrying two leaked witness wisps:

```
gci-wisp-bvr  open  molecule  assignee=gas-city/gastown.witness  mol-witness-patrol
gci-wisp-8kq  open  molecule  assignee=gas-city/gastown.witness  mol-witness-patrol

$ gc bd list --assignee=gas-city/gastown.witness --status=open --type=molecule --limit=0 --json
EMPTY (0 rows)          # and also 0 rows with no assignee filter
```

So `OPEN_WISPS`/`ASSIGNED_WISP` resolve empty on every cycle, the surplus loop burns nothing, and the reconcile falls through to a fresh pour — which is how those two wisps accumulated in the first place. The self-exclusion logic here is right, but it is filtering a list that is always empty.

On the invariant in the description — "a patrol wisp stays `status=open` for its entire working life (the bootstrap/resume path only ever sets `assignee`)" — that is accurate today, but only because the pour path never passes `--status=in_progress`. `gc bd update <id> --assignee=X --status=in_progress` does transition it; `--assignee` alone does not; and `--claim` errors `already claimed` without repairing status. Building on the open-status assumption locks in the state that makes the resume check (`status=in_progress`) unable to match its own wisp.

#264 now carries both halves for all three patrol roles — `gc bd query` on `ephemeral=true` for every lookup, `--status=in_progress` on every pour — and carries the exclude-self-by-id insight from here and #253, with credit. If you'd rather land this one, the swap to `gc bd query` is the part worth taking either way. Happy to close ours in favour of whichever you prefer.

PR #195, opened 2026-07-11
feat: add native Beads PostgreSQL and SQLite backend packs

duncan4123:
Publishes registry-discoverable Gas City packs for upstream Beads native PostgreSQL and SQLite backends.

Depends on the native backend provider follow-up stacked above gastownhall/gascity#3980. The PostgreSQL pack uses per-scope schemas; the SQLite pack uses per-scope database files. DoltLite remains on the external plugin bridge.

The postgres registry discovery marker moves from the older plugin pack to bd-native-postgres; old pinned plugin releases remain available. SQLite is registered under backend key sqlite. Registry unit tests pass; the full validator still reports the pre-existing bd-gc-dl source-fragment error.


duncan4123:
Ready for public registry review. This publishes bd-native-postgres and bd-native-sqlite as default-registry beads-backend providers. The postgres discovery marker moves off the legacy plugin pack; its pinned releases remain installable. Merge after gastownhall/gascity#3980.

PR #197, opened 2026-07-11
feat: publish native PostgreSQL and SQLite Beads backend packs

duncan4123:
Publishes two registry-discoverable backend packs for stock upstream bd:

- bd-native-postgres: automatic local role/database/password provisioning by default, explicit URL support for external PostgreSQL, and per-scope schemas
- bd-native-sqlite: per-scope SQLite database files

Depends on gastownhall/gascity#3980. DoltLite remains on the external plugin bridge.

This branch is rebased on current upstream main and changes exactly five files: the two pack manifests, two READMEs, and registry.toml.

Validation:
- registry.toml validator passes
- registry unit tests: 13 passed
- clean registry-driven gc init smoke passes for SQLite
- clean registry-driven gc init smoke passes for PostgreSQL with no URL/password environment; it provisions the local role/database/schema, writes mode-0600 credentials, and supports real gc bd create/list

### DIFF
--- a/bd-native-postgres/README.md
+++ b/bd-native-postgres/README.md
@@ -1,0 +1,32 @@
+# Native PostgreSQL Beads backend
+
+This pack selects the PostgreSQL backend compiled into upstream `bd`. Gas City
+uses one PostgreSQL database connection and derives a separate schema for each
+city or rig bead scope. There is no external Beads backend-plugin process.
+
+Requires a `bd` release with `bd init --backend=postgres` support and a local
+PostgreSQL installation reachable through the current OS user's admin access.
+With no URL configured, `gc init` creates the database, login role, generated
+password, and city schema automatically. Set `GC_POSTGRES_ADMIN_URL` when local
+admin access needs an explicit connection.
+
+```toml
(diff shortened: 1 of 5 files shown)


duncan4123:
we probably dont need to publish this now that beads has these built in.

julianknutsen:
Thanks for this, @duncan4123, and for the follow-up above. Agreed: now that
stock `bd` provides the PostgreSQL and SQLite backends, the provisioning goal
now lives directly in the product. I'm closing this separate-pack PR on that
basis.

If either backend becomes useful as an independent pack again, the supported
route is now https://registry.gascity.com/publish rather than a pointer PR
here. Keep the source in a public repo you own, give each pack a scoped name
such as `duncan4123/bd-native-postgres`, install the Registry GitHub App, then
sign in and use **Find packs**. For a tagged commit, use the GitHub Actions
OIDC workflow in that guide.

Production currently sends every release to staff review; repository proof
clears the ownership gate but is not automatic approval. Only an `approved`
request is live.

Either way, thank you for chasing the backend provisioning story down as far
as you did.

Issue #167, opened 2026-07-06
formula definitions carry deprecated contract = "graph.v2" (69 doctor warnings in every consuming city)


alexputici:
**Gas City version:** gc 1.3.2 (homebrew, darwin/arm64); linked beads library v1.0.4 (via `go version -m`); packs pinned at gascity-packs `0c1fc594`. Where noted, findings re-verified against the unmodified 1.3.3 release binary.
**Environment:** macOS (darwin arm64), multi-rig city (5 rigs + HQ), Dolt-backed beads store, launchd-managed supervisor, custom tmux socket.

- **Where:** gascity-packs monorepo @ 0c1fc594 — 83 files match
  `contract = "graph.v2"` (gascity/formulas/*, compound-engineering/formulas/*,
  bmad/formulas/*, plus docs); the 69 counted by `gc doctor
  formula-requirements` are the formulas this city imports (gastown, gc,
  gstack, superpowers, compound-engineering pins).
- **Effect:** `gc doctor` emits `⚠ formula-requirements — 69 formula
  requirement warning(s)` with hint `replace deprecated contract =
  "graph.v2" with [requires] formula_compiler = ">=2.0.0"; enable [daemon]
  formula_v2 or lower requirements` — in every city consuming these packs.
  Stable count since Phase 1 (P1.3, 2026-07-02) through the 0c1fc594 re-pin.
- **Ask:** migrate the pack formula definitions from `contract = "graph.v2"`
  to `[requires] formula_compiler = ">=2.0.0"` per gc's own deprecation hint.
- **Non-fix locally:** zero city-owned files reference graph.v2 (verified
  P3.7: city.toml/pack.toml/site.toml/orders/local-packs all clean); the
  pinned cache is read-only. Enabling `[daemon] formula_v2` city-side is a
  compiler behavior change requiring its own canaried window — out of
  close-out scope, and pointless while the pack files still say graph.v2.
- **Local disposition:** 69 warnings EXPLAINED in phase 3 close-out
  (phase3-notes.md); doctor acceptance is explained-or-zero.
---
*Filed from a downstream Gas City deployment as part of an operator-approved findings bundle (18 items; cross-references like “ask #N” refer to sibling issues in this batch — index in the first comment once all are filed). Happy to provide deeper traces or re-run repros on request.*

PR #202, opened 2026-07-13
fix(gascity): migrate formulas off deprecated graph.v2 contract

figgeous:
Migrates all 34 formulas in `gascity/formulas` from the deprecated
`contract = "graph.v2"` declaration to the `[requires] formula_compiler = ">=2.0.0"`
form that `gc doctor`'s `formula-requirements` check now expects.

**Why:** every consuming city currently emits a `⚠ formula-requirements` doctor
warning per imported formula. This clears the `gascity/formulas` subset of those.

**Scope:** `gascity/formulas` only — a **partial fix for #167**. The
`compound-engineering/` and `bmad/` formulas still carry `contract = "graph.v2"`
and are intentionally left for a follow-up.

**Verification:** all 34 files parse as valid TOML and assert
`requires.formula_compiler == ">=2.0.0"` with no residual `contract` key. The one
remaining `graph.v2` string in `implement.formula.toml` is prose inside a
description, not a declaration.

---

### Mechanical transform
For each `gascity/formulas/*.formula.toml`:
- removed the top-level `contract = "graph.v2"` line
- inserted a `[requires]` table with `formula_compiler = ">=2.0.0"` immediately
  before the first table section (preserving TOML key/table ordering so no
  subsequent top-level key is captured by the new table)

Also updated the stale global invariant in `gascity/formulas/REQUIREMENTS.md`
("Every formula uses `contract = "graph.v2"`") to describe the new `[requires]` form.

35 files changed (34 formulas + REQUIREMENTS.md), +104 / -35.

Fixes the `gascity/formulas` portion of #167.

### DIFF
--- a/gascity/formulas/REQUIREMENTS.md
+++ b/gascity/formulas/REQUIREMENTS.md
@@ -45,7 +45,8 @@ For every formula change:
 
 ## Global Invariants
 
-- Every formula uses `contract = "graph.v2"`.
+- Every formula declares the formulas v2 contract via `[requires]
+  formula_compiler = ">=2.0.0"`.
 - No formula declares reserved runtime vars `issue`, `bead_id`, or `convoy_id`.
 - Cataloged formulas are user-runnable; internal/base/helper formulas are not
   cataloged unless deliberately promoted.
--- a/gascity/formulas/build-base.formula.toml
+++ b/gascity/formulas/build-base.formula.toml
@@ -27,9 +27,11 @@ Launch inputs:
 """
 formula = "build-base"
 version = 1
-contract = "graph.v2"
 target_required = true
 internal = true
+
+[requires]
+formula_compiler = ">=2.0.0"
 
 [metadata.gc.methodology]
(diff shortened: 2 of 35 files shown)


aphexcx:
From the 2026-08-01 sweep of open PRs: this PR and #220 are **near-identical graph.v2 → `formula_compiler` migrations**. We compared the two diffs directly: the hunks on all 34 formula files shared by both PRs are **byte-identical**; the only difference is that this PR *also* updates `gascity/formulas/REQUIREMENTS.md`, rewriting the global invariant from "Every formula uses `contract = "graph.v2"`" to the `[requires] formula_compiler = ">=2.0.0"` form.

**Recommendation: keep this one (#202) and close #220 as a duplicate.** This PR is a strict superset — without the REQUIREMENTS.md hunk, the documented invariant would be left actively wrong after the migration (still demanding the contract the migration removes). It was also opened first. Both PRs are currently MERGEABLE, so this is purely a pick-one for the maintainers.

PR #203, opened 2026-07-14
fix(ci): restore the supported-pack nightly baseline

julianknutsen:
## Summary

- Align the nightly runner with the supported Beads `v1.1.0`, Dolt `2.1.7`, Go `1.26.5`, and Claude toolchain, with an explicit Beads CLI/module compatibility preflight.
- Resolve moving Gas City refs once per matrix and reuse the immutable snapshot across setup and runtime installation.
- Include closed workflow beads and discover review artifacts across title-suffixed workflow worktrees.
- Pass absolute rig-root subject/report paths to review formulas so retries and isolated worktrees share one stable artifact location.

This repairs the nightly backstop that had been red for an extended period and prevents merges from being judged against a broken baseline.

## Root-cause evidence

- Run 29299747588 proved the workflow itself succeeded, but outer report discovery searched only the rig root while the report lived in a workflow worktree.
- Run 29300440024 exposed a Gas City live-retry worktree metadata bug; the prerequisite fix is gastownhall/gascity#4230.
- Run 29303531414 is the focused `gascity / review` canary against exact commits from this PR and Gas City #4230.

## Testing

- `python3 -m pytest tests contributing/tests gascity/tests discord/tests github/tests slack-full/tests slack-channel/tests pr-pipeline/tests -q`
  - `735 passed, 11082 subtests passed`
- `python3 -m pytest tests/test_gascity_pack_inference_gate.py -q`
  - `45 passed`
- New absolute-path test was observed failing before the implementation and passing after it.
- Full supported-pack nightly matrix will be run before merge.

## Checklist

- [x] No separate issue: this restores the scheduled nightly quality gate.
- [x] Added regression tests for each corrected failure mode.
- [x] No user-facing pack behavior or migration.

PR #204, opened 2026-07-15
refactor(gascity): deduplicate gc role worker prompts, reduce tokens 

sarendipitee:
The exact same gc-role-worker prompt templated was duplicated in 30+ places.

To me this is tremendous risk for drift and makes it harder for any would-be contributors to reason about, so I refactored every place it was used to import a singular copy of the gc-role-worker prompt. Now an edit to this file will propagate everywhere it's used.

Cross-pack internal imports still work since `gc import` caches the whole repo anyways.

Not only was the previous gc-role-worker prompt needlessly wordy. It also had a massive 140 line bash script embedded in it for claiming work.

Instead, the work claim script is now an actual script so you don't have to pay for the input cost of every single fresh session reloading the prompt **and** the output cost of the LLM re-synthesizing the script (and potentially changing something along the way). This should aid in determinism and stability as work flows through complicated formula graphs.

In addition it went from 3183 tokens to a mere 579 tokens, an ~80% reduction while maintaining functionality.

I tested this extensively in my own cities+bins built from source to make sure spawned agents have the prompt and can use it.


julianknutsen:
## Maintainer Adoption Review

**Verdict:** approved after maintainer fixes; ready to merge once required CI is green.

The prompt-centralization and token-reduction direction is sound. Real rendering confirmed the shared fragment resolves for all Gas City roles and the four derived methodology packs, while preserving their non-worker persona content. The original adjacent copies were mostly renderer-local symlinks, so their layout was deliberate compatibility scaffolding rather than unrelated prompt variants.

### Behavior changes found in the contributor patch

- The new command and prompt replaced store-aware `gc bd` calls with bare `bd`, regressing graph/rig store routing and failing the repository guard.
- The wrapper returned `bead_id`, `root_bead_id`, and `continuation_group`, but the prompt preserved only `CLAIMED_BEAD_ID`; unchanged Compound and Superpowers steps still consume `CLAIMED_ROOT_BEAD_ID`.
- The prompt promised terminal handling for malformed/non-zero claims, while `run.sh` retried operational failures forever. Missing identity/Python also stopped drain-acking configuration rejection.
- The compact continuation text dropped the tested fresh-session boundary from PR #203. Empty groups must drain; after a later successful claim, that result is authoritative.
- Tests invoked `run.sh` directly and did not prove city-scope command registration, transitive prompt rendering, or dispatch through `gc bd`.

### Maintainer changes in `d5168f9`

- Restored `gc bd show/update/close` throughout and removed the raw `bd` dependency.
- Restored explicit handoff of all three `CLAIMED_*` identifiers and the non-empty-ID safeguards.
- Split claim and verification into bounded three-attempt phases. Ambiguous post-hook failures return non-zero without mutating claims or acknowledging drain, because core may already have assigned the primary bead and continuation siblings.
- Restored pre-hook configuration drain acknowledgement, surfaced acknowledgement failure, made TERM/HUP/INT actually terminate, and retried incomplete bead records.
- Aligned continuation behavior with PR #203: empty continuation groups drain into a fresh session; non-empty groups may claim again; every successful claim is executed immediately.
- Documented the required city-scoped Gas City import, added `gascity` and `gascity/roles` lint coverage, and added real CLI tests for registration, dispatch, and strict rendering through every derived pack.

### Verification

- Python surfaces: 743 passed, 7 expected skips, 9,412 subtests.
- Gas City focused gate: 177 passed, 9,385 subtests.
- `gc@latest` v1.3.5: all six pack roots lint; 7/7 real command/render integration tests pass.
- `sh -n`, `dash -n`, `bash -n`, ShellCheck, `git diff --check`, registry validation, Gastown shell tests, root Go tests, Slack Go tests, and runtime-cloudflare vet/tests pass.
- Three delegated behavior/parity reviewers completed the final pass with no remaining blockers.

**Review iterations:** 4 (initial review, two maintainer correction rounds, final no-blocker pass).

Contributor commit `4c2e982` is preserved unchanged; the maintainer fixes are a separate commit.

PR #205, opened 2026-07-15
fix(tmux): don't force copy-mode over alternate-screen apps ([0/0] scroll)

outdoorsea:
## Problem

Scrolling the wheel over a full-screen TUI on the **alternate screen** (Claude Code, vim, less) opens an empty copy-mode — the indicator shows `[0/0]` and nothing scrolls. Users read this as "tmux can't scroll back."

## Why

`tmux-keybindings.sh` binds `WheelUpPane` to force copy-mode over any pane not already in copy-mode. That's deliberate (per the comment): so tmux scrollback wins over app-local scrolling on mouse-reporting apps. But the alternate screen has **no tmux scrollback to enter** — the TUI owns the alternate buffer — so forcing copy-mode there just opens an empty `[0/0]`. The binding's own goal ("scrollback wins") can't be met on the alternate screen; the result is "nothing works."

## Fix

Add `#{alternate_on}` to the passthrough condition:

```diff
-if-shell -F -t= "#{pane_in_mode}" "send-keys -M" "copy-mode -e"
+if-shell -F -t= "#{||:#{pane_in_mode},#{alternate_on}}" "send-keys -M" "copy-mode -e"
```

- **Alternate screen** → `send -M`: hand the wheel to the app so it scrolls its own history (Claude Code scrolls its conversation; less/vim scroll).
- **Main screen** (incl. mouse-reporting apps) → `copy-mode -e`: unchanged — the original "scrollback wins" intent is preserved exactly where scrollback actually exists.
- **Already in copy-mode** → `send -M`: unchanged passthrough.

Keying on `alternate_on` (rather than `mouse_any_flag`) is intentional: it targets the actual defect (no scrollback on the alternate screen) without reversing the deliberate main-screen behavior.

## Testing

- `sh -n` clean.
- tmux accepts the format string live (`if-shell -F "#{||:#{pane_in_mode},#{alternate_on}}" ...`).
- Verified on a live Claude Code pane: `alternate_on=1`, so the wheel now routes to the app instead of an empty copy-mode; main-screen shell panes still enter scrollback.

### DIFF
--- a/gastown/assets/scripts/tmux-keybindings.sh
+++ b/gastown/assets/scripts/tmux-keybindings.sh
@@ -29,5 +29,11 @@ fi
 # even over mouse-reporting apps (no mouse_any_flag check) so scrollback wins;
 # once in copy-mode the wheel passes through (-M) for normal scrolling, and -e
 # exits at the bottom. Shift+wheel still does native terminal selection.
-gcmux bind-key -T root WheelUpPane   if-shell -F -t= "#{pane_in_mode}" "send-keys -M" "copy-mode -e"
+#
+# Exception — the alternate screen (#{alternate_on}): a full-screen TUI on the
+# alternate buffer (Claude Code, vim, less) has NO tmux scrollback, so forcing
+# copy-mode there just opens an empty [0/0] — "can't scroll back". Hand the wheel
+# to the app instead so it scrolls its own history; scrollback still wins on the
+# main screen, including mouse-reporting apps.
+gcmux bind-key -T root WheelUpPane   if-shell -F -t= "#{||:#{pane_in_mode},#{alternate_on}}" "send-keys -M" "copy-mode -e"
 gcmux bind-key -T root WheelDownPane send-keys -M


outdoorsea:
@julianknutsen — could I get a review on this when you have a moment?

It's been open since Jul 15 with no reviewer assigned. I only have pull access on this repo, so I can't request one or merge it myself — flagging directly since I think it just never surfaced rather than anyone having passed on it.

Small change: +7/−1 in `gastown/assets/scripts/tmux-keybindings.sh`. The current binding forces `copy-mode -e` on `WheelUpPane` unconditionally. On the alternate screen there's no tmux scrollback (`history_size=0` against `history-limit 50000`), so full-screen TUIs — Claude Code, vim, less — open an empty `[0/0]` and the user can't scroll at all. Adding the `#{alternate_on}` test hands the wheel to the app in that case, while scrollback still wins on the main screen, including for mouse-reporting apps.

Worth noting the blast radius: tmux key bindings are server-global, so one agent session running `tmux-keybindings.sh` clobbers the wheel in *every* pane on the gc server, including panes that already existed. Downstream this has cost a couple of debugging sessions, and the only local workaround is runtime hooks that die on a tmux-server restart.

No failing checks, not a draft, currently MERGEABLE. Happy to adjust the predicate if you'd prefer something narrower.

aphexcx:
From the 2026-08-01 sweep of open PRs: this PR and #136 patch the **same `WheelUpPane` binding** in `gastown/assets/scripts/tmux-keybindings.sh` for the same bug (copy-mode forced over alternate-screen apps → empty `[0/0]` scrollback).

**Recommendation: keep this one (#205).** It is the surgical fix — it adds only `#{alternate_on}` to the pass-through condition (`#{||:#{pane_in_mode},#{alternate_on}}`), preserving the binding's documented design that scrollback wins over mouse-reporting apps on the main screen. #136 fixes the same bug but also re-adds a `#{mouse_any_flag}` term, which reverts that deliberate choice (the original comment explicitly says "no mouse_any_flag check … so scrollback wins"). Both PRs are MERGEABLE, so this is purely a pick-one for the maintainers.

PR #213, opened 2026-07-17
slack-full: inbound durability (persist-and-retry) + reliability hardening

aphexcx:
## Summary

Hardens slack-full's inbound path, which is currently **at-most-once**: the adapter acks Slack with 200 before any processing (deliberately suppressing Slack's own redelivery), then forwards to gc with a single unretried HTTP POST — a gc hiccup, dispatch-semaphore saturation, or shutdown loses the message silently. This series makes inbound delivery durable and ports four smaller reliability behaviors proven in production on a private Slack daemon.

Six commits, each self-contained:

1. **Persist-and-retry for inbound forwards** — spool each event (tmp+fsync+rename, 0600/0700) before the first forward attempt; bounded retries (5s/15s/30s/60s, 5 attempts); dead-letter directory on exhaustion with a grep-stable `inbound POST failed` log line; startup replay with corrupt-entry quarantine. `INBOUND_SPOOL_DIR` override, default under `GC_SERVICE_STATE_ROOT/data`. Replay is idempotent via the existing extmsg DedupKey.
2. **Echo-loop guard** — drop inbound messages containing the literal `*Sent using* <@` signature emitted by user-token Slack MCP integrations; without this, a bound channel shared with such an integration can loop. (Observed live in our workspace.)
3. **Attachment path presentation guard** — the inbound-file header now instructs agents to read `file://` paths only when relevant and never paste them into replies; bare image paths cause coding-agent auto-attach loops that wedge sessions with provider 400s.
4. **Fail-closed DM membership gate** — verify bot membership via `conversations.info` before processing `message.im` (Slack Assistant-mode surfaces can deliver DM events for conversations the bot is not in); 5min/30s positive/negative cache; API errors fail closed and are not cached.
5. **Busy-reaction lifecycle** — add a busy reaction (`BUSY_REACTION`, default hourglass) on targeted inbound dispatch and remove it when the session's reply publishes to that thread; adds `reactions.remove` support and `remove:true` on `/react`; replaces the previous add-only eyes reaction.

6. **Portable `os.Root` confined open (darwin build fix)** — `openBeneath`'s hand-rolled `openat(2)` walk used raw `syscall.Openat`, which doesn't exist on darwin, so the adapter package failed to build (and its tests never ran) on macOS — upstream `main` has the same breakage. Replaced with stdlib `os.Root` (Go 1.24+): `OpenRoot` pins the root fd and `OpenFile` resolves every component beneath it — the portable equivalent of `openat2(RESOLVE_BENEATH)` — with `O_NOFOLLOW` kept on the leaf for race-window inode-swap detection. Semantics delta: an intermediate symlink confined to the root is now followed rather than ELOOP-rejected; anything resolving outside the root still fails.

## Testing

- `go vet ./... && go test -race -count=1 ./...` in `slack-full/adapter` (golang:1.25 container): all pass.
- With commit 6 the adapter now also builds and tests green natively on darwin/arm64 (`go build && go vet && go test ./...`, go1.26.4).
- `python3 -m pytest slack-full/tests/`: 86 passed.
- CHANGELOG updated under [Unreleased].

Note: a pre-existing `-race` flake on main (`TestThreadContext_*` reading `slackAPIBase` after cleanup) reproduces at the base commit and is unrelated; happy to file/fix separately.


aphexcx:
**Ready to merge, pending maintainer action** (gp-l7d, per Afik's 2026-08-01 approval: "fix codex recommendations … merge them when green").

Four codex review rounds run over the full diff vs main; every P1/P2 fixed except one documented tradeoff:
- **im:read** added to manifest + OAuth installer scopes; DM-gate `missing_scope` drops named and fixed; concurrent DM probes coalesced
- `openBeneath` rejects symlinked upload roots **and** leaf symlinks via Lstat+SameFile identity checks (os.Root follows both otherwise — verified by probe test)
- Spool entry now survives until the alias dispatch confirms; startup replay redoes the targeted dispatch through a bounded worker pool; dir-fsync after rename; `INBOUND_SPOOL_DIR=` honored as opt-out; dispatch slots released only for durably-spooled retries
- Busy-reaction races closed: mark-before-forward with cancel, sibling thread keys, deferred removal behind in-flight adds, displaced-mark cleanup on all paths, atomic redelivery admission (5-min cleared tombstones)
- Company path: DM membership gate (owner-token conversations.info probe; not-a-member terminalizes, transport errors park), MCP-echo drop before every company admission branch
- **Documented tradeoff, not fixed**: forwards are at-least-once until gc core consumes `dedup_key` (docs/phase5-ledger-readiness.md) — the alternative is losing acked Slack messages

Verification: gofmt/vet clean, `go test ./...` + `-race` green, 383 pytests green.

**Blocked on:** CI workflow approval (all runs `action_required`; aphexcx lacks rights) and branch-protection merge (base policy prohibits; needs a gastownhall maintainer). Note: #247/#248 touch the same slack-full files — whichever of this PR and that stack merges second needs a rebase pass.

Issue #192, opened 2026-07-10
gastown refinery Startup never reconciles OPEN mol-refinery-patrol wisps (+1 open wisp leak per restart; witness has the correct parity)


superlzyguy:
The gastown **refinery** role's Startup protocol never reconciles OPEN `mol-refinery-patrol` wisps, so each refinery restart leaks +1 open wisp. The **witness** role already has the correct behavior.

## Symptom

Observed on a live city (2026-07-08): a rig's `gastown.refinery` accumulated 4 OPEN `mol-refinery-patrol` wisps alongside its fresh `in_progress` wisp — one leaked per restart. The witness patrol flagged it and nudged the refinery to burn the surplus (manual mitigation).

## Root cause (confirmed 2026-07-09)

- The **witness** Startup Protocol reconciles BOTH `in_progress` AND `open` molecule wisps and burns surplus.
- The **refinery** Startup only queries `{{.AssignedInProgressQuery}}` (in_progress-only), so the open successor wisp poured by next-iteration is never reconciled → +1 open leak per restart.

## Fix

Make the refinery `prompt.template.md` Startup reconcile open wisps with witness-parity (reconcile open + in_progress, burn surplus).

The fix was authored and validated against a scratch copy of the pack (fails-before/passes-after), but the reconcile logic lives in this repo's pack templates, which are consumed read-only downstream — so it can't be landed from the consuming side. A guard test that enforces witness-parity once the pack ships the fix (skipped until then) is preserved at `superlzyguy/gascity` branch [`fix/su-8w67m-refinery-wisp-reconcile`](https://github.com/superlzyguy/gascity/tree/fix/su-8w67m-refinery-wisp-reconcile) (`examples/gastown/gastown_test.go`).

rbriski:
Additional live evidence from 2026-07-20 supports the underlying ephemeral-visibility problem:

A running witness restart had a queued open patrol wisp, but both bd list reconciliation queries returned empty because ephemeral molecules are not surfaced. Startup and next-iteration then poured two surplus wisps. The ephemeral-aware query found all three and allowed manual reconciliation.

This latest observation is witness-specific, so it does not replace the refinery reproduction documented here. It does show that copying the current witness shape without fixing storage visibility would remain insufficient.

[...]

Issue #145, opened 2026-06-28
Patrol molecule wisps accumulate: deacon/refinery startup re-pours a duplicate successor (missing open-check + over-broad molecule selector)


iwata-1116:
## Summary

Patrol molecules (`mol-deacon-patrol`, `mol-refinery-patrol`, `mol-witness-patrol`) accumulate **duplicate open wisps** over time. Each patrol loop is meant to keep exactly one live wisp (pour successor → burn current), but on a fresh session restart the agent re-pours a duplicate successor instead of adopting the open one already on its hook.

Live evidence (a running gastown city on an older pack pin): the **deacon held 2 open patrol wisps simultaneously** (both `issue_type=molecule`, assigned to `gastown.deacon`, neither ever `in_progress`), while the **witness held 1** — witness already carries the relevant fix, deacon and refinery were left behind.

## Root cause

Two compounding defects in the deacon/refinery startup + loop logic:

1. **Startup only checks `in_progress` work.** The Startup Protocol resolves assigned work via an in-progress-only query (`AssignedInProgressQuery`). A patrol loop's `next-iteration` pours its successor as an **OPEN** wisp. On a fresh restart the in-progress check returns nothing, so the agent pours a *second* successor — the original open wisp is orphaned and leaks. (This is the dominant mechanism; a restart in the pour/burn window is a secondary latent path.)

2. **Molecule selector is too broad.** `#121` (1407301, port of gascity#3416) correctly fixed the dead `--type=wisp` filter (patrol wisps are `issue_type=molecule`), but it queries bare `--type=molecule` + assignee. That matches *any* molecule assigned to the agent, not just this patrol loop. There is no first-class patrol discriminator (`--mol-type=patrol` resolves null on root-only wisps); `title` (= formula name) is the only reliable discriminator.

`#121` is necessary but not sufficient: it makes the queries see molecules, but does not add the startup open-check, and its selector is not loop-scoped.

## Proposed fix (staged)

**Stage 1 — deacon, minimal & non-destructive (stop new accumulation):**
- Add a startup **open-check**: query `--status=open,in_progress --type=molecule --title="mol-deacon-patrol"` and gate the pour on it being empty — so a fresh restart adopts the existing open successor instead of pouring a duplicate.
- **Title-narrow** every patrol wisp query (`--title="mol-<role>-patrol"` + exact-match in jq) so they match only this loop, never unrelated molecule work.
[...]

PR #215, opened 2026-07-18
fix(gastown): use ephemeral-aware bd query for wisp reconciliation

rbriski:
## Summary
- The witness/refinery/deacon patrol recipes (prompt.template.md + mol-witness-patrol.toml, mol-refinery-patrol.toml, mol-deacon-patrol.toml) reconciled queued patrol wisps via `gc bd list --type=molecule` variants.
- `bd list` never returns ephemeral beads (even with `--all`), so the recipes were blind to the wisps they were meant to reuse: a witness/refinery/deacon restart poured a duplicate wisp and leaked one per restart (boombox repro bb-wisp-2ii; gastown ledger showed 5 leaked wisps: ga-wisp-xoq/8wh/mky, ga-wisp-iyf/kmq).
- Replaced every wisp-reconciliation `bd list --type=molecule ...` call site with a bounded `gc bd query "type=molecule AND ephemeral=true AND status=... AND assignee=..." -n <limit>`, which does see ephemeral beads.
- Scope is limited to wisp-reconciliation call sites; generic orphan/session-liveness `bd list` scans elsewhere in these recipes are unaffected (unrelated to the ephemeral-wisp bug).

Tracked by gastown/gascity ga-2n1 (split from ga-u0o).

## Test plan
- [x] `PATH=<py3.11+>:$PATH bash gastown/tests/test_gastown_pack_assets.sh` passes
- [x] `PATH=<py3.11+>:$PATH bash gastown/tests/test_gastown_theme_scripts.sh` passes
- [x] All three formula TOML files parse (`tomli.load`)
- [x] Verified `gc bd query "type=molecule AND ephemeral=true AND assignee=\"...\"" --json` returns ephemeral wisps that `gc bd list --type=molecule` cannot see

### DIFF
--- a/gastown/agents/deacon/prompt.template.md
+++ b/gastown/agents/deacon/prompt.template.md
@@ -89,9 +89,9 @@ If `next-iteration` already ran, do not pour again; run `gc hook`.
 ```bash
 CURRENT_WISP=${GC_BEAD_ID:-}
 if [ -z "$CURRENT_WISP" ]; then
-  CURRENT_WISP=$(gc bd list --assignee="$GC_AGENT" --status=in_progress --type=molecule --limit=1 --json | jq -r '.[0].id // empty')
+  CURRENT_WISP=$(gc bd query "type=molecule AND ephemeral=true AND status=in_progress AND assignee=\"$GC_AGENT\"" -n 1 --json | jq -r '.[0].id // empty')
 fi
-ASSIGNED_WISP=$(gc bd list --assignee="$GC_AGENT" --status=open --type=molecule --limit=1 --json | jq -r '.[0].id // empty')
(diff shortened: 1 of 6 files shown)


rbriski:
Closing this PR because I opened it before auditing the existing issue and PR history.

The same problem is already tracked by #145 and #192, with overlapping implementations in #146, #177, and #189, building on merged #121. This patch also needs further design work before it would be safe upstream: patrol queries should be title-scoped, destructive reconciliation needs deterministic and revalidated selection, refinery open-successor behavior remains incomplete, and focused regression coverage plus the repository validation gates are missing.

I am moving the fresh production evidence to #145 and #192 and carrying an urgent mitigation locally. We will not open a replacement PR unless maintainers request a specific consolidated approach. The fork branch is being preserved only as evidence and local patch material.

PR #216, opened 2026-07-18
fix(gascity): make worker claims binding-independent

sarendipitee:
## Why

Gas City pack commands are invoked as `gc <binding> <command>`. Import bindings are deployment configuration, so worker prompts that hard-code `gc gc claim` only work when the city-scoped Gas City pack happens to be imported as `gc`. Derived packs commonly use other top-level bindings while importing `gascity/roles` on a rig, making claim correctness depend on an unrelated alias choice.

Native `gc hook --claim --drain-ack --json` is the binding-independent claim boundary. It already owns session identity, federated store selection, atomic claim handling, continuation preassignment, and drain acknowledgement. Source PR https://github.com/gastownhall/gascity/pull/4415 adds the two workflow-context fields these prompts consume: `root_bead_id` and `continuation_group`.

## What changed

- Call native `gc hook --claim --drain-ack --json` for initial and continuation claims.
- Assert rendered workers from Gas City and all derived packs contain the native command and no `gc gc claim` dependency.
- Update setup docs to separate required formula/skill imports from claim mechanics.
- Retain the city-scoped `claim` command as a compatibility wrapper; this PR removes it from workflow correctness, not from the public surface.

This is intentionally dependent on gascity PR #4415. Merge source first, then this pack change.

All 61 current open pack PR heads were fetched once and searched; none removes the binding-specific worker claim. PR #203 also edits the shared worker fragment to clarify exact claimed identifiers, but `git merge-tree` combines both changes without conflict.

## Validation

- `741 passed, 7 skipped, 9412 subtests passed`
- `GC_TEST_BIN=/var/tmp/gc-hook-context-4415 ... tests/test_gc_role_prompt_integration.py`: 7 passed using exact source commit `40f9cbc74`
- `go test ./...`
- `make registry-format-validate`
- `GC=/var/tmp/gc-hook-context-4415 make registry-validate`

### DIFF
--- a/gascity/README.md
+++ b/gascity/README.md
@@ -21,7 +21,7 @@ Prerequisites: Gas City installed and a city running (`gc init`, `gc start`),
 and your project added as a rig (`gc rig add .` inside the repo). See the
 [repository README](../README.md) for the from-scratch path.
 
-1. Import formulas, claim command, and rig roles. From city directory:
+1. Import formulas, coordinator skill, and rig roles. From city directory:
 
    ```sh
    gc import add --name gc https://github.com/gastownhall/gascity-packs.git//gascity
@@ -37,10 +37,11 @@ and your project added as a rig (`gc rig add .` inside the repo). See the
    source = "https://github.com/gastownhall/gascity-packs.git//gascity/roles"
    ```
 
-   Both imports are required. The rig-scoped roles pack supplies agents but,
-   by design, rig imports do not register city commands; importing roles alone
-   renders prompts that reference `gc gc claim` without installing that
-   command. Keep the top-level Gas City pack imported at city scope.
+   Both imports are required for the full build pack. The rig-scoped roles pack
+   supplies agents; worker prompts claim through native
+   `gc hook --claim --drain-ack --json`, independent of the top-level pack's
+   import binding. Keep the top-level Gas City pack imported at city scope for
+   formulas and the coordinator skill.
 
    (Contributors hacking on packs can point this source at a local clone.)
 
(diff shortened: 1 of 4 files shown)


SamuelCook:
We hit this in production and would very much like to see #216 land.

**We're an affected city.** Our Gas City deployment imports the pack as `gascity` rather than `gc`, so the hardcoded `gc gc claim` in `gascity/template-fragments/gc-role-worker.template.md` renders as a command that doesn't exist. Every worker role's documented *first action* exits non-zero with `unknown command "gc"`,  and because the same fragment tells the agent not to go looking for work after a failed claim, an obedient session then does nothing at all. It fails closed and silently, which is why it took us a while to see it.

This affected 24 of 24 `gc.*` roles across both our rigs.

**On why this probably hasn't bitten you:** `gascity/roles/pack.toml` binds the parent pack as `[imports.gc]`, so within this repo the hardcoded string is accidentally correct. The bug only appears from outside, under a different import  binding.

gastownhall/gascity#4508 ("bind the Gas City pack as `gc`", merged 2026-07-23) fixes this going forward by making `gc init` emit `gc` as the import key - and quad341's approval there names this exact failure mode. But that fix is only reachable by *new* cities. Ours was generated by a `gc` predating it, and there's no migration path; we can't rename the binding on a live city without breaking everything keyed to it. **#216 is the only change that repairs cities that already exist.**

#216 declared itself dependent on gastownhall/gascity#4415 (`feat(hook): expose workflow context in claim results`), which merged on 2026-07-20. As far as I can tell that's been the only thing holding it, so it's been mergeable for about a month.

Our `gc` build already accepts `gc hook --claim --drain-ack --json` with all three flags, and its `--json-schema result` already carries `root_bead_id` and `continuation_group`, so #4415 is in our binary and #216 would work here as written, with no binary upgrade needed on our side.

PR #219, opened 2026-07-21
feat(gastown/boot): gate reasoning triage behind a deterministic stuck-probe

atbrace:
## Problem

`gastown/boot` (the deacon watchdog) is a `mode="always"`, `wake_mode="fresh"` named session whose agent does **one triage pass then exits**. Because it's `always`-mode, the controller respawns it immediately on exit → a fresh full-reasoning session every **~80–90s**.

On a healthy town that's pure churn. A 24h sample on a live city logged:
- **134 boot passes**, **601 read-only `gc` calls**, **ZERO interventions** — every pass re-confirmed the same idle deacon state.
- Entirely automated burn; it was the **3rd-largest token consumer** in the city.

The deacon it watches patrols on a **~10-minute** cadence, so a per-90s reasoning watchdog is **~7× oversampled**.

## Fix

Gate the expensive reasoning pass behind a **cheap deterministic Step-0 probe** that runs first on every wake and escalates to the existing observe/decide/act logic **only** when a real stuck-signal trips:

- error markers in the deacon pane (`error|panic|traceback|exception|rate-limit|permission denied|fatal|command not found`), **or**
- unread mail for the deacon (`> 0`), **or**
- an in-progress patrol wisp older than **20m** (> the deacon's 300s backoff cap + a full ~10m patrol).

`PROBE: CLEAR` → `drain-ack` + exit, **no reasoning**. `PROBE: STUCK` → fall through to the (unchanged) triage / nudge / warrant path.

**Coverage is preserved** — the same signals are evaluated every tick, so a genuinely stuck deacon still escalates within one cadence interval. Absent/young wisps remain legitimate idle (not a stuck signal), matching the prior guidance.

## Why gate the pass instead of slowing the cadence

`sleep_after_idle` is rejected on `mode="always"` sessions, and converting boot to `on_demand` risks the watchdog not spawning (a monitoring gap). Gating the *reasoning* makes the frequent cadence cheap, which is the safe lever and needs no lifecycle/config change.

## Verification

Applied via a local `prompt_template` override on a running city:
- **Live:** gated `CLEAR` wakes drop from a multi-call reasoning triage to a **2-call** `probe → drain-ack → exit` with **zero reasoning text** (confirmed in boot session transcripts); `session.stopped` reason stays `drain acknowledged`.
- **STUCK paths unit-verified:** a 40-min stale wisp (2400s) trips `stale-wisp`; `rate limit` / `Traceback` trip `pane-errors`; benign pane output does not false-trip.

All Go template variables (`{{ cmd }}`, `{{ .BindingPrefix }}`, `{{ template "architecture" . }}`, `{{ .WorkDir }}`) are preserved.

### DIFF
--- a/gastown/agents/boot/prompt.template.md
+++ b/gastown/agents/boot/prompt.template.md
@@ -17,47 +17,98 @@ from wisps, pane output, and mail.
 gives each wake a new provider context. Observe, decide, act, drain-ack, exit.
 Do not rely on prior conversation context or handoff mail. Narrow scope keeps each wake cheap.
 
+**Cost discipline:** the deacon patrols on a ~10-minute cadence and
+is healthy on the overwhelming majority of wakes. A full reasoning triage every
+wake is pure churn (measured: 134 passes/24h, 601 read-only gc calls, ZERO
+interventions). So each wake runs a **cheap deterministic stuck-probe FIRST**
+and does the expensive reasoning triage **only when a stuck-signal actually
+trips**. On a clear probe you exit immediately — no further tool calls, no
+reasoning.
+
 ---
 
-## Triage Steps
+## Step 0 — Deterministic stuck-probe (ALWAYS run this FIRST)
 
-### Step 1: Check if deacon session exists
-
-```bash
-{{ cmd }} session peek {{ .BindingPrefix }}deacon --lines 1
-```
-
-If the deacon session does not exist, drain-ack and exit. The controller will
-restart dead agents.
-
-### Step 2: Observe deacon state
+Run this single block verbatim. It emits exactly one line: `PROBE: CLEAR ...`
+or `PROBE: STUCK ...`. Do not reason about deacon health before running it, and
+do not run any other command first.
 
 ```bash
-# Recent pane output — is the deacon actively working?
-{{ cmd }} session peek {{ .BindingPrefix }}deacon --lines 30
+{{ cmd }} runtime heartbeat 2>/dev/null || true
+STALE_WISP_SECS=1200   # 20m: > deacon's 300s backoff cap + a full ~10m patrol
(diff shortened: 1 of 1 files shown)


jm2:
Applied this on a live city to confirm it before merge. The deterministic gate
does what it claims, and doing so exposes a second effect worth naming in the
PR.

**The reasoning gate works.** After the change, a CLEAR wake costs roughly 663
output tokens, consistent with `probe -> drain-ack -> exit` and no reasoning
text. Our sample includes a few wakes that ran the ungated prompt, so that is a
slight over-estimate of the gated cost.

**But it raises the wake rate.** We measured about 24 Boot starts in 10 minutes,
or one every 25 seconds, versus the roughly 80–90 seconds reported for the
ungated prompt. Live event reconstruction later confirmed 25 starts in 9m50s.

That is structural: with `mode="always"` plus a prompt that drain-acks and
exits, the controller immediately reasserts the named session. Making each pass
cheaper also makes each generation finish and respawn sooner. This PR therefore
reduces cost per pass, but does not fix lifecycle cadence or the provider-start
storm.

**Correction to my original conclusion:** a new core lifecycle knob is not
currently supported by the evidence. I had overlooked existing pack PR #154,
which already changes Boot to `on_demand` and adds a scheduled cooldown order
to wake it. That is the right existing mechanism to evaluate. The remaining
work is to rebase/harden that approach and dynamically prove one generation per
due tick, zero respawns between ticks, no duplicate queued nudges under repeated
or overlapping order runs, and a second clean generation on the next interval.

So #219 remains useful and complementary as a cheaper triage pass, but it
should not be described as fixing cadence. I would keep Boot suspended on an
affected city until a #154-style candidate passes a measured multi-interval
canary.

PR #222, opened 2026-07-22
fix(gastown): fail closed on polecat claim failures

jm2:
## Problem

Gastown's polecat startup claim block converted two claim-system failures into ordinary idle/drain behavior:

1. After exhausted or malformed `gc hook --claim` attempts, it drain-acked and exited successfully.
2. After a hook reported a successful claim, a disagreeing `gc bd show` caused the prompt to release/reopen the bead, drain-ack, and exit.

Both paths hide infrastructure failures as no-work. Routed demand then wakes or recreates the polecat, producing a wake/claim/drain loop with no useful progress.

The second path was reproduced in a live v1.3.5 canary. `gc hook --claim` discovered canonical `ac-3iv` in the city store but, because v1.3.5 discarded the selected federated-store context, applied the mutation to a same-ID row in the rig store. The normal work-context read correctly still saw the canonical bead as `open` and unassigned. The old prompt then mutated/released through an ambiguous context and drain-acked. Five provider generations consumed roughly 70k input tokens in about two minutes without starting the task.

The selected-store bug is a Gas City core issue; upstream main fixed it in gastownhall/gascity#3785. This pack change is the required fail-closed boundary and does not pretend to replace that core fix.

## Fix

- Validate the schema-v1 `gc hook --claim --json` envelope and work receipt (`bead_id`, `assignee`, and route).
- Allow only schema-v1 `action=drain, reason=no_work` to take the terminal idle/drain path.
- Treat exhausted, malformed, identity-mismatched, and other non-idle hook results as `CLAIM_INFRA_FAILURE`: send one Witness escalation, exit nonzero, and do not drain-ack.
- Require the direct work-context read to confirm `in_progress` plus the hook receipt's assignee before emitting `CLAIMED_BEAD_ID` or starting work.
- If the hook receipt and work read disagree, send one high-priority Witness escalation and exit nonzero. Do not release, reopen, mutate, drain-ack, or begin work through an ambiguous store context.
- Stamp optional `polecat_session` observability metadata only after confirmation.
- Missing runtime identity also fails visibly without claiming or draining.

This makes the pack safe on both sides of the core fix: pre-fix, wrong-store claims stop visibly instead of churning or doing work against the wrong row; post-fix, a correctly selected claim proceeds only after its normal work view confirms ownership.

## Regression coverage

The executable startup-block regression models the production `ac-3iv` disagreement:

- hook returns a schema-valid work receipt;
- repeated direct work reads remain `open` and unassigned;
- the block must exit 1, emit `CLAIM_INFRA_FAILURE`, send exactly one Witness escalation, issue no `bd update`, issue no `runtime drain-ack`, and emit no `CLAIMED_BEAD_ID`.

A structured non-idle/invalid hook result is also pinned to the visible failure path. The original hook-infrastructure escalation contract remains covered.

## Validation

- `bash gastown/tests/test_gastown_pack_assets.sh`
- `bash gastown/tests/test_gastown_theme_scripts.sh`
- `bash gastown/tests/test_polecat_churn_watcher.sh`
- `go test .`
- `gc lint gastown`
- `git diff --check`

All pass locally.

### DIFF
--- a/gastown/agents/polecat/prompt.template.md
+++ b/gastown/agents/polecat/prompt.template.md
@@ -195,9 +195,17 @@ while [ "$CLAIM_TRY" -lt 3 ]; do
   sleep 2
 done
 if [ -z "$WORK_ID" ]; then
-  echo "CLAIM_REJECTED gc hook --claim returned no workable bead after retries"
-  gc runtime drain-ack
-  exit 0
+  # Hook infrastructure failure, NOT "no work". Do NOT drain-ack: drain-ack
+  # reports idle and puts the session to sleep, which hides the outage from
+  # town oversight and loops wake->drain->sleep forever (a polecat that never
+  # claims leaves no stale bead, so no health check ever fires). Escalate and
+  # exit non-zero WITHOUT acknowledging drain, so the failure stays visible
+  # and the controller's crash-loop path engages instead of the sleep cycle.
+  echo "CLAIM_INFRA_FAILURE gc hook --claim failed after $CLAIM_TRY attempts (last code=$CLAIM_CODE): ${CLAIM_ERR_TEXT:-malformed claim result}"
+  WITNESS_TARGET="${GC_RIG:+$GC_RIG/}{{ .BindingPrefix }}witness"
+  gc mail send "$WITNESS_TARGET" -s "ESCALATION: gc hook --claim failing [HIGH]" \
+    -m "gc hook --claim failed $CLAIM_TRY times in a row (last exit=$CLAIM_CODE): ${CLAIM_ERR_TEXT:-malformed claim result}. Session is NOT drain-acking; this is an infrastructure fault (agent resolution, daemon, or store), not a work shortage."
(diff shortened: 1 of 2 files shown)


jm2:
Withdrawing after checking this prompt against the complete current
`gc hook --claim --json` contract.

Two valid core outcomes are handled incorrectly:

- `action=work, reason=ready_assignment` represents already-assigned work that
  may legitimately remain `status=open`; this prompt requires `in_progress`
  and rejects it.
- `action=drain, reason=stale_session` is an intentional terminal drain. Core
  explicitly drain-acks it so a stale provider does not retry forever; this
  prompt treats it as infrastructure failure and retries/escalates.

The result can refuse valid preassigned work and turn a stale-session stop into
the retry loop this PR was meant to prevent. The tests do not cover either
current-core outcome.

The underlying fail-closed goal remains valid, but the replacement must be
designed and tested against every supported hook result before it is proposed
again.

PR #226, opened 2026-07-23
fix(gastown): reconcile patrol wisps across restarts

jm2:
## Problem

Deacon, Witness, and Refinery patrol roots are ephemeral molecule rows, but
their startup and handoff logic searched durable or only `in_progress` work.
After a persistent-provider restart with an empty `GC_BEAD_ID`, an existing
open patrol root could therefore be missed and another root poured.

There is a second restart window: `gc bd mol wisp` commits the new root before
the following assignment command. A crash between those commands leaves an
unassigned open root. Looking only for the canonical assignee also misses that
root and pours another one on restart.

## Change

A shared patrol-ledger contract now:

- queries both open and in-progress roots from the ephemeral tier;
- filters the exact formula and scoped role identity;
- also recognizes an exact-formula unassigned `open` root as a recoverable
  post-pour/pre-assignment result;
- prefers the explicit current ID, then assigned in-progress/open roots, then
  the oldest unassigned open root;
- re-reads the selected root before assigning or promoting it;
- reuses one queued successor instead of pouring another; and
- burns only surplus roots that still read as open, ephemeral, empty,
  exact-formula molecules assigned to this role or still unassigned.

Active, materialized, foreign, or changed roots are preserved for inspection.
Query, identity, parse, and revalidation failures stop without draining or
mutating patrol state.

The same contract is used by Deacon, Witness, and Refinery startup and
next-iteration handoff.

## Regression coverage

`gastown/tests/test_patrol_wisp_reconciliation.sh` executes the rendered
startup and handoff blocks for all three roles. It covers:

- restart with only open assigned roots and an empty `GC_BEAD_ID`;
- adoption of an unassigned post-pour root at startup;
- adoption of an unassigned queued successor at handoff;
- repeated same-provider handoffs;
- deterministic survivor selection and conservative surplus cleanup;
- query failure and empty identity;
- failed validation of active/materialized roots; and
- preservation of provider state on every failure.

## Validation

- all `gastown/tests/test_*.sh` suites
- `python3 -m pytest ...`: 1,170 passed, 8 skipped, 9,412 subtests passed
- `gc lint gastown`
- `go test .`
- Bash syntax over tracked shell scripts
- `git diff --check`

### DIFF
--- a/gastown/agents/deacon/prompt.template.md
+++ b/gastown/agents/deacon/prompt.template.md
@@ -10,6 +10,10 @@
 
 ---
 
+{{ template "patrol-wisp-ledger" . }}
+
+---
+
 ## Your Role: DEACON (Town-Wide Coordination for {{ .CityRoot }})
 
 You are the controller's judgment layer for periodic, cross-rig, and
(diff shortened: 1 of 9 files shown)


edwarddavison:
Flagging overlap so you don't review the same block three times: this PR, #189, #253 and #264 all rewrite the patrol wisp reconcile.

This one is the most thorough — it covers all three roles' formulas *and* prompt templates, queries both open and in-progress ephemeral rows, filters exact formula and role identity, and re-reads before assigning. From a production deployment carrying this defect live, the two mechanisms it fixes are both real and independently sufficient to leak:

1. pour + `--assignee` alone leaves the wisp at `status=open`, invisible to the `status=in_progress` resume check;
2. `gc bd list` excludes ephemeral beads regardless of `--type` — measured, it returns 0 rows against a ledger holding two leaked witness wisps that `gc bd query 'ephemeral=true'` returns fine.

**If this merges, #264 should be closed in its favour** — ours is the surgical version of the same fix and has no coverage this lacks. I've kept it open only because this has been unreviewed since 23 July and the defect is costing us wisp leaks in two rigs today. Either outcome works for us; landing one of them is what matters.

PR #231, opened 2026-07-24
fix(gastown): require a recorded reason before a polecat blocks work

jm2:
## Problem

A Gastown polecat could set a work bead to a terminal status with no reason recorded anywhere, while keeping the claim.

Observed live on 2026-07-24 ~16:45 EDT, minutes after routing seven beads and scaling a pool 1 → 8. Worker `gastown.cheedo` (`ac-nkk`) claimed `ki-0aq` ("ABI: SP int-truncation removal") and `ki-gu2` ("Integration hygiene: upstream re-audit"), set **both** to `status=blocked`, and kept both claims. It recorded nothing: empty `notes`, no metadata, no mail. A human had to guess why two ready, unblocked, polecat-sized beads had stopped.

The pack already had the answer to half of this.

The refinery never rejects work silently. It writes `metadata.rejection_reason`, and that record is what drives the normal refinery → polecat correction cycle — `mol-polecat-work`'s `workspace-setup` reads it, resumes the existing branch instead of redoing the work, and clears it. Its two block helpers go further: `blocked_reason` plus a structured result key, `--assignee=""`, and `gc.routed_to=human`. Real record from the affected city:

```
rejection_reason: "Acceptance unmet: candidate has no production route
                   call site; legacy db_load path remains the only path"
recovery.disposition: "production-wiring-rework-required"
```

The polecat had no equivalent. Its entire blocked-path guidance was one unactionable table row in `mol-polecat-work`:

```
| Blocked on external | Mail Witness, mark yourself stuck |
```

and one prompt line offering a bare `gc bd update <bead> --status=escalated`. **There is no `--status=blocked` anywhere in the polecat's assets** — the transition was completely unconstrained, so the agent invented it. `--status=escalated` is the documented sibling and has the same hole, so both tokens are covered here.

### Second-order effect (separate fix)

Because the beads stayed `gc.routed_to=<pool>` while being unclaimable, route discovery kept counting them as demand while the claim path kept refusing them. The controller respawned the worker every 20–40 seconds — spawn → prime → `gc hook --claim` returns nothing → drain-ack → respawn — reaching **15 provider starts and 3 `NO_ROUTED_WORK` in a 10-minute window** and still climbing.

Making route discovery skip non-claimable statuses is a Gas City core fix and is **not** in this PR. This is the pack-side boundary that stops a polecat from creating the state in the first place.

## Fix

A fail-closed **Blocked Work Contract**, mirroring the refinery's shape and reusing its existing `blocked_reason` key rather than inventing a parallel convention.

| Metadata key | Value |
|---|---|
| `blocked_reason` | Free text: what blocked you AND what would unblock it |
| `blocked_disposition` | Kebab-case token naming the required unblock, e.g. `upstream-dependency-required` |
| `blocked_by` | The session identity that blocked it |
| `gc.routed_to` | `human` — a blocked bead leaves the polecat pool |

The polecat prompt gains an executable `GC_BLOCKED` block that is the only sanctioned way to stop on a bead:

1. **Refuse before mutating.** An empty or placeholder reason/disposition (`blocked`, `stuck`, `unknown`, `n/a`, `tbd`), or a reason under 24 characters, prints `BLOCK_REFUSED` and exits 2 having issued **zero** `gc` commands. A refusal cannot leave a bead half-blocked.
2. **Reason first, then read it back.** Metadata and `--notes` are written, then re-read and compared.
3. **If the reason did not persist, do not block.** Print `BLOCK_ABORTED`, leave the bead `in_progress` and claimed, escalate once to the witness, exit non-zero, and do **not** drain-ack. A stale `in_progress` claim is visible to the witness stale/stuck-polecat path; a silently `blocked` bead is not. That is the direction the failure must fall.
4. **Only then** flip status, release the claim, and route to `human` in a single update; re-verify; mail the witness once; `drain-ack`; exit non-zero — the same terminal shape as the refinery's `block_existing_pr` and `halt_false_completion`.

`mol-polecat-work` replaces `mark yourself stuck` with a pointer to the contract and states the five requirements normatively. The executable block is not duplicated.

**Releasing the claim is deliberate.** A blocked bead held by a session that is about to exit can never be progressed by that session, and holding the assignee hides it from the pool and from witness orphan classification while `gc.routed_to` still advertises pool demand — precisely the combination that respawned `cheedo`. Releasing to `gc.routed_to=human` is the same disposal the refinery already performs, and it is what the operator did by hand to stop the live loop.

## Regression coverage

New standalone suite `gastown/tests/test_polecat_blocked_reason.sh` extracts the shipped `GC_BLOCKED` block from the prompt asset and executes it against a fake `gc`, modelling the production incident with the real bead IDs:

- `ki-0aq` with no reason at all — the live defect — must exit 2, print `BLOCK_REFUSED`, and issue **no `gc` command whatsoever**;
- `ki-gu2` with each of the five placeholder reasons must be refused the same way, asserting the *placeholder* refusal message so the list cannot be deleted and still pass on the length check alone;
- a reason write that silently does not persist must never flip status, never release the claim, never drain-ack, and must send exactly one witness escalation;
- the recorded path must write `blocked_reason` **strictly before** the status flip, release the claim in that same update, set `gc.routed_to=human`, and drain-ack;
- a fenced-snippet scan rejects any copyable bare `--status=blocked|escalated` outside `GC_BLOCKED`.

Mutation-tested: neutering the reason guard fails the suite with `a blocked transition with no reason must be refused`; neutering the read-back gate fails with `status must never flip to blocked when the reason did not persist`.

Deliberately a **new** suite file rather than an addition to `test_gastown_pack_assets.sh`, which is contested by #222, #171 and #227. Discovery is convention-based via the CI glob `for test_script in gastown/tests/test_*.sh`, so no registration edit exists to make.

## Validation

- Gastown shell suites: 4/4 pass
- `bash -n` over changed scripts and the extracted `GC_BLOCKED` block
- TOML parse: 15 gastown TOML files incl. `mol-polecat-work.toml`
- `python3 -m pytest -q`: 773 passed, 7 skipped, 9412 subtests
- `tests/test_no_bare_bd_commands.py`: 2 passed (the fake `gc` dispatches on argv positions, not literal `bd <subcommand>` strings)
- `gc lint gastown`: ok
- `git diff --check`: clean

## Composition with open PRs

Merges cleanly with **#221, #222, #226 and #227** (verified by `git merge --no-commit` against freshly fetched heads). The prompt edits are disjoint from #222, which rewrites the `GC_CLAIM` startup block ~50 lines above; the two share the same fail-closed posture (escalate, exit non-zero, never drain-ack a hidden failure). I deliberately stayed out of #227's regions — the new keys are documented in a self-contained table inside the Escalation section rather than added to the Work Bead Metadata Contract table or the Command Quick-Reference, both of which #227 rewrites. A trivial follow-up can hoist them once #227 lands.

### DIFF
--- a/gastown/agents/polecat/prompt.template.md
+++ b/gastown/agents/polecat/prompt.template.md
@@ -341,7 +341,145 @@ gc mail send "$WITNESS_TARGET" -s "ESCALATION: Brief description [HIGH]" -m "Det
 gc mail send mayor/ -s "BLOCKED: <topic>" -m "Context"
 ```
 
-After escalating: continue if possible, otherwise `gc bd update <bead> --status=escalated && gc runtime drain-ack && exit`.
+After escalating: continue if possible, otherwise stop through the Blocked
+Work Contract below. **Never** hand-write `gc bd update <bead>
+--status=blocked` or `--status=escalated`. A bare status flip is the exact
+defect this contract exists to prevent.
+
+### Blocked Work Contract (fail-closed — the ONLY way to stop on a bead)
+
+A bead you leave `blocked` (or `escalated`) with no recorded reason is
+indistinguishable from a crashed session. The next human or agent has to
+re-derive your entire investigation from nothing, and — because the bead is
+still routed and still claimed — the pool keeps counting it as demand while
+the claim path refuses to hand it out, respawning workers in a loop.
+
+This mirrors the refinery's rejection contract exactly: the durable record
+lives in bead metadata, not in your session transcript.
+
+| Metadata key | Set by | Value |
+|---|---|---|
+| `blocked_reason` | polecat (on block) | Free text: what blocked you AND what would unblock it |
+| `blocked_disposition` | polecat (on block) | One kebab-case token naming the required unblock, e.g. `upstream-dependency-required`, `credentials-required`, `human-decision-required`, `target-platform-validation-required` |
+| `blocked_by` | polecat (on block) | The session identity that blocked it |
+| `gc.routed_to` | polecat (on block) | `human` — a blocked bead must leave the polecat pool |
+
+Run this as ONE Bash command. It records the reason **first**, reads it back,
+and only then flips status and releases the claim. If the reason cannot be
+persisted, it refuses to change status at all: a stale `in_progress` claim is
+recoverable by the witness, a silent `blocked` bead is not.
+
+```bash
+export BLOCKED_BEAD="<the id printed as CLAIMED_BEAD_ID>"
+export BLOCKED_REASON="<what blocked you and what would unblock it — specific, mechanical, no placeholders>"
+export BLOCKED_DISPOSITION="<one kebab-case token: what must happen to unblock>"
+export BLOCKED_STATUS="blocked"   # or "escalated"
+bash <<'GC_BLOCKED'
+# BEGIN POLECAT_BLOCKED_CONTRACT
+set +e
+
+BLOCKED_ACTOR="${BEADS_ACTOR:-${GC_SESSION_NAME:-${GC_SESSION_ID:-${GC_AGENT:-}}}}"
+WITNESS_TARGET="${GC_RIG:+$GC_RIG/}{{ .BindingPrefix }}witness"
+STATUS="${BLOCKED_STATUS:-blocked}"
+
+# --- 0. Refuse before touching the bead. A refusal must mutate nothing. ---
+if [ -z "$BLOCKED_BEAD" ]; then
+  echo "BLOCK_REFUSED no bead id supplied; refusing to touch any bead"
+  exit 2
+fi
+if [ "$STATUS" != "blocked" ] && [ "$STATUS" != "escalated" ]; then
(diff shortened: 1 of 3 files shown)


jm2:
Superseded by #240, which is a much smaller version of the same idea.

Two corrections to this PR that #240 makes. First, its framing was wrong: it said the polecat had no record path, but the line immediately above the bare `--status=escalated` already instructs a `gc mail send` to the witness. The pack does require a record — it puts it in mail. The real gap is addressing, not absence: mail beads carry no `ParentID`, no dependency edge and no bead reference, so there is no bead→mail lookup, and the next agent to claim the bead cannot find it.

Second, this PR introduced `blocked_by` as bead metadata, which collides with bd's existing dependency array of the same name.

#240 is 31 lines of shipped content instead of 168: reuse `blocked_reason` only, zero new keys, write→read-back→flip ordering, fail closed. The routing policy, disposition token, placeholder blocklist and length floor are all dropped.

Closing in favour of that.

PR #232, opened 2026-07-24
fix(gastown): wire real dependency edges from the planner

jm2:
## Problem

`mol-idea-to-plan`'s `create-beads` step ships roadmap blocking as **labels**, not dependency edges. Observed on a live city: beads carry `blocked-on:abi-widening`, `blocked-on:production-ci` — human-readable slugs, not bead IDs. Nothing machine-readable exists, so dependency scheduling is silently off.

To be precise about the defect: `blocked-on` appears nowhere in this repo. The planner **invented** that encoding, because it is the only thing reachable from what the formula actually asks for. The whole dependency half of the step is five lines (`gastown/formulas/mol-idea-to-plan.toml:463-467`):

```
**4. Wire dependencies with `gc bd dep add`:**
Model the actual "X needs Y" relationships. Verify with:
```bash
gc bd blocked
```
```

No command form, no ID source, no failure condition. The design doc names blockers by slug; `gc bd dep add` accepts only bead IDs; step 3 creates the beads and never records the slug→ID mapping. By step 4 the only artifact that could resolve a slug is gone, and the one field on `gc bd create` that takes a free-form human string is `--labels`. The lone "verification", `gc bd blocked`, prints nothing and exits 0 on a graph with zero edges, so the hand-wave self-certifies.

This is a deliberate design that never got a second half — not a missing step, and not a failing one.

### Reproduction

On the affected city, `ki-msb` (M5), `ki-cw9` (M7), `ki-7ea` (M0) and `ki-yvj` all appear in `gc bd ready` with "no active blockers" while carrying `blocked-on:` labels that name unfinished work. Minimal reproduction against a scratch beads DB:

```console
$ bd create "provider b" -t task --labels capability:abi-widening   # -> pr-cry
$ bd create "dependent c" -t task --labels blocked-on:abi-widening  # -> pr-zob
$ bd ready --json | jq -r '.[].id'
pr-cry
pr-zob          # <-- genuinely blocked, reported ready

$ bd dep add pr-zob pr-cry
$ bd ready --json | jq -r '.[].id'
pr-cry          # <-- pr-zob correctly withheld
```

Downstream: nothing auto-schedules. When `ki-vxc` merged, satisfying `blocked-on:bounded-asset-reads`, three dependents stayed dormant until a human noticed and routed them.

### Why the convoy wiring works and this does not

The same run produced correct `tracks` edges — the convoy bead tracks all 10 of its children. Both paths live in the same step, ~5 lines apart, and differ on exactly one property: **who resolves the names.**

Convoy membership (`:458-461`) is one mechanical command, `gc convoy add <convoy-id> <task-id>`, whose arguments are IDs the agent is already holding, and whose `tracks` edge is a **side effect of a core primitive**. No name resolution, no modeling, no judgement.

Blocking (`:463-467`) is a modeling exercise: extract relations from prose, resolve slugs with no mapping available, invert temporal language, and invoke a command that was never spelled out. The pack shipped a primitive invocation for one and a documentation sentence for the other. `tracks` never blocks, so convoy membership could not have compensated.

The pack's own guard did not catch this: `scripts/gascity_pack_inference_gate.py:81` asserts only that the literal substring `gc bd dep add` appears somewhere in the formula. The backtick mention on line 463 satisfies it.

## Fix

Give the planner the ID source it never had, then make the wiring mechanical and fail-closed.

- **Capability ledger** (new step 4): one row per task bead, written at the only moment the mapping exists — as `gc bd create` returns the ID. `.plan-reviews/$REVIEW_ID/plan-ledger.tsv`, three whitespace-separated columns: `<bead-id> <capability-delivered> <capabilities-waited-on|->`. Task beads also carry a `capability:<slug>` label so the mapping outlives the run.
- **`PLAN_DEPENDENCY_WIRING` block** (new step 5): resolves each blocker slug to every bead delivering it, adds the `blocks` edges, and reads each one back through `gc bd dep list --type blocks`. Idempotent — safe to rerun after a partial or restarted planning run. Prints `PLAN_DEP_WIRING_OK edges=<n>`.
- **Fails closed** with `PLAN_DEP_WIRING_FAILURE` and a nonzero return on: missing or empty ledger, wrong column count, identifiers outside `[A-Za-z0-9_-]`, a slug no bead delivers, a self-blocking slug, a failed add, an edge that does not read back, or **any surviving `blocked-on:` label**. A partially wired graph is never reported as done.
- **States the invariant**: a label describes what a bead *is*; only an edge describes what it *waits for*. Convoy `tracks` is membership and reporting only.
- **Verification** upgraded from `gc bd blocked` alone to `gc bd blocked` + `gc bd ready`, with the empty-output trap called out explicitly.

### Design decision: fan-out, not a grouping bead

A slug can name work that is not a single bead — `abi-widening` covered three. The dependent gets **one edge per delivering bead**, so it unblocks exactly when the last piece of that capability closes.

Rejected the grouping-bead alternative: a grouping node must be closed by someone, and if nobody closes it every dependent stays blocked forever — a new silent-stall mode, invisible until a human audits, strictly worse than the bug being fixed. It also collides with convoys, which already occupy the grouping role via `tracks`, a deliberately non-blocking edge type.

**Tradeoff, stated plainly:** fan-out is O(dependents × providers) edges, so `bd dep list` gets noisier, and the edge set is a **snapshot of plan time** — a bead added to a capability after the run does not retroactively block existing dependents. A grouping bead would have absorbed late additions automatically. Accepted because the planning run is the authority, post-plan scope growth is already a human replan event, and the wiring block is idempotent, so re-running it after amending the ledger reconciles the graph without duplicating edges.

## Regression coverage

New standalone suite `gastown/tests/test_idea_to_plan_dependencies.sh` extracts the wiring block from the formula via `tomllib` and runs it against a minimal `gc` bead-store stub. The only way an assertion sees an edge is if `gc bd dep add` actually wrote one, so a step that emits a label cannot pass.

- a plan with one blocking relationship produces a real `blocks` edge — plus direction (`dep add <dependent> <blocker>`, never inverted) and **no `blocked-on:` label in any emitted call**
- a capability delivered by three beads fans out to exactly three edges
- rerun adds nothing and duplicates nothing
- a surviving `blocked-on:` label fails the step and names the bead
- an unresolvable slug, a missing ledger, a malformed row, and a shell-injection payload in a bead ID each fail closed with no edge written
- an add that "succeeds" without persisting, and a hard add failure, both fail closed — the store is the authority, not the exit status
- the formula must *execute* `gc bd dep add`, not merely mention it

Verified the suite fails on `main` (`create-beads has no PLAN_DEPENDENCY_WIRING block; dependencies are unwired`) and passes with the fix. The block was also validated against a real scratch beads DB before the pack test was written: 5 beads, a capability delivered by 3, fan-out produced 4 edges, `bd ready` dropped the dependent, rerun produced 0 new edges, and all five fail-closed paths reproduced.

## Validation

- Gastown shell suites: 4 passed, 0 failed
- `bash -n`: 104/104 tracked shell scripts
- TOML parse: 291/291 tracked `.toml`
- `gc lint gastown`: ok
- `python3 -m pytest -q`: 773 passed, 7 skipped, 9412 subtests — unchanged from the `main` baseline measured in a throwaway worktree
- `tests/test_gascity_pack_inference_gate.py`: 63 passed
- `tests/test_no_bare_bd_commands.py`: 2 passed
- `go test -count=1 .`: ok
- `git diff --check`: clean

No model-backed inference gate was run (needs a live city and paid inference).

## Composition

Touches only `gastown/formulas/mol-idea-to-plan.toml` and one new test file. No overlap with #221, #222, #226 or #227 — none of them touch that formula, and the new suite avoids the three-way-contested `test_gastown_pack_assets.sh`.

---

## Open question for maintainers — please push back if you disagree

This adds ~172 lines where the formula previously had 5, and it makes one design choice that a maintainer may reasonably want the other way. Presenting it as an open question rather than settled:

**Fan-out vs. a grouping bead.** A capability slug can name work spread across several beads (`abi-widening` covered three in the observed run). This PR gives the dependent **one edge per delivering bead**, so it unblocks exactly when the last piece closes. The alternative — synthesising a grouping bead per capability and having dependents block on that — was rejected here because a grouping node must be closed by someone, and if nobody does, every dependent blocks forever: a new silent-stall mode, invisible until a human audits, arguably worse than the bug being fixed. It also collides with convoys, which already occupy the grouping role via the deliberately non-blocking `tracks` edge.

The cost of fan-out, stated plainly: O(dependents x providers) edges, noisier `bd dep list`, and the edge set is a **snapshot of plan time** — a bead added to a capability *after* the run does not retroactively block existing dependents. A grouping bead would have absorbed late additions automatically.

If you would rather have the grouping-bead shape, or a third option, the wiring block is self-contained and idempotent — it is a small change to swap. The part I would argue is not optional is the **fail-closed** behaviour: the current formula's only verification is `gc bd blocked`, which prints nothing and exits 0 on a graph with zero edges, so a step that silently wires nothing self-certifies as success.

### DIFF
--- a/gastown/formulas/mol-idea-to-plan.toml
+++ b/gastown/formulas/mol-idea-to-plan.toml
@@ -13,7 +13,8 @@ have:
 - parallel dispatch via `gc sling ... --on mol-review-leg`
 - completion mail via `gc mail send`
 - one human clarification gate in the live conversation
-- final bead graph creation with `gc bd create` + `gc bd dep add`
+- final bead graph creation with `gc bd create` + `gc bd dep add`, wired from a
+  capability ledger that maps plan slugs to the bead IDs created in this run
 
 Run this from a coordinator workspace that has the target repo checked out
 (typically a crew worker, or a mayor who has already `cd`'d into the rig repo).
@@ -26,7 +27,7 @@ Run this from a coordinator workspace that has the target repo checked out
 4. Dispatch 6 design-exploration legs in parallel
 5. Run 3 PRD-alignment rounds (2 legs each)
 6. Run 3 plan self-review rounds (2 legs each)
-7. Convert the refined plan into beads with dependencies
+7. Convert the refined plan into beads and verified dependency edges
 
 The goal is parity of outcome, not a verbatim port of upstream's convoy DSL.
 """
@@ -427,6 +428,14 @@ needs = ["plan-review-3"]
 description = """
 Turn the final design doc into executable work.
 
+The deliverable of this step is a machine-readable graph, not a prose roadmap.
+Everything downstream of the planner reads **dependency edges**: `gc bd ready`
+computes readiness from them, `gc bd blocked` lists them, the refinery honours
+them, and the `cascade-nudge-on-blocker-close` order exists only to wake the
+dependents of a closed blocker. None of those readers parse labels, bead notes,
+or the design doc. An edge that was never written is a scheduling relationship
+that does not exist.
+
 **1. Read the final plan:**
 ```bash
 cat ".designs/$REVIEW_ID/design-doc.md"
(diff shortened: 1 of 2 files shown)


jm2:
Withdrawing after a source-level review found that both the diagnosis and the
proposed contract are too strong.

The observed missing dependency edges are real, but the current formula already
instructs the planner to run `gc bd dep add`; labels were not the only reachable
representation, and neither the design document nor the newly created bead IDs
are inherently unavailable at that step.

This change also does not mechanically guarantee that the planner records or
runs the new ledger block. It replaces one model instruction with a much larger
one while imposing additional policy—exactly one capability per task, all
blockers internal to the plan, and fan-out to every provider—that upstream has
not selected.

A replacement should start smaller: make the existing ID-to-ID command and
verification concrete, preserve the planner's design freedom, and introduce a
core primitive only if deterministic enforcement is actually required.

PR #239, opened 2026-07-25
fix(gastown): gate refinery closes on hosted-CI REGRESSIONS, not on redness

jm2:
A refinery must not close work that makes hosted CI **worse**.

Not "work whose CI is red" — worse. That distinction is the entire design, and getting it wrong is what makes a CI gate unshippable.

| on the target | on the branch | verdict |
|---|---|---|
| green | red | **REGRESSION — reject, naming the check** |
| absent | red | **REGRESSION — the branch added a failing job** |
| still running | red | not *proven* inherited — reject |
| red | red | pre-existing — **allow**, report it loudly, stamp it on the bead |
| anything | green | pass |
| anything | still running | bounded wait, then UNDETERMINED |
| zero checks | zero checks | no CI on this path — allow |
| has checks | zero checks | coverage the target has was lost here — reject |

That is the whole scope of this PR. It does not review diffs, does not read PR comments, does not name any bot, and does not change what any existing rig pushes.

## Why

One rig's `master`, merged bead by bead by a refinery whose entire quality gate was local:

```
07-22 15:04  beda5d39   success:9            <- last fully green
07-23 09:34  0d5a7558   failure:3 success:6  <- first red
07-24 01:25  dd1f26ba   failure:8 success:1
07-24 22:00  08050416   failure:11
```

Fully green to eleven failing checks in 55 hours. Every failing check was a cross-platform target a Linux worker cannot build: `Windows x86 / {Release,Debug}`, `Portable tests / {Linux,macOS,Windows} {amd64,arm64}`. Polecat green on Linux, refinery green on Linux, merge, hosted CI red, nobody reads it, next merge stacks on top of the last one's red.

## Mechanism

The refinery's quality gate is *entirely local*, and that is visible in the assets rather than inferred:

- `mol-refinery-patrol.toml:272` `run-tests` executes exactly `{{setup_command}}`, `{{typecheck_command}}`, `{{lint_command}}`, `{{build_command}}` (`:285`) and `{{test_command}}` (`:293`) — five shell commands on the machine the refinery runs on.
- `mol-refinery-patrol.toml:298` the Quality-Gate Fallback, for rigs that configure none of those, reads the repo's instructions file and runs *the gates documented there*. Also local.
- `prompt.template.md:47` says the same thing to the agent.
- `mol-refinery-patrol.toml:385` `merge-push` then fast-forwards the target and closes the bead (`:806` direct, `:1066` mr). Its terminal `**GATE**` (`:1088`) requires "verified target push" / "verified PR URL" — the handoff mechanics, not the verdict.

```
$ grep -rn 'check-run\|check_run\|gh pr checks\|commits/.*/status' gastown/agents/refinery/ gastown/formulas/mol-refinery-patrol.toml
$
```

Nothing in the refinery's assets queries a hosted check-run. A Linux worker cannot build `Windows x86 / Release`, so a green local run says nothing about it, and the refinery had no way to find out. That is the bug: not a missing test, a missing *reader*.

## A red base does not block work, and this is not a footnote

**The obvious fix — "hosted CI must be green before close" — is wrong, and we have the receipts.**

On the same operator's rig, `master` went red on 2026-07-23 and *stayed* red. Every branch cut from it afterwards inherited those eleven failures. Under a gate that demanded green, no branch could ever have landed — **including the branches that fixed the failures causing the red**. It sat frozen for roughly 16 hours and only moved when a human authorised a deliberate gate override. An absolute gate would have made that permanent instead of temporary.

So a check that is already failing on the target is **announced, recorded, and allowed through**:

```
CI_GATE_PREEXISTING_RED: 11 checks on 2b2c5626 with no regression against main at 8d446d23;
11 of them are already red on the base and are INHERITED, not caused here:
Portable tests / Linux amd64 [failure]; Portable tests / Linux arm64 [failure]; ...
This work is allowed through because it did not make hosted CI worse, NOT because hosted CI
is well. The base is red and stays red until something fixes it; ci_gate_inherited on the
bead names what was let through.
```

`test_a_red_base_does_not_freeze_the_repository` pins exactly that case — the incident's real eleven failures on both the base and the head — and asserts exit 0, no bead-back-to-pool write, and every inherited name on the record. If that test ever goes red, this gate has become a freeze and must not ship.

The same reasoning drives two other rows people get backwards:

- **Zero checks on both sides passes.** A repository that publishes no CI on this path has nothing for a branch to degrade. Rejecting there locks out every repo without a forge CI — the same class of mistake as freezing a red base. It announces `CI_GATE_NO_HOSTED_CI` and stamps the bead, so nobody mistakes it for a proof.
- **Pending ends in UNDETERMINED, not rejection.** The gate waits, bounded by `ci_timeout_seconds`, and then stops without touching bead state. An unfinished check is not evidence the branch degraded anything, and a slow CI queue is not the branch's fault. Blaming it would be the absolute gate leaking back in through the timeout.

## The fix

One gate between a rebased branch and `gc bd close` (`mol-refinery-patrol.toml:649-1032`, block `REFINERY_CI_GATE`). It resolves the SHA the merge rests on **and** the SHA of the target it lands on, reads check-runs **and** legacy commit statuses for both (a repo can publish either or both; reading one API is a way to mistake silence for success), matches them **by check name**, and rules on the difference.

A check is RED on `failure`, `timed_out`, `cancelled`, `action_required`, or **any conclusion not on the accept list** (`success`, `neutral`, `skipped`) — default deny, because an unrecognized conclusion is not a pass. Default-deny decides what is red; the base comparison decides who answers for it. `test_unknown_conclusions_are_red_and_regress` pins both halves: `stale` against a green base rejects, `stale` against a `stale` base is inherited.

Repeat runs of the same check name are collapsed per name, **asymmetrically**, and both directions are fail-closed (`ci_gate_board`, `:793`):

- **head**: RED beats PENDING beats GREEN. A red run on the head is a red check.
- **base**: GREEN beats RED beats PENDING. A name only counts as already-broken if nothing on the base ever passed it, so a flaky base cannot be used to excuse a head failure.

Three dispositions, no fourth, and no path that silently continues (`:1014` `ci_gate_run`):

- **no attributable degradation** → proceed. `ci_gate_sha`, `ci_gate_base_sha`, `ci_gate_inherited` and `ci_gate_result` are stamped the moment the gate rules (`:750`), not at close time. A `--ff-only` merge makes `merged_sha` the same object as `temp`, so `ci_gate_sha == merged_sha` is readable proof months later that hosted CI ruled on exactly what landed.
- **degraded** → reject through the **existing** rejection flow (`:702`), the same shape `rebase` (`:233`) and `handle-failures` (`:334`) already use: `--status=open --assignee="" --set-metadata rejection_reason=... --set-metadata gc.routed_to=...polecat`, then pour the next wisp and burn this one so the queue keeps turning. The branch is **left intact** — a CI rejection is fix-forward, the polecat pushes another commit to the same branch, and in `mr` mode deleting it would orphan the open PR.
- **unreadable / unresolvable / unfinished** → *undetermined*: STOP without mutating bead state. Cannot-tell is never no-regression, and it is never blamed on the branch either.

`ci_gate=false` is the explicit operator waiver. It is announced on stdout and stamps `ci_gate_result=disabled` on every bead it lets through. The gate never self-disables on error — no error path assigns `CI_GATE`.

This reuses the formula's existing GitHub plumbing rather than introducing new coupling: `resolve_github_token` (`:567`), `init_github_rest` (`:579`), `curl_gh_api` (`:605`), and the `gh`-when-present / token-curl-otherwise pattern that `lookup_pr_info` (`:616`) and the whole PR handoff already run on. `ci_gate_api` (`:736`) is one more reader over the same two paths.

## Which base: the target tip, not the merge-base

**Decision: `refs/remotes/origin/$TARGET`, resolved locally at gate time (`ci_gate_resolve_base`, `:762`).**

The merge-base is the more honest attribution boundary in theory — it is exactly where the branch diverged, so everything after it is the branch's own doing. It is also frequently **un-gradeable**, and that is decisive. Hosted CI attaches check-runs to the commits it was *triggered for*. An arbitrary interior commit of the target's history very often was never a branch tip any workflow ran against, so it carries an empty board — and a comparison against an empty board classifies every red check on the head as "absent on the base", i.e. as a regression. That is the absolute gate wearing a disguise, and it reintroduces the 16-hour freeze through the back door.

The target tip does not have that problem. It is the commit `on: push` workflows fire for, so a repository with CI has a real board there, and it is also the commit this work will actually land on.

**They mostly coincide anyway.** This step rebases `temp` onto `origin/$TARGET` immediately before the gate runs, and after a successful rebase `git merge-base origin/$TARGET temp` **is** `origin/$TARGET`. The two differ only when the target moved between the rebase and the gate — and in that window the `--ff-only` merge below fails anyway, the branch comes back through here after another rebase, and the gate re-runs. A race of seconds, not a design gap.

**The residual, stated plainly:** if the target moved and broke a check in that window, the breakage is charged to the base, not the branch — the direction that errs toward letting work land. And if `origin/$TARGET` cannot be resolved at all, the gate is **undetermined, not a pass** (`test_base_lookup_failure_is_undetermined_not_a_pass`); a gate that quietly stops gating when a fetch is stale is worse than no gate, because it is trusted.

## Zero check-runs: two cases, not one

The previous revision of this PR rejected on zero checks, full stop. Under regression-scoping that is wrong in one direction and right in the other, so it splits:

| | verdict | why |
|---|---|---|
| zero on the head, zero on the base | **allow** | This repository publishes no CI on this path. There is nothing to degrade, and rejecting would lock out every rig without a forge CI for the same bad reason an absolute gate locks out every rig with a red base. Announced as `CI_GATE_NO_HOSTED_CI` and stamped on the bead, so it is never mistaken for proof. |
| zero on the head, the base has checks | **reject** | CI ran for the target and did not run here. The branch lost coverage the target already has: a workflow removed or renamed, or a ref nothing triggers for. "Nothing ran" cannot be compared to anything, and a branch that silences the board has degraded it. |

`ci_zero_check_grace_seconds` still applies before either verdict — a push takes seconds to register workflow runs, and `test_zero_checks_on_both_sides_allows` includes the case where the board appears on the second poll and must be read rather than treated as absent.

## Forensics: you can see *why*, not just *what*

A reviewer must never have to guess which check triggered a rejection, or whether a red target was caused by a given bead or merely survived it. Four metadata fields, on both the pass and the reject path:

| field | records |
|---|---|
| `ci_gate_sha` | the SHA the verdict was measured *on* |
| `ci_gate_base_sha` | the SHA it was measured *against* — what "worse" was worse than |
| `ci_gate_inherited` | every failure let through as pre-existing, by name |
| `ci_gate_result` | the full verdict sentence, counts and names included |

Rejection reasons name the specific checks **and what those checks were doing on the base**:

```
2 of 4 checks are red on <sha> and are NOT red on main at <base>:
Windows x86 SP / Debug [failure; on the base: no such check];
Windows x86 SP / Release [failure; on the base: no such check].
This branch made hosted CI worse. Fix these checks on the same branch and reassign.
```

That is the real observed case behind the `absent → red` row: a bead added two new CI jobs that failed immediately, taking the target from 9 failures to 11, while its own acceptance criteria claimed a hosted validation it never performed. Nothing that was passing broke — and the branch still made things worse. `test_absent_on_base_and_red_on_head_is_a_regression` runs it against an **already-red** base, so it also proves that rule is not the absolute gate reappearing on a green base.

## Direct mode: gated, but it publishes nothing new

**Decision: the gate runs in both modes. `direct` mode does not push the source branch today, and this change does not make it start.**

Exempting `direct` would exempt the failure — it is the default (`metadata.merge_strategy`), it lands straight onto the target, and it is the mode the incident happened in. But hosted CI rules on SHAs that exist on the forge, and the obvious way to get the rebased head onto the forge is to push it. That push is a behaviour change every direct-mode rig would inherit on upgrade: remote branches where it never had them. Not worth taking uninvited.

So the gate reads the SHA it can honestly read, and says which one it read (`:1184`):

| | what the gate rules on |
|---|---|
| `mr` / `pr` | the PR head SHA. Step 1 already pushed `temp` to `origin/$BRANCH`, so this is the exact object, with **no new push**. The gate asks GitHub for the head rather than trusting the local ref: if they disagree, something pushed after us and the checks belong to a different commit. |
| `direct`, rebase was a no-op | `origin/$BRANCH`, which **is** the commit being landed. Exact, and it announces `CI_GATE_EXACT`. |
| `direct`, rebase rewrote the branch | the published tip. It announces `CI_GATE_REBASED_HEAD` naming both SHAs, and records `landing <sha>, a rebase of it` in `ci_gate_result`, rather than implying the two match. |
| `direct` + `ci_gate_publish_head=true` | the exact landing SHA. Opt-in, default `false`. This is the only thing in the change that pushes a ref, and only a rig that asks for it gets it. |

The **base** side is the same in both modes: `origin/$TARGET`, which in `mr` mode is also the branch the PR targets (step 3 already refuses a PR whose base is anything else).

The middle case is the honest compromise, and it is worth being precise about what it does and does not prove. It proves hosted CI ruled on the branch's own content — which is exactly the incident's failure mode, since rebasing onto a newer target does not fix a Windows build break in the branch's diff. It does not prove the *combination* of branch and newer target. That residual is the same one every "CI on the PR branch" merge queue carries and the reason GitHub ships "require branches to be up to date before merging"; `ci_gate_publish_head=true` closes it for rigs that want it closed.

New vars, all namespaced `ci_*`, zero collisions with existing pack vocabulary: `ci_gate` (`true`), `ci_timeout_seconds` (`900`), `ci_poll_seconds` (`60`), `ci_zero_check_grace_seconds` (`300`), `ci_gate_publish_head` (`false`).

## Upgrade behaviour

- A direct-mode rig **pushes exactly what it pushed before**. Pinned by `test_direct_mode_publishes_nothing_by_default`, which fetches and compares `origin/$BRANCH` on both the pass and the reject path, and by a structural assertion that the single `git push origin "HEAD:$BRANCH"` in direct mode sits inside the `ci_gate_publish_head` opt-in.
- A rig **with** hosted CI starts rejecting branches that break checks the target passes. That is the fix.
- A rig **with a red target branch keeps landing work.** That is the other fix, and it is the one that decides whether the first is usable.
- A rig **without** hosted CI is unaffected: zero on both sides passes. No flag needed, no wedge, no `ci_gate=false` required just to keep moving.
- No step-graph change: `merge-push` still `needs = ["handle-failures"]`, asserted.
- No metadata key is repurposed. `ci_gate_sha` / `ci_gate_base_sha` / `ci_gate_inherited` / `ci_gate_result` are new; `rejection_reason` and `gc.routed_to` are used exactly as the existing rejection paths use them.

## Regression coverage

`gastown/tests/test_refinery_ci_gate.sh`, 19 cases. It extracts the **shipped** gate block and the **shipped** call chains out of the TOML, substitutes the **shipped** var defaults, and runs them against a hermetic `gh`/`gc` and a real throwaway git repo with a local bare `origin` — so `--force-with-lease`, ref comparison, and the gate's own `git rev-parse refs/remotes/origin/$TARGET` behave as they do against GitHub. Fixtures are per-commit, because the gate reads two boards. It asserts on behaviour, not prose. Timing knobs are overridden in the driver so the bounded waits resolve immediately; the shipped defaults are asserted separately.

Every row of the table, plus the machinery around it:

green→red rejects naming the check *and* what it was doing on the base · red→red allows and reports · **eleven-red base with an eleven-red head allows, proving a red repo is not frozen** · absent→red rejects, against an already-red base · pending-on-base does not excuse a red head · green→green passes and the shipped close chain closes the bead · zero/zero allows · base-has-checks/head-has-none rejects · a base with no board at all cannot attribute a red head, and says so · pending waits, then goes undetermined, but passes when it finishes green inside the window · a proven regression does not wait for the rest of the board · unknown conclusions are red, and are inherited when equally unknown on the base · legacy commit statuses are compared on both sides · an unresolvable base, an unreadable base board, and a truncated base page are all undetermined, never a pass · every head-side API failure mode is undetermined, never green and never a branch rejection · direct mode publishes nothing by default on both paths · the opt-in publishes and gates the exact landing SHA · the waiver is loud and recorded and skips the API entirely · mr mode gates the PR head SHA, refuses a head that is not what we pushed, and measures against the same target tip.

Every rejection assertion requires a *specific* `rejection_reason` — the regressed check names with their base state, or the lost-coverage sentence — plus `ci_gate_base_sha` on the record, bead-to-pool, claim released, routed to polecat, next wisp poured, current wisp burned, branch intact, and bead not closed.

Mutation-tested; 10 of 10 caught, with the message each produces:

| mutation | caught by |
|---|---|
| the absolute gate returns: every red check on the head is a regression | `FAIL: a head that is red in exactly the ways its base is already red must be allowed through (exit 1)` |
| a failing job the branch ADDED is excused as pre-existing | `FAIL: the CI gate must reject with exit 1 (got 0)` |
| a base with no board at all is used to attribute the head's failures | `FAIL: a red head measured against a base that publishes no board must be allowed through (exit 1)` |
| a still-running base check is treated as proof the failure is inherited | `FAIL: the CI gate must reject with exit 1 (got 0)` |
| a head that publishes no checks while the base does is waved through | `FAIL: the CI gate must reject with exit 1 (got 0)` |
| zero checks on both sides rejects, as the absolute gate did | `FAIL: a repository that publishes no checks on either side must be allowed through (exit 1)` |
| an unresolvable base falls through instead of stopping | `FAIL: an unresolvable base ref must not pass` |
| a head still running at the deadline is blamed on the branch | `FAIL: a head board that never finished must report CI_GATE_UNDETERMINED` |
| a rejection does not record the base it measured against | `FAIL: a rejection must record the base SHA it measured against; a reviewer must never have to guess what 'worse' was worse than` |
| the gate stops reading a base board at all | `the gate must resolve and read a BASE board; without one every red check on the head looks like a regression and a repository with a red target can never land its own fix` |

The structural suite additionally pins that `ci_gate` defaults on, `ci_gate_publish_head` defaults off, direct mode runs the gate before it touches the target, `mr` mode runs it before the close, and that `prompt.template.md:19` — `FORBIDDEN: Reading polecat code to "understand what they were trying to do."` — stays untouched, so nothing later drifts a CI reader into a code reviewer. Whether the refinery should also review diffs is a separate question with an existing upstream answer (`superpowers-code-review`, `compound-code-review`, `gstack-code-review`), and it is not this PR's question.

## What this deliberately does not catch

A regression gate is **narrower than an absolute gate**, on purpose, and it is worth naming the gap rather than letting a reviewer find it:

- **It does not make a red target green.** Eleven inherited failures land eleven inherited failures, one bead at a time, forever, until something fixes them. The gate reports them on every bead (`ci_gate_inherited`) and stops there. Driving them to zero is a human's job, or a separate bead; this gate only guarantees the number never goes up because of a merge.
- **It does not catch a branch that trades one failure for another** if both check names are already red on the base. Name-level comparison sees `X: red → red`; it does not read the failure text.
- **It cannot attribute anything when the base publishes no board but the head does.** A repository whose CI runs only `on: pull_request` never has checks on its target tip. Charging the head's failures to the branch on that evidence would freeze exactly the repositories this rescoping exists to unfreeze, so those failures are reported as `CI_GATE_UNATTRIBUTABLE`, recorded on the bead, and allowed (`test_a_base_with_no_board_cannot_attribute_a_red_head`). It is a real hole, and it is the deliberate choice over a real freeze.
- **It cannot see past one page.** More than 100 check-runs or statuses on either side is truncation, and truncation is undetermined rather than a guess. A repo that large needs pagination here before it needs anything else in this PR.
- **A `paths:`-filtered workflow can produce a false rejection.** A docs-only branch that legitimately triggers no jobs, in a repo whose target tip has a full board, hits the lost-coverage rule. It is recoverable (bead back to pool, branch intact, reason self-describing) but it is not fix-forwardable by a polecat, and a rig that hits it repeatedly should be running an always-succeeding required job — the same remedy GitHub's own required-checks documentation gives.

## Validation

- `bash gastown/tests/test_refinery_ci_gate.sh` — pass (19 cases)
- all five `gastown/tests/test_*.sh` suites — pass (`test_gastown_pack_assets`, `test_gastown_theme_scripts`, `test_polecat_churn_watcher`, `test_witness_heartbeat_check`, and this one). `test_gastown_pack_assets.sh` is untouched.
- `python3 -m pytest tests -q` — 101 passed, 8 skipped (includes `test_no_bare_bd_commands`, which the hermetic `gc` stub is written to satisfy)
- `bash -n` on every `gastown/**/*.sh`, and on the extracted gate block and both extracted call chains at the top of every test case — a fragment that is not valid shell fails the suite before it asserts anything
- TOML: all 307 TOML files in the repo parse; the gate block is asserted to carry no unsubstituted `{{var}}` after default substitution
- `gc lint gastown` — ok
- each mutation above confirmed FAILING with the suite applied

## Composition with open PRs

Test-merged against `upstream/main` (6373985) + this branch, one at a time; each merge clean, full suite + `gc lint gastown` run on every merged tree that shares a file:

| PR | merge | note |
|---|---|---|
| #221 witness orphan-close evidence | clean | no shared file |
| #222 polecat claim fail-closed | clean | no shared file |
| #226 patrol wisp reconciliation | clean textually, **one semantic follow-up** | below |
| #227 artifact worktree safety | clean | shares both refinery files; full suite + `gc lint gastown` pass on the merged tree |
| #232 planner dependency edges | clean | no shared file |
| #240 polecat blocked-row prompt contract | clean | no shared file |
| #190 forge-aware `mr` handoff | clean | shares `prompt.template.md`; full suite + `gc lint gastown` pass on the merged tree |

**#226** replaces the patrol-root query idiom repo-wide and adds a test forbidding `gc bd list --status=in_progress --type=molecule` in refinery assets. `ci_gate_reject` reuses the idiom that is on `main` today — the same one five other sites in this file use, and the one `block_existing_pr` uses for exactly this mid-script terminal case. After a merge, #226's sweep leaves exactly **one** remaining hit (the line in `ci_gate_reject`), and #226's own `test_patrol_wisp_reconciliation.sh` fails on it — verified on the merged tree, which is the failure mode you want. Whoever lands second replaces that one line with #226's reconciliation block. The alternative — omitting the pour/burn to dodge the idiom — would strand the patrol loop on every CI rejection, so this branch keeps it and takes the one-line rebase.

**#190** makes the refinery's `mr` handoff forge-aware (`gh` or `glab`) and is prose-only in `prompt.template.md`. No textual conflict, and no directional tension in the shipped code: the gate's GitHub calls go through `ci_gate_api`, a single reader that already has the `gh` / token-curl split, so making it forge-aware later is one function rather than a scattered edit. The prompt's Command Quick-Reference does gain three rows; if #190 lands first they should be spelled in its two-column GitHub/GitLab table. That is a table row, not a design conflict — and the RED/acceptable vocabulary above is GitHub check-run vocabulary that a GitLab reader would need mapped, which is #190's shape of problem to solve, not a reason to withhold the reader from the rigs that have it today.

## Not in this PR

Deliberately, so this stays one reviewable idea:

- requiring hosted CI to be *green* — that is the design this PR exists to reject, for the reasons in the second section
- reviewing the diff on its merits — inverts `prompt.template.md:19` and duplicates the pack's existing review workflows
- any gate on automated reviewer comments — that hardcodes *someone's* bot list into an upstream pack
- publishing the source branch in direct mode by default — the opt-in is here; the default is not
- pagination past 100 checks — truncation is honest today; making it complete is a separate change

### DIFF
--- a/gastown/agents/refinery/prompt.template.md
+++ b/gastown/agents/refinery/prompt.template.md
@@ -39,6 +39,12 @@ the bead. No separate MR beads.
 | Push fails | Retry with backoff, or abort and investigate |
 | Pre-existing test failure | File bead for tracking (NEVER fix it yourself) — check for duplicates first |
 | Uncertain merge order | Choose based on priority, dependencies, timing |
+| Hosted check red here, green on {{ .DefaultBranch }} | Reject, naming the check. This branch broke it. |
+| Hosted check red here AND red on {{ .DefaultBranch }} | Allow, and say so. Inherited, not caused here. |
+| Hosted check red here, absent on {{ .DefaultBranch }} | Reject: the branch added a failing job. |
+| Hosted CI still running | Wait, bounded. Then UNDETERMINED — stop, do not blame the branch. |
+| Hosted CI reports no checks at all, on both sides | Allow: there is no CI here to degrade. |
+| Hosted CI reports no checks here but does on {{ .DefaultBranch }} | Reject: coverage the target has was lost. |
 
 {{ template "following-mol" . }}
 
@@ -55,6 +61,70 @@ the quality gates documented there instead. Treat their failures the
 same as failures from configured commands (reject or file pre-existing
 bug, per the formula's `handle-failures` step). The fallback preserves
 the quality-gate intent even when pack-specific guidance is missing.
+
+## Hosted-CI Regression Gate
+
+**That section is about running commands. This one is not.** Everything the
+`run-tests` step executes runs on this machine, so it can only prove what this
+machine builds. A Linux worker cannot build `Windows x86 / Release` or
+`Portable tests / macOS arm64`. Passing every configured command earns you the
+right to reach this gate, never the right to skip it.
+
+This is the failure it exists to prevent. One rig's `{{ .DefaultBranch }}`,
+merged bead by bead by a refinery whose entire quality gate was local:
+
+```
+07-22 15:04  beda5d39   success:9            <- last fully green
+07-23 09:34  0d5a7558   failure:3 success:6  <- first red
+07-24 01:25  dd1f26ba   failure:8 success:1
+07-24 22:00  08050416   failure:11
+```
+
+Every failing check was a cross-platform target. Polecat green on Linux,
+refinery green on Linux, merge, hosted CI red, nobody reads it, next merge
+stacks on top.
+
+**The gate is scoped to REGRESSIONS, and this distinction is the whole
+design.** Before a work bead closes, the formula's `merge-push` step reads the
+hosted verdict for the merge SHA *and* for `origin/{{ .DefaultBranch }}`, and
+compares them check name by check name:
+
+| on `{{ .DefaultBranch }}` | on this branch | verdict |
+|---|---|---|
+| green | red | **REGRESSION — reject, naming the check** |
+| absent | red | **REGRESSION — a new failing job is worse** |
+| still running | red | not *proven* inherited — reject |
+| red | red | pre-existing — **allow**, report it, stamp it |
+| anything | green | pass |
+| anything | still running | bounded wait, then UNDETERMINED |
+
+A check is red on `failure`, `timed_out`, `cancelled`, `action_required`, or
+any conclusion not on the accept list (`success`, `neutral`, `skipped`) —
+an unrecognized conclusion is not a pass.
+
+**A red {{ .DefaultBranch }} does not block work, and must not.** On one rig
+the target branch went red and stayed red; every branch cut from it inherited
+those failures. Under a gate that demanded green, no fix could have landed —
+including the fixes for the failures causing the red. That repository sat
+frozen for roughly 16 hours and only moved when a human authorised a
+deliberate override. So an already-failing check is announced
+(`CI_GATE_PREEXISTING_RED`), recorded on the bead in `ci_gate_inherited`, and
+**allowed through**. Report it; do not treat it as your merge's fault, and do
+not treat it as permission to break anything else.
+
+Zero checks follows the same rule. Zero on both sides means this repository
+publishes no CI on this path and there is nothing to degrade: allow. Zero here
+while `{{ .DefaultBranch }}` publishes checks means coverage the target has was
+lost on this branch: reject. Pending is the case people get wrong in the other
+direction — the gate waits, bounded by `ci_timeout_seconds`, and then reports
+UNDETERMINED rather than rejecting. A check that never finished is not evidence
+this branch degraded anything, and a slow CI queue is not the branch's fault.
+
+A rig with no hosted CI at all sets `ci_gate=false`. That is an explicit
+operator waiver, it is announced, and it stamps `ci_gate_result=disabled` on
+every bead it lets through. The gate never turns itself off: an unreadable API
+is *undetermined*, which STOPs without mutating bead state, and is never
+treated as a pass.
 
 ---
 
@@ -228,13 +298,23 @@ Never infer a branch name. If `metadata.branch` is missing, reject the bead.
 
 ## Rejection Flow
 
-On rebase conflict or test failure:
+On rebase conflict, test failure, or a hosted-CI rejection:
 1. Put work bead back in pool:
    `gc bd update $WORK --status=open --assignee="" --set-metadata rejection_reason="..."`
 2. Branch handling depends on failure type:
    - Conflict: leave branch intact (polecat needs it for rebase)
    - Test failure: delete branch (polecat redoes work)
+   - Hosted-CI regression: **leave branch intact.** A CI rejection is
+     fix-forward — the polecat pushes another commit to the same branch and
+     reassigns. Deleting it would throw away work that is one commit from
+     correct, and in `mr` mode it would orphan an open PR.
 3. Pour next wisp, burn current one
+
+A CI `rejection_reason` must be specific enough to act on: which checks this
+branch broke and what they were doing on the base, or that the SHA carried no
+checks at all while the base did. `"CI failed"` is not a rejection reason, and
+neither is `"CI is red"` when the base is red too — that one is not a rejection
+at all.
 
 A new polecat picks up the bead, sees `metadata.branch` and
 `metadata.rejection_reason`, rebases or redoes work, reassigns to refinery.
@@ -257,6 +337,18 @@ contradictory record on the bead.
 
 - `direct` — merge to target and push normally
 - `mr` / `pr` — push the rebased source branch and create or update a GitHub PR
+
+**The hosted-CI gate runs in both modes; only the SHA it can read differs.**
+`mr` mode already pushes the rebased branch, so the gate reads the PR's head
+SHA and rules on the exact object. `direct` mode does not push the source
+branch and this gate does not make it start: it rules on the published tip of
+`metadata.branch`, which IS the commit being landed whenever the rebase was a
+no-op, and when a rebase rewrote the branch it says which SHA it proved and
+which one lands. A rig that wants the exact-object guarantee in `direct` mode
+sets `ci_gate_publish_head=true` and accepts the branch push that comes with
+it; a rig that does not set it pushes exactly what it pushed before. The BASE
+side of the comparison is the same in both modes: `origin/$TARGET`, the tip the
+work lands on and the commit hosted CI actually publishes a board for.
 
 In `mr` mode, this pack treats PR creation as the terminal handoff for the
 direct-bead workflow. Record `pr_url` on the work bead, close the bead, and
@@ -316,6 +408,9 @@ alert the witness, not `gc mail send`.
 | Read work metadata | `gc bd show $WORK --json \| jq '.[0].metadata'` |
 | Set metadata field | `gc bd update $WORK --set-metadata key=value` |
 | Remove metadata field | `gc bd update $WORK --unset-metadata key` |
+| Read hosted check-runs for a SHA | `gh api "repos/$ORIGIN_REPO/commits/$SHA/check-runs?per_page=100"` |
+| Read legacy commit statuses for a SHA | `gh api "repos/$ORIGIN_REPO/commits/$SHA/status?per_page=100"` |
+| Resolve the base the gate compares against | `git rev-parse refs/remotes/origin/$TARGET` |
 | Fetch remote branches | `git fetch --prune origin` |
 | Rebase on target | `git rebase origin/$TARGET` |
 | Fast-forward merge | `git merge --ff-only temp` |
--- a/gastown/formulas/mol-refinery-patrol.toml
+++ b/gastown/formulas/mol-refinery-patrol.toml
@@ -27,6 +27,13 @@ In `mr` mode, refinery treats PR publication as the terminal handoff for
 the direct-bead workflow: it records the PR URL on the work bead and
 closes the bead once the PR is verified.
 
+Before a work bead closes in either mode, hosted CI must be green on the
+merge SHA. The `run-tests` step runs configured commands on THIS machine, so
+it can only prove what this machine builds; the CI gate reads the verdict of
+the checks it cannot run. Pending is not green and zero checks is not green.
+A CI rejection uses the SAME flow as a conflict or a test failure: bead back
+in the pool with a specific `rejection_reason`, branch left intact.
+
 Read each step's description before acting — Config values override defaults."""
 formula = "mol-refinery-patrol"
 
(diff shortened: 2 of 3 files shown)


jm2:
Withdrawing after an independent correctness and portability audit.

This cannot safely be a default upstream refinery gate:

- it is GitHub-specific and hard-stops non-GitHub or unreadable origins;
- in direct mode it can evaluate the published pre-rebase SHA and then land a
  different rebased SHA that hosted CI never tested;
- its shipped introduction promises merge-SHA/green semantics that the actual
  inherited-red and zero-check allowances do not provide;
- common PR-only and path-filtered workflow shapes create acknowledged false
  passes or false rejections; and
- the default-on policy is 640 lines of prompt/formula behavior before its
  1,153-line test suite.

The missing hosted-CI signal is a real problem, but the policy, forge
abstraction, tested-SHA contract, and lifecycle behavior need upstream design
agreement before another implementation is proposed.

PR #241, opened 2026-07-25
fix: add --include-infra to wisp reconciliation queries in patrol formulas

hunterBhough:
Wisp reconciliation queries were missing --include-infra, causing them to be blind to ephemeral wisps. This led to duplicate wisp leaks on agent restarts (~1 per restart across witness, refinery, and deacon patrol loops).

## Changes

- mol-witness-patrol.toml: add --include-infra to OPEN_WISPS query
- mol-refinery-patrol.toml: add --include-infra to all wisp queries and add full reconciliation logic (reuse open wisps instead of always pouring fresh)
- mol-deacon-patrol.toml: add --include-infra and full reconciliation logic

## Why This Fixes It

The fix ensures patrol agents reconcile to exactly one open wisp per cycle, reusing any queued wisp from a prior cycle instead of leaking duplicates on restart. This prevents the unbounded wisp accumulation observed in production.

## Testing

- Verified on gastown.witness, gastown.refinery, and gastown.deacon patrol loops
- Reconciles 2+ open wisps down to exactly 1 on restart
- No work lost (formula steps are not materialized as child beads)

Fixes gc-c6l

### DIFF
--- a/gastown/formulas/mol-deacon-patrol.toml
+++ b/gastown/formulas/mol-deacon-patrol.toml
@@ -517,28 +517,39 @@ gc mail inbox
 ```
 Handle any urgent messages that arrived during patrol. Archive the rest.
 
-**3. Resolve current wisp and pour next iteration BEFORE burning:**
-```bash
-CURRENT_WISP=${GC_BEAD_ID:-}
-if [ -z "$CURRENT_WISP" ]; then
-  CURRENT_WISP=$(gc bd list --assignee="$GC_AGENT" --status=in_progress --type=molecule --limit=1 --json | jq -r '.[0].id // empty')
-fi
+**3. Reconcile to exactly one open wisp (reuse or pour), then burn current:**
 
-NEXT=$(gc bd mol wisp mol-deacon-patrol --root-only --var binding_prefix='{{binding_prefix}}' --json | jq -r '.new_epic_id // empty')
+Reuse an already-queued wisp instead of pouring a duplicate. A prior cycle
+may have poured a next wisp without burning, or a restart may have raced —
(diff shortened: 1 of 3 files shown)


edwarddavison:
Fresh production evidence for prioritizing this: on 2026-08-01, **three independent witness sessions on three different rigs** of a live gas-city deployment (v1.4.0) hit and reported this exact bug within ~40 minutes of each other — each found a leaked pre-existing patrol wisp its reconcile couldn't see, poured a duplicate, and had to clean up by hand:

- arnotts witness: `gc bd list --type=molecule` → none; `bd query 'ephemeral=true AND status=open'` → 2 open wisps (one leaked from a prior restart)
- gas-city witness: verified the one-flag repro live — `gc bd list --type=molecule` → empty; adding `--include-infra` → shows the wisp
- detmold witness: same reconcile blindness, same manual cleanup

Every witness/refinery/deacon cycle re-leaks on restart until this merges. The town is self-healing per-cycle via ephemeral-aware queries in the meantime, but the formula fix here resolves it fleet-wide. (Tracking bead on our side: gci-9tw.)

edwarddavison:
Evidence update from our deployment: the report count for this class is now **eight, across five rigs, in 48 hours** — including two new shapes overnight: (a) on one rig `gc bd list --type=molecule` returns 0 results *unconditionally* (isolation-tested across assignee/status/limit/--all combinations), and (b) wisps never transition to in_progress there, compounding the reconcile blindness. Witness patrols are self-healing every cycle via `bd query 'ephemeral=true'`, but the interval between recurrences is tightening. This PR (plus the runtime-side #252/#145 family) is our highest-frequency operational papercut — a review would be very welcome.

PR #247, opened 2026-07-27
fix(slack-full): file ingest gates, name-bound react lookup, busy-reaction lifecycle, event_id dedup

aphexcx:
Four adapter defects found by running this pack against a live Slack workspace, plus the darwin build fix that was a prerequisite for testing them on macOS. Each fix ships with regression tests; the branch is rebased onto current `main` with the recent company-rooms/DM work reconciled.

## Root causes and fixes

**1. Human file posts never reached sessions.** The system-noise subtype gate dropped subtype `file_share` (a human posting a file), and the empty-text gate dropped file-only posts (a screenshot with no caption). Both now flow through to the existing `downloadSlackFiles` path; other subtypes and bot-authored file posts stay filtered.

**2. `/react` and reply-current lookups missed name-registered sessions.** Bindings register whatever identifier the operator passed — for named sessions, the *name* — and gc stamps it into the inbound `target_session` verbatim, while the scan matched only `GC_SESSION_ID`. The scan now matches the id, `GC_SESSION_NAME`, and the gc-reported alias/session name for the id (best-effort `/sessions` resolve). The ambient `GC_SESSION_NAME` only joins the match set when the requested session *is* the current one, so an explicit `--session <other>` cannot be misrouted.

**3. No busy affordance on targeted inbounds.** New busy-reaction lifecycle: `BUSY_REACTION` (default `hourglass`; set-but-empty disables) is added to a targeted inbound message on dispatch and removed when the agent's reply publishes back into the same conversation/thread — a channel-native replacement for Slack Assistant-mode `assistant.threads.setStatus`. Bounded-TTL registry with cap eviction; the remove is ordered after the in-flight add (a fast reply cannot leave a stale emoji); failed forwards cancel the mark; re-targets move the affordance to the newest targeted message and restore the displaced mark if the re-target fails. `/react` gains `remove:true`. Only parsed targets (`@handle:` prefix, user-group mention, sticky thread handle) get the reaction — channel-binding-only inbounds carry no parsed target by design, since the adapter has no visibility into gc-side bindings.

**4. Slack Events API redeliveries double-delivered messages.** Slack retries any delivery it considers unacknowledged (network hiccup, slow first ack, its own read timeout) with the same `event_id`, and each retry that slipped through forwarded a byte-identical duplicate into the bound session. A short-TTL seen-set on envelope `event_id` now dedups redeliveries. The claim is taken synchronously before load-shedding (two simultaneous copies cannot both slip through); a delivery whose forward into gc fails releases its claim so Slack's retry ladder can still deliver — a transient downstream hiccup no longer becomes message loss; and a redelivery that races the first delivery's still-running forward parks until that claim concludes instead of being dropped.

**Darwin build (prerequisite):** `openBeneath`'s hand-rolled `openat(2)` walk used raw `syscall.Openat`, which does not exist on darwin — the adapter package failed to compile there, so its tests never ran on macOS. Replaced with stdlib `os.Root` confined resolution, with the root pre-opened `O_NOFOLLOW|O_DIRECTORY` and identity-matched against the `os.Root` handle, and the leaf `Lstat`'d (symlink → hard failure) plus inode-pinned — keeping the symlink-race protections of the old walk.

## Rebase notes

Reconciled with the company-rooms/DM work on `main`:

- `postReactionMethod` stays the single reactions POST path; `callSlackReactions` / `removeReactionFromSlack` are thin wrappers over it using the timeout-bounded `slackAPIClient` (the busy-reaction remove orders itself after the add with a wait derived from that timeout, so the add must stay on a bounded client).
- The dedup log lines use `env.TeamID` (the raw-body `teamID` local no longer exists in the events handler).

## Tests

- `busy_reaction_test.go` — 23 tests: add/remove lifecycle, thread keying, retarget/displacement/restore, add-before-remove ordering, TTL expiry, cap eviction, nil-safety.
- `event_dedup_test.go` — 12 tests: retried `event_id` dropped, distinct ids both forward, queue-full drop releases the claim, failed forward releases the claim, in-flight retry parks and takes over after failure / drops after success.
- `file_share_gate_test.go` — `file_share` and file-only posts forward; empty, bot-authored, and other-subtype posts stay dropped.
- `confined_open_test.go` — regression tests for in-root leaf/intermediate symlinks and a symlinked root.

Verified on this branch: `gofmt` / `go build` / `go vet` / full adapter `go test` (plus `-race` on the new and touched tests) on darwin/arm64 go1.26.5, and the 387 slack-full pytests.

### DIFF
--- a/slack-full/CHANGELOG.md
+++ b/slack-full/CHANGELOG.md
@@ -10,6 +10,22 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Busy-reaction clears are now attributed to the publishing agent
+  (codex r11): each mark records the parsed target handle, and a
+  delivered reply consumes only marks whose handle resolves (via the
+  handle-alias registry) to the publishing session — so when M1
+  targets agent A and M2 re-targets agent B in the same thread, A's
+  late reply can no longer strip B's pending hourglass. Marks with no
+  recorded handle, an unregistered handle, or an unattributable
+  publisher clear unconditionally, preserving the prior behavior.
+- Launcher alias bootstrap no longer hijacks a handle onto another
+  handle's thread session (codex r9): the thread-session registry
+  records which `@@handle` created each binding (persisted; absent on
+  pre-existing records), and the unclaimed-handle alias bootstrap on
+  thread reuse fires only for the same handle (the codex r8 retry
+  case) — a different `@@handle` converging on an existing thread
+  still posts into that session but is not globally alias-bound to
+  it.
 - Busy-reaction lifecycle (hq-xizo, ported from
   `feat/hq-xizo-slack-full-hardening`): a targeted inbound gets a busy
   reaction (`BUSY_REACTION`, default `hourglass`; set-but-empty
--- a/slack-full/CHANGELOG.md
+++ b/slack-full/CHANGELOG.md
@@ -8,6 +8,111 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Busy-reaction lifecycle (hq-xizo, ported from
+  `feat/hq-xizo-slack-full-hardening`): a targeted inbound gets a busy
+  reaction (`BUSY_REACTION`, default `hourglass`; set-but-empty
+  disables) added on dispatch and removed when the agent's reply
+  publishes back into the same conversation/thread, tracked in a
+  bounded in-memory registry keyed by (conversation, thread key) with
+  a 30-minute TTL. Replaces the prior unconditional `eyes` reaction on
+  targeted inbounds; the ⚠️ dispatch-failure reaction stays. `/react`
+  gains an optional `remove:true` issuing `reactions.remove`
+  (`no_reaction` treated as benign delivery, mirroring
+  `already_reacted`), surfaced on the CLI as `gc slack react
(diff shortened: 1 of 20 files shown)


aphexcx:

Codex rounds r9–r17 over the full diff; all P1/P2 fixed except one documented residual:
- Busy-reaction clears are **handle-attributed** (publisher matched via the alias registry, per-mark including stale ancestors) — agent A's late reply can't strip agent B's hourglass
- Thread-session bindings record their creating `@@handle`; alias bootstrap on thread reuse fires only for the same handle (hijack closed, r8 retry preserved), reserved **before** the first-message POST with an atomic `DeleteIf` rollback
- A retaken redelivery whose channel leg already reached gc retries **only the failed alias leg** (channel-leg markers on the event-dedup cache) — no duplicate turns
- Size-aware upload deadline (120s + 1s/MiB, 30 min cap) replaces the fixed 120s window that rejected large uploads
- **Documented residual**: a narrow concurrent same-thread `@@handle` race can roll back a successful launch's alias (degrades to re-binding on the next `@@handle`; never loses a message)

Verification: gofmt/vet clean, `go test ./...` + `-race` green, 387 pytests green.

**Blocked on:** CI workflow approval and branch-protection merge — needs a gastownhall maintainer. **Merge-order note:** #248 is stacked on this branch and contains it; merging #248 merges this too (or merge this first and #248 stays a clean delta). #213 touches the same files — whichever merges second needs a rebase pass.

PR #248, opened 2026-07-27
fix(slack-full): surface inbound file attachments in channel-path text so image-only messages reach sessions

aphexcx:
> **Stacked on #247** — this branch includes that PR's commit; only the last two commits are new here. The fix builds on the file-ingest gates from that PR (an image-only message must first pass the `file_share`/empty-text gates before there is anything to surface).

## Root cause

gc's extmsg pipeline renders only the message *text* to the channel-bound session: the `Attachments` field never surfaced, and an image-only message (empty text) produced **no session reminder at all** — the file was downloaded and spooled, then silently invisible to the session.

## Fix

Fold a files block into the text forwarded via `postInbound`, built from the event's `files[]` — not just the downloaded subset, so failed downloads still surface id/name with an API-fetch hint — with the spooled local path when the download succeeded:

```
[1 Slack file attached]
  1. screenshot.png (image/png, id F0XXXXXXX) — saved to <spool>/inbound/<channel>/<ts>-screenshot.png; Read that path to view it
```

Slack-controlled fields (name, mime, id, path) are neutralized against markup-boundary forgery before embedding. The alias-dispatch path keeps the un-augmented text — it renders its own attachments block, so nothing is double-rendered.

Also gitignores the `gc-slack-adapter.real` build artifact that wrapper-script hosts produce (it made the pack cache worktree read dirty and fail gc import checks).

## Tests

Six new adapter tests: spool-and-forward of the local path, alias text staying un-augmented, and `formatInboundFilesBlock` edge cases (empty files list, mixed spool results, name fallbacks / unknown mime, forged-boundary neutralization). Full adapter suite and the 387 slack-full pytests are green on this branch.

## Live verification

Deployed to a production workspace: a human posted an image-only message (no text) in a bound channel; the bound session's reminder carried the attachment block with the spooled path, and the session read the image from disk. Before the fix, the same message produced no session delivery at all.

### DIFF
--- a/slack-full/.gitignore
+++ b/slack-full/.gitignore
@@ -9,3 +9,8 @@ adapter/gc-slack-adapter
 # Built operator CLI binary — produced by `go build` in cli/ and exec'd
 # by the commands/<cmd>.sh wrappers via $GC_PACK_DIR/cli/gc-slack-cli.
 cli/gc-slack-cli
+
+# Wrapper-script deployments rename the built binary to gc-slack-adapter.real
+# (the checked-in path above becomes an env-loading wrapper). Ignore it too so
+# pack-cache worktrees stay clean after local builds.
+adapter/gc-slack-adapter.real
--- a/slack-full/CHANGELOG.md
+++ b/slack-full/CHANGELOG.md
@@ -10,6 +10,22 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Busy-reaction clears are now attributed to the publishing agent
+  (codex r11): each mark records the parsed target handle, and a
+  delivered reply consumes only marks whose handle resolves (via the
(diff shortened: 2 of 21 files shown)


aphexcx:

This branch is **stacked on #247** (contains all its commits + codex fixes) — merging it delivers both. Its own delta got codex rounds r14–r15, all findings fixed: spooled filenames ban `<`/`>` so the path shown to sessions is byte-identical to disk (the reminder's markup neutralization inserts a zero-width space after `<`), and the channel files block now carries the anti-auto-attach wording ("Read only if relevant; NEVER paste these paths into your reply") instead of unconditionally telling the agent to read the path.

Verification: gofmt/vet clean, `go test ./...` green, 387 pytests green.

**Blocked on:** CI workflow approval and branch-protection merge — needs a gastownhall maintainer.

PR #244, opened 2026-07-26
fix(gastown): require durable artifact handoffs

jm2:
Supersedes only the task-artifact/worktree-lifecycle portion of withdrawn
#227. It deliberately excludes that PR's refinery identity-resolution and
durable-work-selection changes.

Validated candidate: `67aed7158c4f0e5d5fd3f5e2e9c5e84400a90522`.

## Problem

Gastown's polecat recipe creates a task worktree relative to the worker's
current directory and records it as `metadata.work_dir`. For pooled workers,
that current directory is the persistent provider home, so task artifacts can
be nested inside controller-owned, independently prunable state.

That conflates two lifecycles:

- the provider home is reusable controller execution context;
- the task worktree is durable per-bead implementation state that witness and
  refinery recovery must be able to validate independently.

Runtime staging can also place `.gc/` metadata and generated `.agents/skills/`
symlinks in a provider home before its Git worktree is prepared. Without
repository-local excludes, that runtime state appears as task-source dirt and
can provoke a tracked `.gitignore` change.

The existing-worktree `--sync` path also ran `pull --rebase || true` on
whatever branch a reusable provider home happened to retain. A restart could
therefore rewrite a preserved task branch while leaving the intended
path-derived `gc-*` provider branch stale.

## Fix

- Create new task worktrees at the canonical sibling path
  `$GC_CITY_PATH/.gc/worktrees/$GC_RIG/artifacts/worktrees/<bead-id>`, never
  below the provider home.
- Record the validated path as `metadata.artifact_dir`.
- Treat the deprecated task `metadata.work_dir` as migration input only after
  proving it is the exact historical current-city/current-rig provider layout
  `$GC_CITY_PATH/.gc/worktrees/$GC_RIG/polecats/<provider>/worktrees/<bead-id>`
  and belongs to the same rig repository. The metadata key is migrated in
  place; the legacy directory remains provider-nested until terminal cleanup
  so in-flight work is not moved underneath a running shell.
- Reject same-repository worktrees from another city, rig, or namespace.
  Same Git common-dir is not an ownership boundary because multiple Gas City
  instances can share one repository.
- Fail closed when the work-bead metadata read fails or does not return exactly
  one valid record. No worktree or metadata is created from an assumed-empty
  response.
- Refuse an occupied invalid canonical path and preserve it for recovery
  instead of creating a randomly located alternate artifact.
- Make polecat, witness, refinery, and shutdown/recovery paths agree on
  validation, migration, cleanup, and fail-closed behavior.
[...]

Issue #191, opened 2026-07-10
mol-polecat-work.submit-and-exit is not idempotent: a re-claimed step bead would push the agent-home branch to origin


outdoorsea:
## Summary

`mol-polecat-work`'s `submit-and-exit` step is not idempotent. If a worker completes the handoff but dies **before closing its own step bead**, the step bead stays open and routed. The reconciler respawns a worker, which re-claims the step and re-enters `submit-and-exit` from the top — now standing on its agent-home branch with no commits.

Observed live (worker pool on a GitLab-hosted rig; step bead, convoy, and work bead ids from that run).

## Sequence

1. Worker completes steps 1–7: pushes `polecat/<bead-id>`, sets `metadata.branch` / `metadata.target`, clears `gc.routed_to`, reassigns the work bead to the refinery.
2. Worker dies before closing its step bead.
3. Step bead remains `open` + routed → reconciler respawns a worker → it re-claims the step and re-enters `submit-and-exit`.
4. Its worktree is on the agent-home branch (`gc-<agent>-<hash>`), 0 ahead / 1 behind `origin/main`. There are no commits to cherry-pick.

## Why this is dangerous

Following steps 2–6 literally, on that second entry, would:

- **push the agent-home branch to origin** as a stray remote branch,
- `git branch -D` the agent-home branch locally,
- re-run `gc bd update <work-bead> --status=open --assignee=<refinery>`, **reopening a work bead the refinery had already merged and closed** — racing the refinery's own close.

## What actually prevented it

The formula's branch-shape gate **failed closed**, and the worker declined to follow the gate's literal recovery text (cherry-pick + re-submit), which is wrong for this state. It closed only its step bead and drained.

That deviation is the correct behavior and is worth encoding as the intended one, rather than leaving it to the agent's judgment.

## Verification

Checked independently of the worker's self-report (an agent's account of its own actions is not evidence):

- `origin/main` advanced as expected; `polecat/<bead-id>` is **deleted from origin** (refinery cleanup completed).
- `git ls-remote --heads origin` shows **no stray agent-home / `gc-*` branch**. The claimed non-push is confirmed.
- Work bead: `status=closed`, `gc.work_outcome=shipped`, `metadata.branch=polecat/<bead-id>`. **Not reopened.**
- Step bead: `status=closed`. No orphan, no lost work.

## Impact

[...]

PR #251, opened 2026-07-28
fix(gastown): make default polecat workflow crash-safe

jm2:
## Summary

- make `mol-polecat-work` the default GasTown polecat sling formula
- forward-integrate the durable `artifact_dir` contract from #244 and the
  one-shot post-gascity#4835 claim contract from #242
- bind claims, continuations, task worktrees, submit proofs, quarantine, and
  Witness escalation to exact store/workflow/session identities
- replace prompt-owned terminal mutation sequences with deterministic,
  replay-safe pack commands
- keep runtime-owned provider state separate from task artifacts
- tolerate only the supervisor-owned root `/.worktree-stale` marker during
  provider sync while continuing to reject unrelated or nested dirt
- keep the empty `affected_tests_command` self-review path valid Bash and fall
  back to the configured full test command

This branch forward-integrates the exact current #244 and #242 heads without
rewriting their histories.

## Scope boundary

This exact head still uses the existing GasTown Refinery handoff for
`auto_push=true`. The independent semantic Reviewer and new deterministic
Refinery/decision-dispatch overlay are being developed separately and are
**not** claimed by this PR. Local Refineries remain suspended while that gate
is completed; the live canaries below deliberately use `auto_push=false`.

### Claim and direct-store authority

- Call `gc hook --claim --drain-ack --json` once and validate the complete
  receipt/reason/state matrix.
- Accept `claimed`, `existing_assignment`, and post-gascity#4835
  `ready_assignment` only after exact durable readback.
- Treat the bounded runtime name/session/alias set as equivalent for discovery
  while preserving the selected row's exact assignee as mutation authority.
- Bind stateful helpers to the selected rig store rather than ambient
  contributor routing.

### Task worktree and nonterminal execution

- Use canonical `metadata.artifact_dir` outside the reusable provider home.
- Validate source identity, physical containment, Git top-level/common
  repository, workflow root, input convoy, branch, and target before execution.
- Run task commands only through the registered artifact worktree.
- Keep artifact cleanup independent of provider-session cleanup.
- Provider synchronization ignores the exact root `/.worktree-stale` runtime
  marker. A nested `subdir/.worktree-stale` or any other source dirt remains
  visible and fails closed without moving HEAD or deleting files.
- Nonterminal failure persists a durable blocked reason before
  drain/escalation; it does not improvise a raw drain or release.

### Deterministic terminal submit

The formula has one terminal fence:

```bash
gc gastown polecat-submit execute
```

`execute` derives and revalidates the exact live submit step, Graph-v2 root,
input convoy, sole source, rig/binding, canonical branch, task artifact,
session, and store. It invokes the constrained lease and requires create-only
proof refs before source/step mutation:

```text
refs/gascity/polecat-submit-proofs/v1/<generation-key>/context
refs/gascity/polecat-submit-proofs/v1/<generation-key>/head
```

The opaque generation key and immutable context bind the
source/root/step/convoy/session, exact assignee, branch, target, `auto_push`,
head, rig/binding/Witness, repository/worktree identity, and origin fetch/push
fingerprints. Exact replay must reproduce the same context/head; partial,
conflicting, stale, or wrong-store state fails closed. Lost responses recover
from the retained exact proof rather than repeating the stateful lease.

### Quarantine and Witness surface

`gc gastown polecat-step block` persists and reads back the exact source/step
quarantine before drain. Witness patrol validates the full
source/step/root/convoy tuple and sends bounded, deduplicated operator evidence
before recording its receipt. Hostile metadata is represented by bounded
printable labels and hashes.

This PR intentionally surfaces but does not automatically replace quarantined
workflows. A safe reject retry needs a single-owner clear/readback, workflow
delete, source reopen, and reroute transition; partial shell mutation would
recreate the race this PR removes.

## Dependencies and merge order

As of 2026-07-31, all four dependencies remain open:

1. [gascity#4835](https://github.com/gastownhall/gascity/pull/4835) —
   preassigned ready continuation is durably promoted before hook return.
2. #244 — durable artifact ownership.
3. #242 — one-shot startup/claim wrapper.
4. Rebase/refresh this PR after those changes land.

[gascity#4789](https://github.com/gastownhall/gascity/pull/4789) remains an
independent scale prerequisite because it retains fresh runnable sessions
safely through claim handoff.

## Validation on exact head `132dde24598d3ae0208a9a2adf0d508519ec0445`

- all 12 GasTown shell suites
- 265 focused Python tests plus 5,916 generated subtests
- root `go test ./...`
- Bash/POSIX shell syntax and `git diff --check`
- hostile coverage for root-only marker exclusion, nested same-name dirt,
  unrelated dirt, sync refusal/HEAD preservation, and empty/nonempty
  affected-test rendering via `bash -n`
- independent exact-head read-only review: SHIP

The earlier integrated head also passed the broader registry/CI, inference,
composition, Slack, runtime-cloudflare, Go vet, proof/replay, identity,
task-worktree, quarantine, hostile-metadata, and Bash-4 matrices. The two
commits after deployed parent `185f5c6` are limited to the marker and
empty-command fixes plus their tests.

The current GitHub check rollup exposes only a skipped CodeSmith check; it does
not provide hosted Actions/CodeQL attestation for this exact head. This PR
remains draft pending dependency order and hosted execution or an explicit
maintainer disposition.

## Live rollout evidence

Exact `132dde245` is installed in all four local GasTown import bindings with
composed core `35b5d7ecd`. Import/check/config resolution is clean.

The frozen head passed staged Tributary no-push cohorts at ceilings
`2 → 4 → 6 → 8 → 10`. The final cap-ten cohort admitted exactly ten distinct
original sessions and no eleventh or replacement:

- sources `tr-xucw`, `tr-ehmx`, `tr-1bny`, `tr-nrn4`, `tr-ssc9`,
  `tr-7ugw`, `tr-vcs5`, `tr-3qj2`, `tr-121i`, and `tr-p609`
- roots `tr-n7ky`, `tr-z5sk`, `tr-z8lk`, `tr-r8fq`, `tr-pvr8`,
  `tr-dy5s`, `tr-l9up`, `tr-0bk8`, `tr-g5nj`, and `tr-0yv5`
- Furiosa `ac-pljto`, Nux `ac-zl101`, Slit `ac-getee`, Dag `ac-gpusu`,
  Rictus `ac-1a4es`, Toast `ac-yyl8b`, Capable `ac-jdeg5`, Cheedo
  `ac-wteu4`, Angharad `ac-er1wy`, and Keeper `ac-8bn8f`

All ten 14-step roots closed with `gc.outcome=pass`. Every retained task
artifact was clean and exactly one commit ahead of the pinned source, its
create-only context/head proof refs read back exactly, and
`git ls-remote --exit-code` confirmed that no task branch was pushed. All ten
reusable provider worktrees remained clean at the pinned base. Ready
continuations reclaimed on their original sessions without manual nudges;
every original session drained after completion and no replacement churn
appeared.

After independent proof capture, all ten synthetic sources were closed
unassigned as retained-proof no-ops with `canary_completed=true`,
`gc.work_outcome=no-op`, and `auto_push=false`. The city returned to its
five coordination sessions with the supervisor stable at zero restarts.
An idle `gc doctor` passed 98 checks; its seven warnings are advisory
maintenance items rather than controller, store, session, or workflow faults.

This certifies the default polecat workflow and a Tributary ceiling of ten for
no-push execution. It does **not** certify `auto_push=true`, the independent
semantic Reviewer, or the deterministic Refinery; those remain separate
fail-closed gates described above.

## Remaining before ready-for-review

1. Land/rebase the dependency stack in the order above.
2. Obtain hosted exact-head checks or an explicit maintainer disposition.
3. Re-run the focused exact-head validation after the final rebase.
4. Keep semantic Reviewer/deterministic Refinery work in its own reviewed
   integration rather than silently folding an uncanaried control-plane
   redesign into this already large PR.

## Known follow-ups

- bounded cleanup policy for retained submit proof refs after terminal
  source/root state plus a replay window
- core single-owner restart/cancel primitive for safe rejected-workflow
  replacement
- semantic Reviewer to deterministic Refinery decision/receipt integration for
  `auto_push=true`


The1nk:
Field report from a production city, in case it is useful evidence for this PR.

We hit the failure mode this PR describes, and the "bead metadata is the
concurrency primitive" framing matches what we saw exactly.

## What we found

25 stranded `mol-polecat-work` step beads across 5 molecules, oldest ~22 hours.
Three of the molecules had their **entire** 7-step chain left open — `Load
context` through `Finalize workflow` — while the molecule root itself was
`closed`. The other two were stranded from `Submit work to refinery and exit`
onward.

The step beads carried `is_blocked=1` pointing at a blocker that had already
closed. Nothing recomputes that flag, so the chain could never resolve on its
own — the steps were not waiting on anything real, they were waiting on a stale
boolean.

## The cost, which is what made us chase it

The supervisor re-evaluates blocked work, so a permanently-unresolvable chain
becomes a permanent poll. On an otherwise idle box (no workers, no builds, no
convoys) we measured:

```
24 bd process spawns per 16s   parent = gc supervisor run
 4 per 16s                     parent = gc convoy control --serve  (for scale)
```

Sampled, it was dominated by repeated `bd show --json <molecule>` against
**closed** molecules, plus `bd sql select id,is_blocked from issues where
status <> 'closed'`.

Peak effect: load 14.4 on 8 cores, 91% CPU busy. Not Dolt (measured 0.00 cores
instantaneously) and not I/O (zero D-state). Pure process-spawn churn.

## A downstream consequence worth flagging

At 00:59:50Z, while that write churn was running, scheduled Dolt compaction on
the rig database quarantined:

```
reason=post-flatten value hash changed without row-count increase
```

The flatten itself succeeded; verification five seconds later did not, so the
`DOLT_GC` was correctly refused and a marker written. The likely cause is a
writer committing inside the flatten window — which is what a supervisor
spinning on unresolvable blocked work produces. The compaction script has a
path that *proves* a concurrent-writer race and defers instead of quarantining;
that proof did not land.

Net effect: the database is now below the flatten threshold (2 commits) **and**
carries a quarantine marker, so scheduled compaction skips it and
`compact --gc-only` refuses it. It needs manual intervention to ever reclaim
that space again.

We are not claiming this PR fixes the compaction interaction. The point is that
the stranded-step defect has costs beyond the obvious, and they compound.

## Mitigation we used, for anyone who lands here first

Restarting the supervisor clears the in-memory poll set (24 → 6 spawns/16s) but
is only a reset — the strand is in the store. Closing the stranded steps is the
durable local fix: 24 spawns/16s → **0 per 30s** afterwards.

Neither is a substitute for this PR. New molecules strand the same way.

## Also

\#191 looks like the re-entry face of the same root cause, and is consistent
with what we saw: the worker completes its handoff and dies before closing its
own step bead.

Happy to provide more detail, the raw samples, or to test a branch against this
city if that would help.

jm2:
Closing this cumulative integration PR as obsolete under the upstream-first reset. Head 132dde2 is 18 upstream commits behind and carries 35 downstream commits across 40 files (about 26k additions), including the separate #244 and #242 work, so rebasing it would preserve the long-lived patch stack we are dismantling. The only still-required behavior unique to this head—the one-line mol-polecat-work default_sling_formula setting—will be re-authored as a minimal PR from current upstream main; the focused source PRs remain separate for upstream review.

PR #255, opened 2026-07-31
fix(gascity): surface gc bd stderr in check-gate failures so gc.attempt_log carries the real error

24601:
## Summary

The shared check-gate scripts suppress `gc bd` stderr
(`2>/dev/null || fail "gc bd show $ID failed"`), so when the nested `gc`
invocation fails for an infrastructure reason, `gc.attempt_log` records only
an opaque one-liner and the bounded repair loop burns its attempts blind.

Live case that motivated this: in a fresh city, every producer gate failed
three attempts with `build-artifact-check: gc bd show xs-... failed` while
the actual error - visible only after patching stderr capture into the
scripts - was:

```
gc bd: loading config: city import "...": import "gc": remote import
https://github.com/gastownhall/gascity-packs/tree/main/gascity is locked but
not cached at <city>/.gc/cache/repos/<hash>; run "gc import install"
```

With the real error in the attempt log, the misconfiguration is a
one-minute fix; without it, the artifacts looked at fault (they were valid)
and diagnosis took hours. The scripts' own header contract already promises
"All failures print machine-readable lines on stderr; the dispatcher
records them in gc.attempt_log as repair context" - this makes the fail
lines carry the context that matters.

## Fix

Capture `gc bd` stderr per call into a `mktemp` file and append a bounded
(400-byte, newline-flattened) tail to the existing fail message. The
machine-readable prefix and exit-code contract are unchanged; success paths
are unchanged (stderr is not merged into stdout, so captured JSON cannot be
corrupted - the reason this isn't a blanket `2>&1`).

Applied uniformly across all four gate scripts' `gc bd show`/`gc bd list`
suppression sites (grepped for siblings; B23). Two site classes needed
adaptation to the scripts as they actually are:

- Sites that deliberately tolerate bd failure (`|| true`,
  `|| printf '[]'`) keep their fallbacks unchanged; they now emit a bounded
  `<prefix>: note: ...` line on stderr instead of discarding the error. Core
  records check stderr on both pass and fail attempt entries, so on passing
  runs the `note:` prefix marks these as informational.
- `design-review-approved.sh` previously exited *silently* on bd failure
  (bare command substitution under `set -euo pipefail`); those sites now
  fail with a machine-readable ERROR line carrying the captured tail.

## Validation

- [x] `python3 -m unittest discover -s gascity/tests` - 177 tests; all pass
      except two pre-existing `claim` command timeout flakes that fail
      identically on a clean checkout of main on my machine (2s subprocess
      timeouts under load; untouched files)
- [x] new fake-gc cases (now four - all four gate scripts covered): `gc bd` exiting non-zero with a distinctive stderr
      line -> the gate's fail line (build-artifact) / note line
      (implementation-review tolerant path) includes that text
- [x] `gc lint gascity` - ok (gc 1.4.1)
- [x] `bash -n` on all four scripts; executables preserved
      (`test_check_scripts_are_executable_and_portable` green)

## Notes

- No `GC-BF-*` ledger row: no formula graph or behavior change; this
  strengthens the existing repair-context contract (cf. `GC-METH-BR-040`,
  validator errors as repair context).
- Anticipated question - secret leakage into bead metadata: the capture is
  stderr of `gc bd` only (config/store errors), bounded to 400 bytes, no
  env dumps; gate stderr already flowed into attempt logs when scripts
  chose to print it.

## Provenance

Found while running an external mission pack extending `build-base`; the
stderr-capturing variant of these gates has been running in that pack and
produced the diagnosis quoted above.

### DIFF
--- a/gascity/assets/scripts/checks/build-artifact-valid.sh
+++ b/gascity/assets/scripts/checks/build-artifact-valid.sh
@@ -50,8 +50,9 @@ print(value if isinstance(value, str) else "")
 }
 
 GC_ERR="$(mktemp)"
+trap 'rm -f "$GC_ERR"' EXIT
 SHOW_JSON="$(gc bd show "$BEAD_ID" --json 2>"$GC_ERR")" \
-  || fail "gc bd show $BEAD_ID failed: $(head -c 400 "$GC_ERR" | tr '\n' ' ')"
+  || fail "gc bd show $BEAD_ID failed: $(tail -c 400 "$GC_ERR" | tr '\n' ' ')"
 
 SCHEMA="$(metadata_value "$SHOW_JSON" "gc.build.artifact_schema")"
 PATH_KEYS="$(metadata_value "$SHOW_JSON" "gc.build.artifact_path_keys")"
@@ -62,7 +63,7 @@ ROOT_ID="$(metadata_value "$SHOW_JSON" "gc.root_bead_id")"
 ROOT_JSON="$SHOW_JSON"
 if [ -n "$ROOT_ID" ] && [ "$ROOT_ID" != "$BEAD_ID" ]; then
   ROOT_JSON="$(gc bd show "$ROOT_ID" --json 2>"$GC_ERR")" \
-    || fail "gc bd show $ROOT_ID failed: $(head -c 400 "$GC_ERR" | tr '\n' ' ')"
+    || fail "gc bd show $ROOT_ID failed: $(tail -c 400 "$GC_ERR" | tr '\n' ' ')"
 fi
 
 ARTIFACT_PATH=""
--- a/gascity/assets/scripts/checks/build-artifact-valid.sh
+++ b/gascity/assets/scripts/checks/build-artifact-valid.sh
@@ -49,7 +49,9 @@ print(value if isinstance(value, str) else "")
 ' "$2"
 }
 
-SHOW_JSON="$(gc bd show "$BEAD_ID" --json 2>/dev/null)" || fail "gc bd show $BEAD_ID failed"
+GC_ERR="$(mktemp)"
+SHOW_JSON="$(gc bd show "$BEAD_ID" --json 2>"$GC_ERR")" \
+  || fail "gc bd show $BEAD_ID failed: $(head -c 400 "$GC_ERR" | tr '\n' ' ')"
 
 SCHEMA="$(metadata_value "$SHOW_JSON" "gc.build.artifact_schema")"
 PATH_KEYS="$(metadata_value "$SHOW_JSON" "gc.build.artifact_path_keys")"
@@ -59,7 +61,8 @@ PATH_KEYS="$(metadata_value "$SHOW_JSON" "gc.build.artifact_path_keys")"
 ROOT_ID="$(metadata_value "$SHOW_JSON" "gc.root_bead_id")"
 ROOT_JSON="$SHOW_JSON"
 if [ -n "$ROOT_ID" ] && [ "$ROOT_ID" != "$BEAD_ID" ]; then
-  ROOT_JSON="$(gc bd show "$ROOT_ID" --json 2>/dev/null)" || fail "gc bd show $ROOT_ID failed"
+  ROOT_JSON="$(gc bd show "$ROOT_ID" --json 2>"$GC_ERR")" \
+    || fail "gc bd show $ROOT_ID failed: $(head -c 400 "$GC_ERR" | tr '\n' ' ')"
 fi
 
 ARTIFACT_PATH=""
--- a/gascity/assets/scripts/checks/build-artifact-valid.sh
+++ b/gascity/assets/scripts/checks/build-artifact-valid.sh
@@ -50,7 +50,10 @@ print(value if isinstance(value, str) else "")
 }
 
 GC_ERR="$(mktemp)"
-trap 'rm -f "$GC_ERR"' EXIT
+# EXIT alone does not fire on an untrapped signal, and check gates run under a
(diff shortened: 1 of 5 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277293736

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277293875

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277293946

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277310691

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @24601! This PR makes the four `gascity` check gates report *why* a `gc bd` read failed instead of discarding the error to `/dev/null`, so `gc.attempt_log` now carries the actual message (`... failed: IMPORT_LOCKED_NOT_CACHED run gc import install`) rather than a bare "failed". That matters because these gates run unattended inside formula workflows: when one fell over, the operator was left with a dead workflow and no diagnostic, and debugging meant re-running the gate by hand just to see an error the gate had already read and thrown away.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- Every `gc bd show` site in `build-artifact-valid.sh`, `design-review-approved.sh`, `gap-analysis-approved.sh`, and `implementation-review-approved.sh` now captures stderr to a trap-cleaned `mktemp` and interpolates `tail -c 400 | tr '\n' ' '` into the failure message. A repo-wide grep confirms no remaining swallow under `gascity/assets/scripts/checks/`.
- Gate semantics are preserved at every touched site: the two fail-hard sites stay fail-hard, the `|| true` tolerant sites keep their `PARENT_ROOT=$ROOT_ID` fallback, and no exit code changed — so the ten formulas referencing these scripts are unaffected. A base-vs-HEAD negative control with a deliberately failing fake `gc` returned identical exit codes at every site, with only the message improved.
- The notes key on exit status rather than on stdout emptiness, so a successful `gc bd show` that prints deprecation or `BD_OTEL` warnings to stderr — the common case on this rig — produces no spurious note.
- The rebase surfaced one genuine contract question, and the outcome is worth stating plainly: your tolerant handling of the member read collided with #310, which had just made that same failure fail the gate. We kept #310's fail-loud behavior at those three sites. The reason is specific to `gc ready`: it runs one query leg per status, so a tolerant read admits a *partial* union — if the leg carrying the newest `iterate` verdict fails while an older `approve` survives in another leg, the implementation-review gate would fabricate an approval and ship unaddressed findings. Your tolerant semantics survive unchanged at the `gc bd show` sites, where the fallback value is well defined, and your underlying goal — a real error in `gc.attempt_log` — is delivered either way.

Review findings addressed:
- **Test Evidence Quality** (minor, high confidence) — three modified capture sites had no test pin, including the fatal `design-review` handler this PR exists to fix. Every pre-existing fixture failed the *first* `gc bd show`, which short-circuited the gate and left those sites unreachable by the suite. Fixed in `144dfb3` with three new fixtures that let the first read succeed so the second one is the one under test.
- **Test Evidence Quality** (minor, high confidence) — the temp-file cleanup claimed by `78eb817` was not asserted anywhere, so a future edit dropping the trap would have passed silently. Fixed in `144dfb3`: the new design-review test points `TMPDIR` at its own directory and asserts it is empty after the run.
- **Test Evidence Quality** (minor, high confidence) — `import shutil` was added in this range and never used, and CI runs only pytest and `gc lint` over this file, so no linter would have flagged it after merge. Removed in `144dfb3`.
- **Resource Lifecycle & Cleanup** (nit, high confidence) — `trap 'rm -f "$GC_ERR"' EXIT` does not fire on an untrapped signal, and check gates run under a documented 10-minute dispatcher budget, so timeout kills are an expected path that leaked a capture file each time. Widened to `EXIT INT TERM HUP` in all four gates in `144dfb3`.

- `mc-pvg3m` — **Debuggability** (nit, out of range): `gastown/assets/scripts/checks/design-review-approved.sh:86` still carries the old `gc bd list --all --json --limit=0 2>/dev/null` swallow. This is explicitly *not* a finding against your PR — gastown has no `gc bd show` sites at all, so no propagation was owed here, and the line sits outside this diff range in a pack excluded from `MAINTAINED_PACKS`. Filed separately so the same class of silent swallow gets cleaned up there too.

### Maintainer Changes
Maintainer fixes were pushed directly to this PR branch; both of your commits and their authorship are preserved:
- `ebb8ac6` — test(gascity): retarget check-gate error tests at the gc ready member read (the rebase-conflict repair against #310) — 1 file changed, 87 insertions(+), 13 deletions(-)
- `144dfb3` — test(gascity): pin the three unpinned stderr-capture sites and drop a dead import (the four findings above) — 5 files changed, 141 insertions(+), 5 deletions(-)

Combined maintainer delta: 5 files changed, 228 insertions(+), 18 deletions(-).

### Final Review Status
Ready for the merge queue: final head `144dfb3` has passing required GitHub checks (Check, CodeQL, Analyze (python), Analyze (go)), and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/gascity-packs/actions/runs/34280812381

### Review Iterations
1 review pass performed across three independent reviewer lanes, all of which completed with full cross-file coverage and no reduced-coverage caveat. No blocker or major finding was raised; all four minor/nit findings were taken in a single maintainer fixup rather than deferred, and each new assertion was mutation-checked — reverting any of the three capture sites to `2>/dev/null`, or deleting the design-review trap, fails its own test.

---

PR #220, opened 2026-07-21
fix(gascity): migrate 34 formulas off deprecated contract=graph.v2

toddmsimmons:
## What

Replaces the deprecated `contract = "graph.v2"` declaration with `[requires] formula_compiler = ">=2.0.0"` in all 34 gascity formulas, per `gc doctor`'s formula-requirements hint:

> replace deprecated contract = "graph.v2" with [requires] formula_compiler = ">=2.0.0"

## Why

Under gc 1.3.5, `gc formula cook --attach` on any `contract = "graph.v2"` formula fails at compile:

```
gc formula cook: decorate formulas v2 recipe: step github-issue-triage.snapshot: unknown formulas v2 target "gc.run-operator"
```

This breaks late-bound DAG expansion for every adapter workflow in the pack (`github-issue-triage`, `github-issue-fix`, the build-from-* family, etc.). Formulas that already declare the `formula_compiler` requirement (e.g. `mol-scoped-work`) cook and attach correctly on the same install — verified empirically on a live city: the graph.v2 formulas reproducibly fail while `mol-scoped-work --attach` created its full 29-bead sub-DAG.

## Validation

- All 34 files re-parsed with `tomllib` after the edit; asserted `[requires]` contains exactly `formula_compiler` (no adjacent top-level keys swallowed into the table) and `contract` is gone.
- `gc lint gascity` (gc 1.3.5): ok.
- The edit is mechanical and placement-consistent: the `contract` line is removed and the `[requires]` block is inserted at the end of the top-level key block, before the first table header.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #257, opened 2026-08-01
fix(formulas): replace deprecated contract="graph.v2" with [requires] formula_compiler

elviskahoro:
## Summary
- Replace the deprecated top-level `contract = "graph.v2"` field with `[requires]\nformula_compiler = ">=2.0.0"` across all 79 formula files (bmad, compound-engineering, gascity, gstack, superpowers, pr-pipeline).
- Update `resolve_formula` / `resolve_formula_from_dirs` merge helpers in `gascity/tests/test_formula_assets.py` to propagate `requires` instead of `contract`.
- Update the corresponding hardcoded test assertions in `gascity/tests/test_formula_assets.py` and `pr-pipeline/tests/test_pr_pipeline_formulas.py`.
- Update the production CI gate script `scripts/gascity_pack_inference_gate.py` (used by `.github/workflows/gascity-pack-inference.yml`) to validate `requires.formula_compiler` instead of the old `contract` field.

## Why
`gc doctor` in downstream rigs flags formulas still using the deprecated `contract = "graph.v2"` field as a `formula-requirements` warning. This migrates the whole pack to the new `[requires]` table so those warnings clear.

## Test plan
- [x] `gascity/tests/test_formula_assets.py` — full suite passes after updating assertions and merge helpers
- [x] `pr-pipeline/tests/test_pr_pipeline_formulas.py` — updated assertion passes
- [x] Full local test suite: 751 passed, 7 skipped (pre-existing, unrelated flakes/collection issues verified present on `main` too)
- [x] `gc lint .` clean

### DIFF
--- a/bmad/formulas/bmad-build.formula.toml
+++ b/bmad/formulas/bmad-build.formula.toml
@@ -7,9 +7,11 @@ epics/stories so it matches the upstream workflow.
 """
 formula = "bmad-build"
 version = 1
-contract = "graph.v2"
 extends = ["build-base"]
 target_required = true
+
+[requires]
+formula_compiler = ">=2.0.0"
 
 [catalog]
 name = "bmad-build"
--- a/bmad/formulas/bmad-code-review-flow.formula.toml
+++ b/bmad/formulas/bmad-code-review-flow.formula.toml
@@ -7,7 +7,9 @@ review lanes, including gap analysis, with synthesis and fix application.
 formula = "bmad-code-review-flow"
 type = "expansion"
 version = 1
-contract = "graph.v2"
+
+[requires]
+formula_compiler = ">=2.0.0"
 
 [vars]
 [vars.implementation_target]
--- a/bmad/formulas/bmad-decomposition.formula.toml
+++ b/bmad/formulas/bmad-decomposition.formula.toml
@@ -3,10 +3,12 @@ BMAD implementation of the decomposition-base methodology contract.
 """
 formula = "bmad-decomposition"
 version = 1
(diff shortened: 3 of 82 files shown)


iamluxe77:
Current-main triage confirms this remains the best focused merge vehicle for #167, but the sweep needs to be refreshed before it can land.

On `gascity-packs/main` `db4bd054af468dce715293544d3a55d94f029a05` there are now **81** `contract = "graph.v2"` occurrences and only 7 explicit `formula_compiler` declarations. The count has grown since this PR's original 79-file sweep.

Comparison of the existing candidates:

- #202 and #220 are partial: 45 legacy occurrences remain on each head, and their CI check is red.
- this PR's focused head removes every legacy occurrence present on its base and replaces them with explicit compiler requirements;
- #329 also reaches zero on its older tree, but mixes the migration with a much broader 153-file defect bundle and is now conflicting.

Suggested landing contract for this PR:

1. Rebase/reforge the single focused migration on current `main` and re-scan the entire repository, including formula files added after `8610cb9`.
2. Add a closed-schema root test that recursively rejects every `contract = "graph.v2"` declaration and validates that graph-v2 formulas opt in through `[requires] formula_compiler = ">=2.0.0"`.
3. Run the maintained-pack live-GC gate in addition to parser/unit tests, so the replacement is checked by the real compiler path.
4. Only after this head is green and accepted, close #202/#220 as superseded; no additional migration PR is needed.

No fork, replacement branch, or competing PR was created during this audit.

_codex-unknown-model-unknown-reasoning on behalf of Your Name_

PR #261, opened 2026-08-02
fix(gastown): give boot a patrol loop so the always-mode watchdog stops hot-respawning

edwarddavison:
## The bug

`boot` is the only `mode = "always"` agent shaped as a **single pass**: observe the deacon, then `gc runtime drain-ack` + `exit` on every wake.

An always-mode named session is *unconditionally desired* (`compute_awake_set.go` — `case "always": desired[sn] = "named-always"`), so the controller re-materializes it the instant its provider session goes away. `drain-ack` + `exit` therefore doesn't mean "go quiet", it means "respawn me now", and there is no re-wake cadence on that path. The loop:

> wake → triage (~7s, fresh context) → drain-ack → exit → controller sees an unconditionally-desired session that is absent → re-materialize → …

Observed in production before an operator drained the agent:

- **~500 fresh sessions/hr**, flat across every hour — 344 transcripts in ~18h, each individually a *healthy* single-pass triage. The bug is the frequency, not any one session.
- **~2.4M output tokens in 5h** (~20% of the city's daily burn); 0 tokens on the preceding baseline days.
- A **per-second `bead.updated`** flood on boot's own session bead, from the reconciler rewriting it every tick.

## Why the other always-mode agents are fine

`mayor`, `deacon`, and `witness` are also `mode = "always"`, but each runs a patrol wisp, ends its turn **without draining**, and lets `idle_timeout` pace the session recycle. They never exit-and-respawn. `refinery` dodges it a third way (`mode = "on_demand"`).

Boot was the only always-mode agent that was single-pass, no-formula, exits-every-wake — and the only one with **no `idle_timeout` at all**.

## The fix

Make boot match the other watchdogs.

- **New `gastown/formulas/mol-boot-patrol.toml`** — the existing deacon triage, plus the standard `next-iteration` step: pour the next wisp, reconcile to exactly one open patrol wisp, burn the current one, end the turn with the `IDLE: no work, exiting turn.` signal and no drain-ack. Warrant filing picks up the dedup guard the other patrols already use, so a stuck deacon can't spawn racing shutdown dances.
- **`idle_timeout = "15m"` on boot** — always-mode sessions are exempt from the idle-*sleep* path, so `idle_timeout` is the only thing pacing the recycle. Without it the watchdog would go silent after a single pass. 15m keeps deacon-stall detection tight (the deacon's own patrol backoff caps at 300s) at ~96 wakes/day, three orders of magnitude below the incident. Deliberately not paired with `sleep_after_idle` — `idle_timeout` takes precedence and `gc doctor` warns when both are set.
- **Prompt rewritten around the loop**, with an explicit "never drain-ack, never exit" section explaining *why*, so the single-pass shape doesn't get reintroduced by a future edit that reads `drain-ack` as ordinary good hygiene.

### What this deliberately does not do

`mode = "always"` is left alone — it is correct for a watchdog. Flipping boot to `on_demand` would silence it entirely: boot has no work beads, so nothing would ever route to it and the deacon would go unwatched.

## Verification

- `gc lint gastown` → exit 0 (the one new warning, `bd create ... --label`, matches the existing convention in `deacon`/`witness`/`mayor` prompts).
- All four `gastown/tests/test_*.sh` pass.
- `tests/test_no_bare_bd_commands.py` passes over the full tracked tree.
- Formula TOML parses; the embedded shell renders correctly (`printf '%s\n'` stays a literal `\n` rather than being split across lines by TOML escape processing).

## Follow-up (separate, upstream in `gastownhall/gascity`)

Defense-in-depth: the controller should impose a **minimum re-wake cadence / idle backoff** for an `always` + `fresh` session that drain-acks with no action taken, so that no misconfigured single-pass agent can hot-loop this way regardless of pack contents. Filed separately — this PR is the pack-side fix and stands on its own.

https://claude.ai/code/session_01XoXpQZWh6nNxUnuwRhQFtx

### DIFF
--- a/gastown/agents/boot/agent.toml
+++ b/gastown/agents/boot/agent.toml
@@ -1,4 +1,14 @@
 scope = "city"
 wake_mode = "fresh"
 work_dir = ".gc/agents/boot"
+nudge = "Run 'gc prime' to check deacon health and begin the watchdog cycle."
+# idle_timeout is boot's re-wake cadence, and it is load-bearing. Boot backs a
+# mode="always" named session, so it is exempt from the idle-sleep path and is
+# re-materialized the instant its provider session goes away. Without a timeout
+# the session never recycles and the watchdog goes silent after one pass; with
+# the former drain-ack + exit shape it respawned every ~7s (~500 sessions/hr,
+# 2.4M output tokens in 5h — gci-fed). 15m keeps deacon-stall detection tight
+# (the deacon's own patrol backoff caps at 300s) at ~96 wakes/day. Do not pair
+# this with sleep_after_idle: idle_timeout takes precedence and gc doctor warns.
+idle_timeout = "15m"
 max_active_sessions = 1
--- a/gastown/agents/boot/agent.toml
+++ b/gastown/agents/boot/agent.toml
@@ -9,6 +9,10 @@ nudge = "Run 'gc prime' to check deacon health and begin the watchdog cycle."
 # the former drain-ack + exit shape it respawned every ~7s (~500 sessions/hr,
 # 2.4M output tokens in 5h — gci-fed). 15m keeps deacon-stall detection tight
 # (the deacon's own patrol backoff caps at 300s) at ~96 wakes/day. Do not pair
-# this with sleep_after_idle: idle_timeout takes precedence and gc doctor warns.
+# this with sleep_after_idle: they are independent switches (the idle-sleep
+# path reads sleep_after_idle and never consults idle_timeout), and on an
+# always-mode named session the pairing is rejected outright by config
+# validation, not merely flagged by gc doctor. Boot is exempt from idle sleep
+# anyway.
 idle_timeout = "15m"
 max_active_sessions = 1
--- a/gastown/agents/boot/prompt.template.md
+++ b/gastown/agents/boot/prompt.template.md
@@ -5,96 +5,115 @@
 ## Your Role: BOOT (Deacon Watchdog)
 
 You are **Boot**, the deacon watchdog. You run as the controller-managed
-configured `boot` named session. Each wake answers one question: **is the
-deacon stuck?** The controller handles process liveness; you judge work health
-from wisps, pane output, and mail.
+configured `boot` named session. Each patrol cycle answers one question: **is
+the deacon stuck?** The controller handles process liveness; you judge work
+health from wisps, pane output, and mail.
 
 {{ template "architecture" . }}
+
+---
 
 ## Your Lifecycle
 
 `mode = "always"` keeps the `boot` identity present. `wake_mode = "fresh"`
-gives each wake a new provider context. Observe, decide, act, drain-ack, exit.
-Do not rely on prior conversation context or handoff mail. Narrow scope keeps each wake cheap.
+gives each wake a new provider context. You run a patrol loop: observe the
+deacon, act if needed, pour the next wisp, burn the current one, then **end
+your turn and go idle**. `idle_timeout` (15m) recycles the session and the
+next cycle begins. Do not rely on prior conversation context or handoff mail —
+read live state each cycle. Narrow scope keeps each cycle cheap.
+
+### CRITICAL: Never drain-ack, never exit
+
+**Do NOT run `{{ cmd }} runtime drain-ack`. Do NOT run `exit`.**
+
+A `mode = "always"` named session is *unconditionally desired*: the moment
+your provider session goes away, the controller re-materializes it on the very
+next tick. Draining is therefore not "going quiet" — it is "respawn me now".
+Boot was previously shaped as a single-pass agent that ended every wake with
+`drain-ack` + `exit`, and it hot-looped: ~500 fresh sessions/hr, ~2.4M output
+tokens in 5h, plus a per-second `bead.updated` flood on its own session bead,
+until an operator drained it (gci-fed).
+
+Every other always-mode agent (mayor, deacon, witness) ends its turn without
+draining and lets `idle_timeout` pace the recycle. You do the same. Ending the
+turn leaves the session alive and idle — that is the correct, cheap resting
+state, and the only one that keeps this watchdog off the hot path.
 
 ---
 
-## Triage Steps
+{{ template "following-mol" . }}
 
-### Step 1: Check if deacon session exists
+Your formula: `mol-boot-patrol`
 
-```bash
-{{ cmd }} session peek {{ .BindingPrefix }}deacon --lines 1
-```
+---
 
-If the deacon session does not exist, drain-ack and exit. The controller will
-restart dead agents.
+## Startup Protocol
 
-### Step 2: Observe deacon state
+> **The Universal Propulsion Principle: If you find something on your hook, YOU RUN IT.**
 
 ```bash
-# Recent pane output — is the deacon actively working?
-{{ cmd }} session peek {{ .BindingPrefix }}deacon --lines 30
+# Step 1: Check for assigned work (your patrol wisp)
+{{ .AssignedInProgressQuery }}
 
-# Deacon's current patrol wisp — how fresh is it?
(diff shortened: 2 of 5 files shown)


edwarddavison:
Cross-referencing: this is one of two production-affecting PRs in a consolidated review request I left on #264, which enumerates all six of our open PRs with priority. Short version for this one — our instance logged ~500 tool calls/hour continuously and 3,412 assistant messages in a single day from this respawn loop; boot is currently drained rather than fixed, so the fix is unlanded and the mitigation is "switched off". No urgency expected, just flagging the live cost. Happy to rebase or narrow scope if that helps review.

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277293592

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277293730

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277310469

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @edwarddavison! This PR gives boot a real patrol loop
in place of the `gc runtime drain-ack` + `exit` tail that made an always-mode
session ask to be respawned the moment it finished a turn — which matters because
boot is the watchdog every other patrol agent is watched by, and a watchdog that
hot-respawns is both expensive and blind. Moving the observation table and
escalation ladder out of the prompt into `mol-boot-patrol.toml` also gives the
loop a resumable, restart-readable shape, and the new warrant duplicate-guard
closes a race the old prompt had.

branch.

### Original PR Review
Decision: request_changes on the first two passes, resolved to approve.

Specific gaps fixed:
- The three new wisp-reconcile queries omitted `--include-infra`, so they were
  structurally blind to the wisps tier and always returned nothing — the
  reconcile that was supposed to keep exactly one open patrol wisp would instead
  have poured a fresh one every cycle and never burned its predecessor. The flag
  is now present at all three sites, and the fix is verified against the bd
  source path that hides wisps rather than against the incident report.
- The bootstrap pour assigned the new wisp to `$GC_ALIAS`, which gc leaves empty
  when no alias is configured — as it is here — so the first wisp of every cold
  start would have been orphaned and invisible to both the formula's own
  reconcile and gc's crash-recovery probe. It now assigns `$GC_AGENT`, which gc
  guarantees is non-empty.
- The repo already had a guard for exactly the `--include-infra` bug class, but
  it is deliberately witness-scoped, so boot's new queries sailed straight past
  it. A boot-scoped sibling guard now covers them, mutation-verified in both
  directions (red before the flag fix, green after, and red again when a new
  unflagged site is added).
- The second pass caught a defect the first round of fixes had introduced: the
  mainline burn was rewritten to target a bare `$GC_BEAD_ID`, which gc only sets
  for ralph check scripts and hook-claim workers, never for boot's session class.
  The burn silently reduced to a no-op. It now seeds `CURRENT_WISP` from
  `${GC_BEAD_ID:-}` with an in-progress fallback query, and burns only when both
  the current wisp and its successor resolve — the idiom this PR's own prompt
  already used ninety lines away.

Review findings addressed:
- Error handling / resilience (minor): the formula's every-cycle pour path was
  unguarded — bare `jq -r '.new_epic_id'`, no check on the pour or the assign,
  and an unconditional burn — while the prompt shipped the guarded twin. All pour
  sites in both files are now guarded, with distinct messages per failure branch
  and without deacon's `drain-ack`. Fixed in `ca9be18`.
- Resource lifecycle (minor): pour-then-assign is not atomic, so a failed assign
  left a freshly poured wisp unassigned and therefore invisible to every
  assignee-scoped reconcile — one permanently orphaned wisp per occurrence. All
  three pour-then-assign sites now reclaim by id best-effort. Fixed in `ca9be18`.
- Architectural consistency (minor): the prompt's crash-recovery fallback had
  dropped the surplus-wisp burn leg that the formula and the witness prompt both
  carry, so the recovery path preserved exactly the accumulation the mainline
  cleaned up. The fallback now lists at `--limit=0`, keeps the first wisp and
  burns the rest. Fixed in `ca9be18`.
- Behavioral correctness (major severity, minor merge impact): the
  deacon-observation query was blind to the wisps tier by the same mechanism this
  PR documents eleven lines below it, making the decision table's wisp-keyed rows
  untriggerable. This one is carried content rather than logic the PR authored,
  but the PR raises its exposure from "drained agent" to a loop running many
  times a day, so it was fixed here anyway. Fixed in `ca9be18`.
- Debuggability (minor and nits): the `idle_timeout` comment described pairing
  with `sleep_after_idle` as a `gc doctor` warning when config validation
  actually rejects it outright for an always-mode session; the "never exit"
  banner read as an absolute sixty lines above two sanctioned `exit 1`s; and the
  burn step carried an unresolvable `<this-wisp-id>` placeholder. All reworded or
  resolved in `ca9be18`.
- Test evidence quality (minor): the un-blinded deacon-observation query shipped
  without a pin — a whole-line revert to its blind form left the suite green,
  because the line leaves the numerator and denominator together. A text pin now
  covers both assets and was validated red in four separate regression
  directions. Fixed in `ca9be18`.
- Cross-file consistency (nit): the guard tests' own comments described a site
  inventory the fixes had made stale. Corrected in `ca9be18`.

- mc-x80bo: Resource lifecycle / operability, minor — `mol-deacon-patrol.toml`
  still carries the `gc runtime drain-ack` + `exit 1` hot-loop shape at its
  pour, assign and burn failure paths, which is the shape this PR removes from
  boot. Deliberately out of scope here: deacon is a different patrol with a
  different pour/burn contract, and the two may not want the same policy.
- gastownhall/gascity-packs#252: Behavioral correctness, minor — deacon and
  refinery still carry the bare `--type=molecule` wisp-query form without
  `--include-infra` at eleven sites. Already tracked; the boot guard is
  deliberately boot-scoped rather than pack-wide precisely because a pack-wide
  assertion would land red against those sites.
- gastownhall/gascity-packs#268: Architectural consistency, nit — three sibling
  patrol formulas promise that "the session_sleep policy will restart this
  session after the configured idle interval", which is not what
  `sleep_after_idle` does. Review confirmed that boot's new wording is the
  accurate pole here, so the siblings were deliberately left alone rather than
  boot's prose being changed to match them. Already tracked.

One local test, `tests/test_gastown_lint_findings.py`, fails in this development
environment. It was A/B-measured at the base commit and at this PR's head across
three review passes and is identically red at both — a known divergence between
the local `gc` binary and CI, present before this PR existed. It was explicitly
left alone rather than papered over.

### Maintainer Changes
One maintainer commit, `ca9be18` "fix(gastown): un-blind boot's wisp reconcile
and resolve its burn target" — 4 files changed, 215 insertions(+), 17
deletions(-), covering every review item above. The branch was also rebased onto
current `main`, which required resolving one conflict in
`gastown/tests/test_gastown_pack_assets.sh`: two commits that landed on main
meanwhile added test functions at the same insertion point this PR uses. Both
sides were kept in full, verified as a purely additive resolution in both
directions with no lines dropped, and the PR's own guards were re-checked with
mutation arms against the refreshed base. Your commit `65baaf2` is unchanged and
retains your authorship.

### Final Review Status
Ready for the merge queue: final head `ca9be18` has passing required GitHub
checks — Check, CodeQL, and both Analyze jobs are green — and the merge-ready
metadata is in place. I am marking this PR `status/merge-ready` so the merge
queue can pick it up.

CI: https://github.com/gastownhall/gascity-packs/actions/runs/34318143038

### Review Iterations
3 review passes performed. Pass 1 gated on the blind reconcile queries, the empty
bootstrap assignee, and the missing boot-scoped guard. Pass 2 confirmed those
three were fixed and gated on the no-op burn target the fixes had introduced,
listing four further non-gating items. Pass 3 confirmed the gate was closed,
found all four voluntary follow-ups had been taken correctly, and approved with
two test-layer recommendations, both of which were then also taken.

---

PR #262, opened 2026-08-02
fix(discord): make launcher rooms usable and add attach-first @@handle

brandonmartin:
## Summary

Six interlocking fixes to the `discord` pack that were needed to get launcher-room ad-hoc sessions working end-to-end for a real workspace, plus a small feature so `@@handle` attaches to an already-running session instead of always spawning a new one.

Verified end-to-end in a real workspace: `@@gasburger.mayor` now routes to the running mayor session, `@@gc-packs/gasburger.crew` spawns/attaches under the correct rig, thread follow-ups reach the same session, and `gc discord reply-current` succeeds — none of which was possible before this change.

## Motivation

Every step of the launcher-room flow was broken for the naming convention this town actually uses (dotted, optionally rig-prefixed agent handles like `gasburger.mayor`, `gc-packs/gasburger.crew`). I traced each failure by watching `data/chat-ingress/*.json` receipts; each item below is what a receipt actually returned before the fix.

### Handle parsing — dotted handles were silently truncated
`AGENT_HANDLE_SEGMENT` (`^[a-z][a-z0-9_-]{0,31}$`) and the char class in `extract_agent_handles` rejected `.`, so `@@gasburger.mayor` parsed as `gasburger` and `@@gc-packs/gasburger.crew` parsed as `gc-packs/gasburger`. Every attempt returned `unknown_handle`. Both call sites now accept `.` inside a segment.

### Session identity after async create — hydration polled forever
`create_agent_session` posts `/v0/sessions` with `async: true`. That endpoint returns only `{status, request_id, event_cursor}` — no `id`/`alias`/`session_name`. Meanwhile the platform stores the alias with the pack namespace prepended (`dc-<hex>-slug` → `gasburger.dc-<hex>-slug`), so the exact-alias hydration in `ensure_room_launch_session_for_handle` never found the just-created session and eventually raised `created launch session is not routable yet`.

Added `resolve_launched_session_identity_eventually` / `_ready_` variants that locate the session by (`template == qualified_handle` **and** alias suffix-matches the requested alias). Safe because `room_launch_session_alias` embeds a sha256 digest of guild/channel/message/handle, so the requested-alias suffix is unique per launch.

### Attach-first launcher (feature)
`resolve_existing_session_for_handle` looks up existing routable sessions by `alias`, `session_name`, or `id` (case-insensitive). If the handle already names a running session, `process_room_launch_message` and `process_room_launch_thread_message` route to it directly; otherwise the previous template-and-spawn path runs unchanged. `ensure_room_launch_session_for_handle` gained an `attached_identity` kwarg that skips create + prime + hydration polling and marks the participant `attached=True`.

Lookup is best-effort: a `GCAPIError` from `list_city_sessions` is swallowed so the spawn path still runs if the API blips.

**Semantics to know:** `@@handle` attaches when a matching routable session exists, and only spawns when none does. To force a fresh spawn, close the existing session first — or use a handle that isn't currently a running session.

### Receipt targets carry full identity
Ingress receipt `targets[]` recorded only `session_name`, so `gc discord reply-current` failed with `no recent discord event with publish metadata found for <selector>` whenever `GC_SESSION_ID` (or alias) was the selector coming from env. New `participant_target_identity()` helper populates `session_name`, `session_id`, and `session_alias` on every launcher target dict (pending/delivered/failed, root and thread paths). Fixes attached **and** spawned sessions.

### Launcher-managed threads bypass the extmsg fabric
`_record_extmsg_inbound` treated every thread message as an extmsg event and returned True, so `process_room_launch_thread_message` never saw follow-ups posted inside a launcher-created thread. Added `launcher_claims_message(config, channel_id, parent_id)` mirroring the existing `bound_room_claims_message` bypass — true when `channel_id` or `parent_id` is a configured launcher room, or when `channel_id` matches a `room_launch` record's `thread_id`. Extmsg falls through in those cases so the launcher pack routes the follow-up.

### Thread name keeps the rig prefix
`room_launch_thread_name` did `rsplit("/", 1)[-1]` on the handle, so `@@gc-packs/gasburger.refinery` produced a Discord thread titled `gasburger.refinery` — indistinguishable from a root-level agent of the same name. Now retains the qualified form; still trimmed to Discord's 100-char thread-name limit.

## Test plan

- [x] `python3 -m unittest discover discord/tests` — 267/268 pass. The one pre-existing failure (`test_run_fix_dispatch_returns_bead_init_failure_without_slinging`) is an environment issue unrelated to this branch: the test hardcodes bare `gc` but this machine's PATH resolves it to `/home/b/go/bin/gc`.
- [x] Six new intake tests: dotted-handle extraction, prefixed-alias hydration under async create, four resolver cases for existing sessions, attach path skips create/prime and populates identity.
- [x] Two existing chat-script assertions updated for the new rig-prefixed thread-name format.
- [x] End-to-end verification in a live workspace: launcher room `@@gasburger.mayor` attached to the running mayor session; a follow-up in the launcher-created Discord thread also reached that same session; `gc discord reply-current` returned a non-empty `record.remote_message_id` and the reply appeared in Discord.

## Files touched

- `discord/scripts/discord_intake_common.py` — regex, resolvers, attach path, thread-name helper
- `discord/scripts/discord_gateway_service.py` — attach-first branches, target-identity helper, extmsg bypass
- `discord/tests/test_discord_intake_common.py` — 6 new tests
- `discord/tests/test_discord_chat_scripts.py` — 2 assertion updates

### DIFF
--- a/discord/scripts/discord_gateway_service.py
+++ b/discord/scripts/discord_gateway_service.py
@@ -90,6 +90,20 @@ def participant_delivery_selector(participant: dict[str, Any]) -> str:
     return ""
 
 
+def participant_target_identity(participant: dict[str, Any]) -> dict[str, str]:
+    """Build the identity dict written into ingress receipt `targets` entries.
+    Including every known identifier lets find_latest_discord_reply_context /
+    reply-current match by whichever selector (id, name, alias) the caller
+    presents from GC_SESSION_* env vars.
+    """
+    fields: dict[str, str] = {}
+    for key in ("session_name", "session_id", "session_alias"):
+        value = str((participant or {}).get(key, "")).strip()
+        if value:
+            fields[key] = value
+    return fields
+
+
 def json_response(handler: BaseHTTPRequestHandler, status: int, payload: dict[str, Any]) -> None:
     body = json.dumps(payload, indent=2, sort_keys=True).encode("utf-8") + b"\n"
     handler.send_response(status)
@@ -695,6 +709,26 @@ def bound_room_claims_message(
     parent = str(parent_id).strip()
     if parent and explicit_room_binding(config, parent, app_name):
         return True
+    return False
+
+
+def launcher_claims_message(config: dict[str, Any], channel_id: str, parent_id: str = "") -> bool:
+    """True if this message belongs to launcher-managed routing: the message
+    is in a configured launcher room OR is inside a launcher-managed thread
+    (parent is a launcher OR channel_id matches a room_launch thread_id).
+    Extmsg's generic thread handler must skip these so the launcher pack's
+    process_room_launch_thread_message can run.
+    """
+    channel = str(channel_id).strip()
+    parent = str(parent_id).strip()
+    if channel and common.resolve_room_launcher(config, channel):
+        return True
+    if parent and common.resolve_room_launcher(config, parent):
+        return True
+    if channel:
+        launch = common.load_room_launch(common.room_launch_record_id(channel))
+        if launch and str(launch.get("thread_id", "")).strip() == channel:
+            return True
     return False
 
 
@@ -1388,7 +1422,27 @@ def process_room_launch_message(
             return {"status": "ignored_untargeted", "ingress_id": ingress_id, "receipt": receipt}
         used_default_handle = True
 
-    if used_default_handle:
+    # Attach-first: if the handle matches an already-running session
+    # (by alias/session_name/id), route to that session directly instead
+    # of resolving as a template + spawning a fresh clone.
+    attached_identity: dict[str, str] = {}
+    try:
+        attached_identity = common.resolve_existing_session_for_handle(requested_handle)
+    except common.GCAPIError as exc:
+        receipt = persist_ingress_receipt(
+            {
+                **base_receipt,
+                "binding_id": str(launcher.get("id", "")).strip(),
+                "status": "failed_lookup",
+                "reason": str(exc),
+                "targets": [],
+            }
+        )
+        return {"status": "failed_lookup", "ingress_id": ingress_id, "receipt": receipt}
+
+    if attached_identity:
+        qualified_handle, resolve_error = requested_handle, ""
+    elif used_default_handle:
         qualified_handle, resolve_error = requested_handle, ""
     else:
         try:
@@ -1442,7 +1496,7 @@ def process_room_launch_message(
         }
     )
     try:
-        launch = common.ensure_room_launch_session(launch)
+        launch = common.ensure_room_launch_session(launch, attached_identity=attached_identity or None)
     except (ValueError, common.GCAPIError) as exc:
         receipt = persist_ingress_receipt(
             {
@@ -1457,6 +1511,7 @@ def process_room_launch_message(
         return {"status": "failed_lookup", "ingress_id": ingress_id, "receipt": receipt}
 
     target_selector = participant_delivery_selector(launch)
+    target_identity = participant_target_identity(launch)
     envelope = build_room_launch_envelope(
         launcher=launcher,
         launch=launch,
@@ -1479,6 +1534,7 @@ def process_room_launch_message(
             "qualified_handle": qualified_handle,
             "targets": [
                 {
+                    **target_identity,
                     "session_name": target_selector,
                     "status": "pending",
                     "intent": "default",
@@ -1538,9 +1594,10 @@ def process_room_launch_thread_message(
         return {"status": "rejected_targeting", "ingress_id": ingress_id, "receipt": receipt}
     target_handle = ""
     routing_mode = ""
+    thread_attached_identity: dict[str, str] = {}
     if mentioned_handles:
         try:
-            qualified_handle, resolve_error = common.resolve_agent_handle(mentioned_handles[0])
+            thread_attached_identity = common.resolve_existing_session_for_handle(mentioned_handles[0])
         except common.GCAPIError as exc:
             receipt = persist_ingress_receipt(
                 {
@@ -1552,20 +1609,37 @@ def process_room_launch_thread_message(
                 }
             )
             return {"status": "failed_lookup", "ingress_id": ingress_id, "receipt": receipt}
-        if resolve_error:
-            receipt = persist_ingress_receipt(
-                {
-                    **base_receipt,
-                    "binding_id": str(launcher.get("id", "")).strip(),
-                    "status": "rejected_targeting",
-                    "reason": resolve_error,
-                    "mentioned_handles": mentioned_handles,
-                    "targets": [],
-                }
-            )
-            return {"status": "rejected_targeting", "ingress_id": ingress_id, "receipt": receipt}
-        target_handle = qualified_handle
-        routing_mode = "explicit_handle"
+        if thread_attached_identity:
+            target_handle = mentioned_handles[0]
+            routing_mode = "explicit_handle_attached"
+        else:
+            try:
+                qualified_handle, resolve_error = common.resolve_agent_handle(mentioned_handles[0])
+            except common.GCAPIError as exc:
+                receipt = persist_ingress_receipt(
+                    {
+                        **base_receipt,
+                        "binding_id": str(launcher.get("id", "")).strip(),
+                        "status": "failed_lookup",
+                        "reason": str(exc),
+                        "targets": [],
+                    }
+                )
+                return {"status": "failed_lookup", "ingress_id": ingress_id, "receipt": receipt}
+            if resolve_error:
+                receipt = persist_ingress_receipt(
+                    {
+                        **base_receipt,
+                        "binding_id": str(launcher.get("id", "")).strip(),
+                        "status": "rejected_targeting",
+                        "reason": resolve_error,
+                        "mentioned_handles": mentioned_handles,
+                        "targets": [],
+                    }
+                )
+                return {"status": "rejected_targeting", "ingress_id": ingress_id, "receipt": receipt}
+            target_handle = qualified_handle
+            routing_mode = "explicit_handle"
     if not target_handle and reply_to_id:
         target_handle = common.room_launch_message_target_handle(launch, reply_to_id)
         if target_handle:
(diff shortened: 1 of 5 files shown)


brandonmartin:
Follow-up commit `59bbbb9` — **feat(discord): launcher wakes declared named sessions on first `@@handle`**

Extends the attach-first routing so `@@handle` also wakes a declared `[[named_session]]` (e.g. `mode = "on_demand"` crew members) when no running session matches yet. Previously the handle would fall through to template resolution and get rejected as `unknown_handle` — you had to wake the session via `gc sling` / `gc session nudge` before Discord could talk to it.

**New routing order in `process_room_launch_message` / `_thread_message`:**
1. attach if a running session matches *(existing behavior)*
2. **new:** if the handle names a declared `[[named_session]]` (rig-qualified `<rig>/<name>` or city-scoped `<name>`), spawn it via that declaration's template + alias
3. otherwise fall through to the previous template-and-spawn path

**Named-session spawn detail**
- `resolve_named_session_for_handle` parses `city.toml` `[[named_session]]` entries and returns `{template, alias, spawn_template, scope, dir}`. `spawn_template` is rig-qualified (`<dir>/<template>`) so the platform resolves the correct rig-instance of the pack template.
- `ensure_room_launch_session[_for_handle]` gained `spawn_template_override` and `session_alias_override` kwargs. When provided, the launcher spawns via that qualified template with the declared alias (e.g. `fattony`) instead of the per-message `dc-<hash>-slug` alias. This keeps the spawned session recognizable so subsequent turns attach to the same instance and don't repeatedly spawn per-message clones.
- Suffix-match hydration keys off the override too, so it locates the created session regardless of the rig/pack prefixing the platform wraps around the requested alias.

**Attach after spawn**
- Verified by direct `/v0/sessions` probe: the stored alias for a rig-scoped named session comes back as `<rig>/<pack>.<alias>` (e.g. `daytripper/gasburger.fattony`). `resolve_existing_session_for_handle` now also matches rig-prefixed aliases: when the handle is `<rig>/<tail>`, it accepts a session whose alias starts with `<rig>/` and ends with `<tail>` or `.<tail>`. The rig prefix on both sides prevents cross-rig collision.

**Tests** — 7 new intake cases:
- rig-prefixed alias suffix match (positive + cross-rig negative)
- `resolve_named_session_for_handle`: rig-scoped happy path, case-insensitivity, empty-on-miss, best-effort when `city.toml` is unreadable
- `ensure_room_launch_session` spawn path with override kwargs asserts `create_agent_session` is called with the qualified template and the declared alias, and that suffix-hydration correctly resolves the rig-prefixed stored alias

All 114 intake + 78 gateway tests pass. Only pre-existing failure remains (`test_run_fix_dispatch_returns_bead_init_failure_without_slinging`, environment-dependent path resolution — noted in the PR body).

**Caveat**: the spawn only actually completes when the rig isn't suspended. Suspended rigs park the session in `start-pending`; the launcher receipt still records `delivered` (the message is queued on the platform), but the session won't come up until the rig is resumed.

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277293449

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277293518

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277310383

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277316852

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @brandonmartin! This PR makes Discord launcher rooms actually usable: it fixes the routing that left launcher rooms inert, adds attach-first `@@handle` resolution so a handle that already has a live session gets attached to instead of re-spawned, and teaches the launcher to wake declared `[[named_session]]` agents on their first `@@handle`. That matters because it turns the launcher from spawn-always into something that reuses live bodies, and it makes pack-qualified handles like `@@gasburger.mayor` route to the intended agent instead of falling through to extmsg's generic thread handler.

### Original PR Review

Specific gaps fixed:
- **Trailing-dot handle extraction.** The first pass caught a real regression: `"ask @@sky."` extracted `sky.` rather than `sky`, so ordinary sentence punctuation broke handle routing. Fixed at both layers — `AGENT_HANDLE_SEGMENT` now refuses a trailing dot, and the extractor `rstrip(".")`s each extracted part — with `"@@corp/sky..."`, `"@@sky. and @@sky"` (collapsing to one seat), and `"@@gasburger.mayor"` (intact, so the headline pack-qualified routing is unaffected) all pinned. Reverting either layer alone reds its own tests; reverting both reds all four.
- **Named-session resolution.** A `scope = "rig"` entry with no `dir` silently registered city scope with an unqualified `spawn_template`, and city-scope/slashless handles were unreachable through the pack-tail tolerance. The resolver now expands a dir-less rig entry across the city's configured rigs — matching gascity's own `ExpandGenericRigNamedSessions`, including its no-rigs skip — and attach lookup and publish-side re-pin now share one `_alias_matches_qualified_handle` policy, so the two can no longer disagree. Round-trip coverage (spawn → stored alias → next-turn attach) landed for both scopes.
- **Hydration latency under the launch advisory lock.** The suffix match ran on its own second deadline, so a pack-prefixed alias cost an extra 30s while holding the launch lock. Both polls now attempt exact and suffix matching in the same iteration under one deadline, behind an injectable `_poll_timeout` seam, with the 30.0/90.0 defaults pinned as literals and the wall-clock sleeps removed from the hydration tests.
- **Unpinned gate predicates.** `launcher_claims_message` and `_alias_matches_qualified_handle` had no test that failed if the guard were deleted or inverted. Both are now pinned to that bar, with live positive controls that stay green under the deletion (so the controls carry signal rather than decoration), per-condition unit tests, and a cross-rig negative at both unit and publish level — a rig-B participant must not be re-pinned by a rig-A publisher.

Review findings addressed:
- **Test evidence, minor:** the ready-side half of the new suffix fold was one mutation from green — deleting it left the whole suite passing. It is now pinned at the unit seam by `test_resolve_ready_session_candidate_eventually_matches_prefixed_alias_by_suffix`, plus a `..._suffix_still_requires_ready` sibling that also pins the `ready=True` argument the poll passes. Commit `b7cdb1d`. One note worth recording: the end-to-end test shape the review proposed turned out to be inert here — the identity poll's twin hydrates the participant first, so an ensure-level test in that shape passes unchanged with the block deleted. The commit keeps that measurement in a comment above the test so the pin does not get "simplified" back into a no-op.
- **Debuggability, nit:** `launcher_claims_message` condition 3 reads a launch record per thread message where its sibling is a pure config lookup. That read is deliberate — it is the only way to catch threads that outlive their launcher's config entry — and now carries a comment saying so, so a later reader does not narrow it back to config-only and silently re-open the gap. Commit `b7cdb1d`.

Non-blocking follow-ups (shared for visibility — NOT completed in this PR). Each is filed so it does not get lost:
- `mc-gl5p6` — contract fidelity, minor: bound-room receipt targets still carry `session_name` only, so they sit outside the identity contract this PR's own `participant_target_identity` docstring declares; an agent replying from a bound-room context whose transcript path fails can miss a delivery the launcher route would have matched. Pre-existing region, newly *exposed* by the contract this PR introduced — which is exactly why it did not gate.
- `mc-1v0u1` — architectural consistency, minor: `_find_launched_session_by_suffix` is still case-sensitive, accepts three separators, and applies no rig guard, and its digest-uniqueness docstring does not hold for the `session_alias_override` alias form (no sha). Folds into the shared tail-match helper below.
- `mc-58amr` — correctness, minor: the named-session resolver does not read the declared `mode`; for `always` entries it should prefer attach and decline to spawn rather than race the supervisor's instance.
- `mc-um4dk` — resilience, minor: make the publish-side participant re-pin best-effort, and/or move it after `save_chat_publish`, so a re-pin failure cannot strand a posted-but-unrecorded reply.
- `mc-trkd1` — scope, minor: manually created threads inside launcher rooms are handled by neither pipeline; document the launcher-owns-its-room policy or narrow condition 2 to threads with a launch record.
- `mc-yt9u8` — scope, minor: attach-first fires on the implicit default handle with no config signal; gate it on an explicit mention or document the new default-handle attach semantics.
- `mc-szgb6` — correctness, minor: ambiguous same-rig tail collisions have no stated policy, and the thread handler prefers an attach lookup over an already-stored routable participant.
- `mc-ebib6` — scope, minor: the `@`-family surfaces never learned the dot; extend `ALIAS_PATTERN`/`resolve_at_mentions` with the dot and tail tolerance, or state in the README that dotted named-session aliases are `@@`-handle-only.
- `mc-nsxdn` — documentation, minor: the README launcher section still describes the pre-PR spawn-always behavior (attach-first precedence, `[[named_session]]` wake and alias shape, dotted handles, and the default-handle interaction all need a pass).
- `mc-9t1wz` — error handling, nit: the two `GCAPIError` arms around the attach lookup are dead because the resolver already swallows; pick one owner — drop the try/excepts, or stop swallowing in the resolver and let the call sites emit the promised receipts.
- `mc-mg6ce` — operability, nit: the attach-skip-primer decision is not persisted and `participant["attached"]` is written but never consumed; persist or remove.

### Maintainer Changes
One maintainer commit on top of your three, `fix(discord): address review blockers on launcher routing (maintainer fixup)` — 4 files changed, 837 insertions(+), 115 deletions(-) — carrying the four review blockers above plus the two second-pass items. Your commits and authorship are preserved; the branch was rebased onto current `main` (`3c15b64`) with the patch content byte-identical to the reviewed tree. Full change set: 5 files changed, 1639 insertions(+), 54 deletions(-), all under `discord/`. The focused suite is green on the final rebased head — `discord/tests`, 414 passed.

### Final Review Status
Ready for the merge queue: final head `b7cdb1d` has passing required GitHub checks, and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/gascity-packs/actions/runs/34318242630

### Review Iterations
2 review passes. Pass 1 (three reviewers, full cross-file coverage) returned request_changes with four required changes; the maintainer fixup addressed all four, and pass 2 re-verified each one against the shipped source with deletion and inversion probes before returning approve. Pass 2 raised four new items, all minor or nit: two were taken into the fixup, and two are filed as follow-ups above.

---

PR #264, opened 2026-08-03
fix(gastown): pour refinery, witness and deacon patrol wisps as in_progress and resolve them via bd query

edwarddavison:
## Problem

Every patrol role — refinery, witness, deacon — pours its next wisp and then assigns it:

```bash
NEXT=$(gc bd mol wisp mol-<role>-patrol --root-only ... | jq -r '.new_epic_id')
gc bd update "$NEXT" --assignee="$GC_AGENT"
```

Two independent mechanisms make that leak wisps, and fixing either one alone is not enough.

**1. The wisp is never transitioned.** `gc bd mol wisp` pours at `status=open`, and `--assignee` alone does not move it. The startup resume check filters explicitly on `status=in_progress`, so **a wisp poured by the documented snippet is invisible to its own resume check.** A restarted session finds nothing in progress, pours another wisp, and orphans the previous one — open, assigned, never burned, never resumed.

**2. The reconciliation cannot see wisps at all.** Wisp roots are ephemeral rows, and `gc bd list` excludes ephemeral beads regardless of `--type`. Every list-based lookup — current-wisp resolution, surplus reconciliation, queued-successor reuse — silently resolves empty. So the block that exists to keep exactly one wisp finds no surplus to burn and no successor to reuse, and pours a fresh one every cycle. This is the #1884 wisp-leak pattern.

The same defect is in the **agent prompt templates**, not just the formulas — and the prompt is what actually bootstraps the first wisp and runs the no-idle-state reconciliation.

## Verified behaviour

Against a live gas-city ledger carrying **two leaked witness wisps**:

| Command | Result |
|---|---|
| `gc bd query 'ephemeral=true'` | returns both wisps: `status=open`, `ephemeral=true`, `issue_type=molecule`, assigned to the witness |
| `gc bd list --assignee=<witness> --status=open --type=molecule` | **0 rows** |
| `gc bd list --status=open --type=molecule` (no assignee filter) | **0 rows** |
| `gc bd mol wisp --root-only` | `status=open`, `assignee=null` |
| `gc bd update <id> --assignee=X` | `status=open` — unchanged |
| `gc bd update <id> --assignee=X --status=in_progress` | `status=in_progress` ✅ |
| `gc bd update <id> --claim` (assignee matches, status open) | errors `already claimed`, status stays `open` |

Dry-run of the old vs new witness reconcile against that live pair:

```
OLD:  OPEN_WISPS=[]            -> NEXT=[]        => pours a duplicate every cycle
NEW:  current=gci-wisp-8kq (adopted+promoted)
      surplus (excl self)=[gci-wisp-bvr]
      -> NEXT = gci-wisp-bvr   (reused, no pour)
      -> burn = gci-wisp-8kq   => 2 wisps become 1, leak drains
```

## Changes

Across all three patrol roles, in both the formula and the agent prompt template:

1. **Every pour assigns with `--status=in_progress`** — including the bootstrap snippets in the formula headers and the prompt startup protocols.
2. **Every wisp lookup goes through `gc bd query` on `ephemeral=true`**, filtered by assignee and exact formula title. (The earlier `--type=wisp` spelling was worse still: `wisp` is not a valid issue type, so the command hard-errors.)
3. **The witness reconcile excludes the wisp it is executing, by id.** Once wisps are correctly `in_progress`, a status-only scan matches the current wisp — it would be reused as the successor and then burned on the next line, leaving the witness holding nothing. `next-iteration` now resolves the current wisp first and burns `$CURRENT_WISP` instead of a `<this-wisp-id>` placeholder.
4. **A repair path** adopts and promotes a wisp left at `status=open` by a session predating this fix. Without it, every already-deployed patrol wedges on an unresolvable current wisp after the pin bump instead of draining its existing leak.
5. **Header notes** explain why `--status=in_progress` is load-bearing and why `--claim` is not a safe substitute.

## Validation

- `test_patrol_wisp_pour_is_resumable` in `gastown/tests/test_gastown_pack_assets.sh` covers all six assets. Mutation-tested: it fails independently when *each* role's formula is reverted, when *each* role's prompt template is reverted, when the witness self-exclusion is removed, and when a single lookup is put back to `gc bd list`.
- All `gastown/tests/test_*.sh` pass.
- `tests/test_no_bare_bd_commands.py` passes (every new call routes through `gc bd`).
- `validate_registry.py --require-git` passes (`registry.toml: ok`); `registry.toml` is untouched.
- All three formulas parse under `tomllib` (9 / 5 / 8 steps intact).
- `bash -n` over all 14 wisp-handling shell blocks, with template placeholders neutralized.

## Relationship to other open PRs

This overlaps work already in your queue, and I would rather you merge whichever shape you prefer than have three variants of the same block sitting open:

- **#226** (`fix(gastown): reconcile patrol wisps across restarts`) covers the same three formulas and prompts with a much larger shared patrol-ledger contract. If you prefer that shape, this PR should be closed in its favour — it is the more thorough change. This one is the surgical version, with the live evidence above.
- **#189** and **#253** both fix witness self-burn in the same `next-iteration` block. Both keep `gc bd list --status=open --type=molecule`, which returns zero rows for ephemeral wisps — so their reconciliation operates on an always-empty list. #189 additionally states as an invariant that "a patrol wisp stays `status=open` for its entire working life"; that is true today only *because* of the defect this PR fixes. The self-exclusion insight in #253 is correct and is carried here, with credit.

Happy to split this back into refinery-only, rebase onto any of the above, or close it — whichever unblocks the queue fastest.

## Not in this change

- `mol-deacon-patrol`'s prompt bootstrap assigns with `$GC_ALIAS` while its reconcile reads `$GC_AGENT`; a mismatch there is the same leak by a different route, but it belongs with the identity-resolution work in #280 / #282.
- Fixing the class in the binary instead of per formula: make `gc bd mol wisp --root-only` default to `in_progress`, or make `--claim` idempotent when assignee matches but status differs.

Filed from a production deployment as gci-ly4 (refinery) and gci-5fa (witness + deacon).

https://claude.ai/code/session_0182tfjiB7nbSt7tJvfnEuZM

### DIFF
--- a/gastown/agents/deacon/prompt.template.md
+++ b/gastown/agents/deacon/prompt.template.md
@@ -61,8 +61,10 @@ Your formula: `mol-deacon-patrol`
 gc mail inbox
 
 # Step 3: Still nothing? Create patrol wisp (root-only — no child step beads)
+# --status=in_progress is load-bearing: a wisp poured at status=open is
+# invisible to the next restart's resume check, which pours a duplicate.
 NEW_WISP=$(gc bd mol wisp mol-deacon-patrol --root-only --var binding_prefix={{ .BindingPrefix }} --json | jq -r '.new_epic_id')
-gc bd update "$NEW_WISP" --assignee="$GC_ALIAS"
+gc bd update "$NEW_WISP" --assignee="$GC_ALIAS" --status=in_progress
 
 # Step 4: Read the formula recipe — these are the steps to execute
 # (Use 'gc bd formula show' for the recipe on disk; 'gc bd mol show' is
@@ -87,18 +89,24 @@ without running `next-iteration` (crash recovery or formula misread).
 If `next-iteration` already ran, do not pour again; run `gc hook`.
 
 ```bash
+# Wisp roots are ephemeral rows, which `gc bd list` hides unless it is given
+# --include-infra; the flagless lookup below used to return zero rows even
+# with wisps open. `gc bd query` selects the ephemeral tier directly, so there
+# is no flag left to forget.
+# ASSIGNED_WISP excludes $CURRENT_WISP by id — the wisp you are executing is
+# burned below, never mistaken for an already-queued successor.
 CURRENT_WISP=${GC_BEAD_ID:-}
 if [ -z "$CURRENT_WISP" ]; then
-  CURRENT_WISP=$(gc bd list --assignee="$GC_AGENT" --status=in_progress --type=molecule --limit=1 --json | jq -r '.[0].id // empty')
+  CURRENT_WISP=$(gc bd query --json 'ephemeral=true AND status=in_progress' --limit=0 | jq -r --arg id "$GC_AGENT" --arg f mol-deacon-patrol '[.[] | select((.assignee // "") == $id and (.title // "") == $f)] | .[0].id // empty')
 fi
-ASSIGNED_WISP=$(gc bd list --assignee="$GC_AGENT" --status=open --type=molecule --limit=1 --json | jq -r '.[0].id // empty')
+ASSIGNED_WISP=$(gc bd query --json 'ephemeral=true AND (status=open OR status=in_progress)' --limit=0 | jq -r --arg id "$GC_AGENT" --arg f mol-deacon-patrol --arg self "$CURRENT_WISP" '[.[] | select((.assignee // "") == $id and (.title // "") == $f and .id != $self)] | sort_by(.created_at) | .[0].id // empty')
 if [ -n "$CURRENT_WISP" ] && [ -z "$ASSIGNED_WISP" ]; then
   NEXT=$(gc bd mol wisp mol-deacon-patrol --root-only --var binding_prefix={{ .BindingPrefix }} --json | jq -r '.new_epic_id // empty')
   if [ -z "$NEXT" ]; then
     echo "Could not pour next deacon wisp; not burning."
     exit 1
   fi
-  if ! gc bd update "$NEXT" --assignee="$GC_AGENT"; then
+  if ! gc bd update "$NEXT" --assignee="$GC_AGENT" --status=in_progress; then
     echo "Could not assign next deacon wisp; not burning."
     exit 1
   fi
@@ -111,7 +119,7 @@ elif [ -z "$ASSIGNED_WISP" ]; then
     echo "Could not bootstrap next deacon wisp."
     exit 1
   fi
-  if ! gc bd update "$NEXT" --assignee="$GC_AGENT"; then
+  if ! gc bd update "$NEXT" --assignee="$GC_AGENT" --status=in_progress; then
     echo "Could not assign bootstrap deacon wisp."
     exit 1
   fi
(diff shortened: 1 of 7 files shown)


edwarddavison:
Consolidated review request from the Below Baseline deployment — we have six PRs open across `gascity-packs` and `gascity`, and two of them fix defects that are live in our production city today. Happy to split, rebase or reduce scope on any of them if that makes review easier.

**Production-affecting (these two are why I'm pinging):**

- **packs#264** (this PR) — refinery patrol wisps are poured with `--assignee` but never `--status=in_progress`, while the Step-1 resume path queries `--status=in_progress --type=wisp`. So a patrol can never re-find its own wisp after a restart. We see this in the wild as duplicate wisp pours that agents work around by hand. `mol-witness-patrol` and `mol-deacon-patrol` carry the identical defect — happy to extend this PR to cover all three if you'd prefer one change over two.
- **packs#261** — `boot` is the only `mode = "always"` agent shaped as a single pass, so `drain-ack` + `exit` is read by the controller as "re-materialise me now" with no re-wake cadence. Our instance logged ~500 tool calls/hour continuously and 3,412 assistant messages in one day before we drained it; it is currently switched off rather than fixed.

**Ready, not urgent:**

- **packs#260** — `mol-port-review`, a 3-stage cross-rig port → review → publish formula.
- **gascity#4913** — regression guard for `--on` source-bead routing, so the cold-pool demand-starvation divergence can't silently return.
- **gascity#4905** — work-formula selection guide plus v2 `--on` semantics.

**Oldest, and probably needs a decision rather than a review:**

- **packs#131** — `mol-digest-generate` dispatch docs plus a window dedup guard. Open since 22 June. If it's not wanted, closing it is a perfectly good outcome and we'll stop carrying it.

All six report `MERGEABLE` as of today. We also left evidence comments on gascity#4365, gascity#4387 and packs#191 rather than opening duplicate reports.

Context on why merges matter to us specifically: we build `gc`/`bd` from source and run a pinned pack snapshot, so a fix only reaches our box after it merges here and we bump the pin. Until then we're carrying the defect knowingly. No expectation of urgency on your side — just flagging that the two above are the ones with a live cost.

edwarddavison:
@julianknutsen @sjarmak — consolidated review request: 5 `gascity-packs` PRs + 2 `gascity` PRs, all from one production deployment. Apologies for the batch ping; happy to split, rebase, or rework any of them.

Context: we run a multi-rig production install built from source (gc 1.4.0, packs pinned at `33d3a430`). Each PR fixes or guards a defect we hit live; repro evidence is on the linked issues, and we've added evidence comments to your existing #4365, #4387 and packs#191 rather than filing duplicates.

In priority order from our side:
1. **packs#264** — patrol wisps pour as `in_progress`; resumed refinery/witness/deacon patrols lose wisp continuity and double-pour. Live P1 for us (same class as packs#252/#145).
2. **packs#261** — boot always-mode respawn loop; missing backoff. Live P1 — ~20% of one day's token burn before we drained boot.
3. **packs#273** — polecat push gate falls through to `git push` when a prose no-push rail has no `auto_push` metadata; makes it fail closed. Guards customer repos.
4. **packs#260** — `mol-port-review`, 3-stage cross-rig review formula.
5. **packs#131** — `mol-digest-generate` dispatch docs + window-dedup guard (open since June).
6. **gascity#4913** — cold-pool regression test for `--on` source-bead routing.
7. **gascity#4905** — work-formula selection guide + v2 `--on` semantics.

All seven are green and mergeable as of 2026-08-06. If any need changes we'll turn them around quickly — and if you'd rather we stage them differently, say so. Thanks for gascity; it's carrying a real production city well.

PR #267, opened 2026-08-03
fix: scope patrol-project-leads as a city order to stop per-rig duplicate patrols

sjarmak:
## Summary
- scope `patrol-project-leads` as a city order so imported rigs do not repeat the same citywide patrol
- preserve `escalate-rollups` and unrelated unscoped orders as rig-scoped identities
- add an installed-`gc` integration regression and CI contract check that prove exactly one complete patrol record

## Why
`patrol-project-leads` enumerates all active project leads on every invocation. With the default rig scope, the same source was materialized and executed once for the city plus once per importing rig. This created duplicate global work and avoidable process churn. GasCity order discovery is behaving as designed; the pack must declare this particular order city-scoped.

## Test plan
- `python3 -m pytest -q`
- `GC_TEST_BIN=$(command -v gc) python3 -m pytest oversight-rig/tests/test_order_scopes.py -q`
- `python3 -m pytest tests/test_ci_workflow.py -q`
- `go test ./...`
- pack lint, Ruff, py_compile, and diff checks

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -61,6 +61,7 @@ jobs:
                       pr-pipeline slack-channel slack-full slack-mini; do
             "$GC_BIN" lint "$pack"
           done
+          GC_TEST_BIN="$GC_BIN" python3 -m pytest oversight-rig/tests/test_order_scopes.py -q
           GC_TEST_BIN="$GC_BIN" python3 -m pytest tests/test_gc_role_prompt_integration.py tests/test_pack_composition_warnings.py -q
 
       # Installed, not parsed. `gc lint` above reads each pack on its own and
--- a/oversight-rig/orders/escalate-rollups.toml
+++ b/oversight-rig/orders/escalate-rollups.toml
@@ -1,5 +1,10 @@
 [order]
 description = "Watch for severity:escalate rollup beads and deliver them. Pure mechanical — no agent decision."
+# Deliberately NOT city-scoped, unlike patrol-project-leads: deliver-rollup.sh
+# routes each bead by its own `rig:` label and is idempotent per bead
+# (idempotency_key `rollup-<id>`, plus an `--add-label delivered` on success),
+# so overlapping per-rig instances converge on the same delivery instead of
+# fanning out. oversight-rig/tests/test_order_scopes.py pins this fan-out.
 trigger = "condition"
 check = "bash $GC_PACK_DIR/assets/scripts/has-undelivered-escalates.sh"
 exec = "bash $GC_PACK_DIR/assets/scripts/deliver-rollup.sh"
--- a/oversight-rig/orders/patrol-project-leads.toml
+++ b/oversight-rig/orders/patrol-project-leads.toml
@@ -1,5 +1,6 @@
 [order]
 description = "Nudge every project-lead to run a triage tick. Cooldown sets the cadence."
+scope = "city"
 trigger = "cooldown"
 interval = "15m"
 exec = "bash $GC_PACK_DIR/assets/scripts/nudge-project-leads.sh"
--- a/oversight-rig/orders/patrol-project-leads.toml
+++ b/oversight-rig/orders/patrol-project-leads.toml
@@ -1,5 +1,13 @@
 [order]
 description = "Nudge every project-lead to run a triage tick. Cooldown sets the cadence."
+# City-scoped: nudge-project-leads.sh enumerates every active project-lead
+# session city-wide (`gc session list` filtered on template + state, no rig
+# filter), so it already covers the whole city in one run. Left unscoped, each
+# per-rig pack import registers its own instance, and every one of them fires
+# on its own independent cooldown -- N rigs means each lead is nudged N+1 times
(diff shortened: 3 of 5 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277293374

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277310683

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @sjarmak! This PR scopes `patrol-project-leads` as a city order, so a city that imports `oversight-rig` per rig registers one patrol instance instead of one per rig plus the city. That matters because `nudge-project-leads.sh` already enumerates every active project-lead session city-wide with no rig filter, and every extra instance fires on its own independent 15m cooldown — left unscoped, each lead gets nudged N+1 times per tick. You also brought a regression test *and* a CI pin for it, which is why a one-line fix landed as a defended one.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- The duplicate-nudge behavior is fixed at the right layer and measured rather than assumed: with `scope = "city"` the order materializes exactly one instance, and deleting the line turns the new test red with a diagnostic naming all three identities (`assert 3 == 1`, `(None, alpha, beta)`).
- The review confirmed the premise the PR rests on, which was not obvious from the repo: per-rig instances of an *exec* order really do fire, unlike `gastown/orders/digest-generate.toml`, whose per-rig instances never reach the city-level pool. So this is duplicated work, not a stuck order — a distinction now recorded in the file itself.
- A typo'd scope value does not fail loudly: with `scope = "citty"`, `gc order list` prints `unknown scope` to stderr and still exits 0, dropping the order from the schedule entirely. Your test is the only CI-side guard that catches that, because `ci.yml` deliberately excludes `oversight-rig` from the `gc lint` loop.
- The regression test cannot pass vacuously. The `escalate-rollups` / `rig-health` assertions are a live negative control: removing the fixture's per-rig imports leaves `assert len(patrol) == 1` passing and fails on those pins instead.
- Release surface is clean — `validate_registry.py --require-git` is green on the PR tree, so this correctly ships without a registry change.

Review findings addressed:
- Test evidence quality — the CI pin asserted a substring of the step's raw `run:` block, so it stayed green when the invocation was commented out or suffixed `|| true`, the two likeliest ways a CI line actually gets disabled. Fixed in `4d77468`: the pin now matches the step's *active* (non-comment) lines and rejects a short-circuited invocation. Mutation-verified in both directions, including a legitimate flag reflow that must stay green.
- Debuggability — `workflow["jobs"]["check"]` raised a bare `KeyError` on a job rename, inside the file whose stated purpose is failing actionably. Fixed in the same commit; a rename now reports `no job 'check' in CI; found: ['checkx']`.
- Architectural consistency — the new suite shelled out to the real `gc` with the caller's environment inherited, while the repo's established live-gc harness strips every `GC_*`/`BEADS_*`/`XDG_*` key and pins `HOME` plus the XDG set (and carries an in-code writeup of the incident behind that). Benign today because `order list` is a pure config scan; `4d77468` passes an explicit `env=` so the first store-touching assertion added there does not inherit the failure mode silently.
- Maintainability — `scope = "city"` carried no why-comment, and the repo's only precedent documents the opposite mechanism, so a copied comment would have been wrong. `4d77468` adds one written from patrol's own mechanism, plus a companion note on `escalate-rollups.toml` recording that *its* fan-out is deliberate and safe because delivery is idempotent per bead.

- `mc-3kgts` — nit, test evidence quality: `oversight-rig/tests/test_order_scopes.py` holds a third verbatim copy of the live-gc skip/fail fixture gate (byte-identical to `tests/gc_live_city.py` and `tests/test_gc_role_prompt_integration.py`). Left as-is deliberately, since pack tests are self-contained and the shared helper is not importable from a pack test directory; worth extracting a per-pack conftest if a fourth copy appears.
- `mc-21gly` — nit, operability: the CI step "Lint and exercise shared role prompt composition" now also runs the order-scope regression, so its name no longer covers its contents. A rename has to update the step-name pin in `tests/test_ci_workflow.py` in the same commit, which is why it was not folded in here.
- `mc-t3y1s` — registry consumers pinned at oversight-rig 0.1.0 keep the duplicated patrol until a 0.1.x release entry is cut. That is a separate publishing act per repo practice, not something this PR should carry.

### Maintainer Changes
`test: harden the order-scope CI pin and make its suite hermetic` (`4d77468`) — 4 files changed, 58 insertions(+), 4 deletions(-). Your three commits are unchanged and preserved with original authorship; the branch was rebased onto current `main` before adoption.

### Final Review Status
Ready for the merge queue: final head `4d77468` has passing required GitHub checks, and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/gascity-packs/actions/runs/34259814871

### Review Iterations
1 review pass across three lanes (design-implementation, implementation, and design-test-risk). No blockers and no majors; all six findings were minor or nit, four were fixed in a single maintainer commit, and the remaining two are filed as follow-ups above.

---

PR #269, opened 2026-08-04
fix(slack-full): make the adapter build and pass tests on darwin (syscall.Openat is Linux-only)

outdoorsea:
`slack-full`'s adapter does not compile on macOS, and fixing that exposes three test failures that were previously unobservable there.

## 1. The build failure

```
./confined_open.go:46:24: undefined: syscall.Openat
```

`syscall.Openat` is Linux-only. darwin's `syscall` package is libSystem-based and exports neither `Openat` nor `SYS_OPENAT`, so there is no stdlib route to the `openat(2)` walk `openBeneath` needs. That single symbol fails the whole package, so every macOS build of the pack is broken — including the documented `cd adapter && go build -o gc-slack-adapter`.

Fixed by switching `openBeneath` to `golang.org/x/sys/unix`, which exposes `Openat` on both platforms with identical semantics.

**`os.Root` (Go 1.24+) was considered and rejected.** It confines correctly, but it *silently ignores* `O_NOFOLLOW` — it would follow a symlink planted inside the file store and quietly relax the exact guarantee `openBeneath` exists to provide. The file's original comment avoided `x/sys` to keep `go.mod` dependency-free; that isn't worth a weakened security primitive.

Semantics verified unchanged on darwin/arm64 — an in-root symlink at any component still fails `ELOOP` under `O_NOFOLLOW` (`TestOpenBeneathRefusesSymlinkComponents`, `TestReadConfinedFileRejectsSymlink`).

## 2. Three tests that could never run on macOS

With the package compiling for the first time on darwin:

```
--- FAIL: TestReadConfinedFileStillReadsAndStillConfines
--- FAIL: TestReadConfinedFileReadsRealFile
--- FAIL: TestReadConfinedFileRejectsSymlink
    readConfinedFile: path "/var/folders/.../real.txt"
    is outside root "/private/var/folders/.../001"
```

This is a **test-harness artifact, not a confinement defect.** `confineFileUploadPath` `EvalSymlinks`-resolves the *root* but only `Abs`+`Clean`s the *path*, because its documented contract is that callers pass an already-resolved path — which the production call site honours. The tests passed a raw `t.TempDir()`.

On Linux that is invisible: `TMPDIR` has no symlinked component, so both forms are identical. On macOS `TMPDIR` lives under `/var`, which **is** a symlink to `private/var` — so the root canonicalized to `/private/var/...` while the path stayed `/var/...`, `filepath.Rel` returned a `../..` escape, and the confinement check correctly rejected its own fixture.

Added a `tempDirResolved` helper that resolves the temp dir (matching what production guarantees) and pointed those three tests at it. **No production behaviour changes** — the only non-test edits are `confined_open.go`, `go.mod`, `go.sum`.

## Verification

On darwin/arm64, go1.26.2:

- `go build ./...` — clean (fails on `main` with the error above)
- `go vet ./...` — clean
- `go test ./...` — **ok**, full suite

Happy to split the test fix into its own commit if you'd prefer to review them separately.

### DIFF
--- a/slack-full/CHANGELOG.md
+++ b/slack-full/CHANGELOG.md
@@ -48,6 +48,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- The adapter builds and its tests pass on darwin again. The confined-open
+  walk called `syscall.Openat`, which exists only on Linux — darwin's
+  libSystem-based `syscall` package exports neither `Openat` nor
+  `SYS_OPENAT` — so the entire adapter failed to compile on macOS. The
+  walk now binds `golang.org/x/sys/unix`; every flag constant and syscall
+  it swaps is a value-identical mirror on Linux, so behavior there is
+  unchanged. Three `readConfinedFile` tests also failed on macOS, where
+  `TMPDIR` lives under `/var`, itself a symlink to `/private/var`:
+  `confineFileUploadPath` EvalSymlinks-resolves only the *root* and
+  documents the path as caller-pre-resolved, so a fixture handing it a
+  raw `t.TempDir()` broke that contract and made every path — including
+  genuinely in-root ones — read as an escape. The fixtures now resolve
+  their temp dir once through `tempDirResolved`. A harness defect, not a
+  confinement one: no production behavior changed.
 - Inbound files recorded inside Slack itself — voice clips (file subtype
   `slack_audio`) and video clips (`slack_video`) — arrive with an empty
   `mimetype` AND `filetype`, which the adapter passed through verbatim.
@@ -100,6 +114,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- The adapter module takes its first dependency, `golang.org/x/sys`
+  (hash-pinned in `adapter/go.sum`), because it is the only maintained
+  source of a darwin-capable `Openat`. The module's zero-dependency
+  property was traded away deliberately; the rejected alternatives and
+  the reason are recorded in `adapter/confined_open.go`. One
+  operator-visible consequence: `adapter/run.sh`'s build-on-missing
+  self-heal used to rebuild the binary with no module downloads at all,
+  and now needs either a warm module cache or GOPROXY egress the first
+  time it builds. On a host behind an egress allowlist, pre-warm with
+  `go mod download` in `slack-full/adapter/` before relying on the
+  self-heal — a cold cache there fails the rebuild with `module lookup
+  disabled by GOPROXY=off`, and the remedy `run.sh` prints on failure
+  hits the same wall.
 - **Behavior change on upgrade:** the slack service now *exits 1 at
   startup* when `GC_SLACK_ADAPTER_ENV` is set to a path that does not
   exist, where it previously logged one warning and started on the
--- a/slack-full/adapter/confined_open.go
+++ b/slack-full/adapter/confined_open.go
@@ -5,7 +5,8 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
-	"syscall"
+
+	"golang.org/x/sys/unix"
 )
 
 // openBeneath opens rel (a Clean, root-relative path containing no
@@ -14,8 +15,20 @@ import (
 // with O_NOFOLLOW set. The parent fd pins the verified directory inode,
 // so a directory swapped for a symlink mid-walk cannot redirect the
 // traversal — the userspace equivalent of openat2(RESOLVE_BENEATH),
-// which stdlib syscall does not expose (and the adapter's
-// zero-dependency go.mod keeps us from importing x/sys for).
+// which stdlib syscall does not expose.
+//
+// Uses golang.org/x/sys/unix rather than stdlib syscall: `syscall.Openat`
+// exists only on Linux, so the stdlib version failed to COMPILE on darwin
+// ("undefined: syscall.Openat"), taking the whole adapter with it. Neither
+// stdlib escape hatch works here — darwin's syscall package is libSystem-
+// based and does not export SYS_OPENAT either, and os.Root (Go 1.24+)
+// confines correctly but silently IGNORES O_NOFOLLOW, so it would follow a
+// symlink planted inside the file store and quietly relax the guarantee
+// this function exists to provide. x/sys/unix keeps the semantics exact
+// (verified on darwin/arm64: an in-root symlink still fails ELOOP under
+// O_NOFOLLOW) at the cost of the one dependency this file's original
+// comment was avoiding. That trade is deliberate: correctness of a security
+// primitive over a dependency count.
 //
 // O_NOFOLLOW applies to every component, not just the leaf, so any
 // symlink anywhere beneath root is a hard failure (ELOOP / ENOTDIR).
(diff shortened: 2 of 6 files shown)


jeffora:
Independent corroboration of this PR's central claim, from a separate effort that reached the same conclusion before finding this PR. I've closed my duplicate (#295) in favour of this one, which is strictly broader — it also fixes the `/var` test failures and adds `TestOpenBeneathRefusesSymlinkComponents`.

Posting the measurement because this PR **asserts** the `os.Root` point and it's worth having demonstrated. A future reviewer looking to drop the `x/sys` dependency will find the existing tests green and reasonably conclude `os.Root` is equivalent — it isn't, and the test suite structurally cannot show that.

Measured on darwin/arm64, Go 1.26.1:

| fixture | `os.Root` | `O_NOFOLLOW` walk |
|---|---|---|
| symlink, **absolute** target | rejected — "path escapes from parent" | rejected — ELOOP |
| symlink, **relative** in-root target | **followed, file contents returned** | rejected — ELOOP |
| control, real file | opened | opened |

**Why the current suite can't catch it:** `TestReadConfinedFileRejectsSymlink` builds its fixture as

```go
target := filepath.Join(dir, "target.txt")
os.Symlink(target, link)   // absolute target
```

`os.Root` rejects that for *escaping the root*, not for *being a symlink* — so both implementations reject it and the difference is invisible. Change that one line to `os.Symlink("target.txt", link)` and an `os.Root` implementation opens the file and returns its contents, while the `O_NOFOLLOW` walk still rejects it. That's the case worth pinning, and `TestOpenBeneathRefusesSymlinkComponents` in this PR looks like the right home for it if it doesn't already cover a relative in-root target.

Also confirming the dead end, since it's the obvious next idea: `root.OpenFile(rel, os.O_RDONLY|syscall.O_NOFOLLOW, 0)` does **not** recover the strict behaviour — `os.Root` doesn't honour the flag, and the relative in-root symlink is still followed. `lstat`-before-open isn't equivalent either; it reintroduces the check-then-open window this function exists to close.

One note on the go.mod trade, in case it comes up in review: the module had no `require` block and no `go.sum` at all, so this genuinely introduces both. That's the strongest argument for `os.Root` and it deserves to be weighed openly — but the guarantee in `openBeneath`'s doc comment ("any symlink anywhere beneath root is a hard failure") is contingent on `Openat`, and `os.Root` cannot express it. Trading a deliberate TOCTOU defence for an empty `go.mod` looks like the wrong direction.

Separately, on part 2: an issue was filed for the `/var` behaviour (#296) before this PR was found, framing it as a production bug on the grounds that `INBOUND_FILE_STORE` defaults under `/tmp`. This PR argues it's a test-harness artifact because the production call site passes an already-resolved path. Those conflict; we're checking which is right and will correct #296 rather than leave two contradictory records.

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277293230

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277293372

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277294091

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277310331

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @outdoorsea! This PR restores the slack-full adapter on macOS: the confined-open walk called `syscall.Openat`, which exists only on Linux, so the entire adapter failed to compile on darwin — and three `readConfinedFile` tests failed there for a second, independent reason (`TMPDIR` lives under `/var`, itself a symlink to `/private/var`). Fixing both is what makes macOS a workable development platform for this pack again, and the diagnosis of the second half — that `confineFileUploadPath` EvalSymlinks-resolves only the *root* and documents the path as caller-pre-resolved, so a fixture handing it a raw `t.TempDir()` breaks that contract — is the part that took real care to get right.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- **The darwin build break was re-measured rather than taken on faith.** At the base commit `GOOS=darwin GOARCH=arm64 go build ./...` fails with exactly `./confined_open.go:46:24: undefined: syscall.Openat`; at your head it succeeds. Linux builds cleanly on both sides, `go vet` is clean, and the binding swap is value-identical there (`unix.O_*` mirrors `syscall.O_*`, and the walk logic is untouched), so Linux behavior does not move.
- **The three macOS test failures were reproduced on Linux**, by routing `TMPDIR` through a symlink to recreate the darwin shape. At base exactly `TestReadConfinedFileStillReadsAndStillConfines`, `TestReadConfinedFileReadsRealFile` and `TestReadConfinedFileRejectsSymlink` fail with the reported `path "…" is outside root "…"` signature; at your head all seven confined tests pass. The review agreed the fix sits at the right layer: `tempDirResolved` makes the fixtures meet the contract production already honors, and no production behavior changes.
- **The new dependency was checked, not just accepted.** `go.sum` carries both hashes for `golang.org/x/sys v0.47.0`, and the full suite, `go vet`, and `go test -race -count=1 ./...` all pass with `GOPROXY=off` against a warm module cache.
- **Completeness was verified package-wide.** The four remaining stdlib `syscall` call sites (`ingress_receipts.go:985`, `:1104`, `:1172`, `company_bot_tokens.go:66`) use only darwin-available symbols, so `Openat` really was the sole non-portable one.

One reviewer concern was checked and then dropped rather than acted on: a lane suggested the root-resolution contract had no Linux CI signal and proposed adding a symlinked-root test. That coverage already exists — `TestConfineFileUploadPath` (`main_test.go:5114`) builds an explicit `linked-root` symlink and asserts both directions, and deleting the root `EvalSymlinks` at `main.go:2004-2006` turns it red on an ordinary `TMPDIR`. The proposed test would have duplicated an existing pin, so nothing was added to your PR for it.

Review findings addressed:
- **No CHANGELOG entry for a user-visible change** (Release Safety, minor; raised independently by two lanes). `CONTRIBUTING.md` requires an `[Unreleased]` bullet, and both the restored macOS build and the module's first dependency are operator-visible. Fixed in maintainer commit `6a0f988`, which adds a `### Fixed` bullet for the darwin build/test repair and a `### Changed` bullet for the `golang.org/x/sys` dependency.
- **The new dependency quietly changed `run.sh`'s offline behavior** (Change Impact, minor). `adapter/run.sh`'s build-on-missing self-heal used to rebuild with no module downloads at all; with a cold module cache and no egress it now fails with `module lookup disabled by GOPROXY=off`, and the remedy the script prints on failure hits the same wall. The review's recommendation was to document rather than vendor, so that same `### Changed` bullet in `6a0f988` records the new first-build requirement and points operators at `go mod download` to pre-warm.

- `mc-r3lev`: no darwin guard in CI — the adapter leg runs Linux-only, so nothing type-checks a non-Linux GOOS; a `GOOS=darwin GOARCH=arm64 go build ./...` step would have caught the original break (Test Evidence Quality, minor).
- `mc-u27st`: `main_test.go:2773-2774` still hands `readConfinedFile` a raw `t.TempDir()`, so on darwin that one negative test cannot distinguish in-root from out-of-root (Test Evidence Quality, minor).
- `mc-03e12`: `cache-dependency-path` in `ci.yml` does not list the adapter's new `go.sum`, so the module cache key never rotates on adapter dependency changes (Release Safety, minor).
- `mc-vb6fa`: the package now carries two bindings for the same `O_NOFOLLOW` open idiom — `unix.Open` here, stdlib `syscall.Open` at four sibling sites (Architectural Consistency, nit). Deliberately left out of this PR, since fixing it would widen a portability fix into a package-wide refactor.

### Maintainer Changes
One commit, docs only: `docs(slack-full): record the darwin build fix in the CHANGELOG` — `slack-full/CHANGELOG.md`, 1 file changed, 27 insertions(+). Your commit is preserved unchanged; the branch was also rebased onto current `main` (`128002e`) so the merge queue sees an up-to-date base.

### Final Review Status
Ready for the merge queue: final head `6a0f988` has passing required GitHub checks, and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/gascity-packs/actions/runs/34226805908

### Review Iterations
One review pass across three independent lanes, all three present, approved on that pass with no blocking findings — followed by one maintainer docs fixup and a rebase onto current `main`, each re-validated on the final head (darwin cross-build, `go vet`, and the full adapter suite green).

---

PR #273, opened 2026-08-05
fix(gastown): fail the polecat push gate closed when a no-push rail has no auto_push

edwarddavison:
## The bug

`mol-polecat-work`'s `submit-and-exit` read the no-push opt-out as:

```bash
AUTO_PUSH=$(gc bd show "$WORK_BEAD_ID" --json | jq -r '.[0].metadata | if has("auto_push") then (.auto_push | tostring) else "" end')
if [ "$AUTO_PUSH" = "false" ]; then <halt> fi
git push origin HEAD
```

An **absent** `auto_push` key yields the empty string, which is not `"false"`, so control falls straight through to the push and the refinery handoff. A bead whose DESCRIPTION says *"HALT branch-ready, the mayor publishes, no push"* but which carries no `auto_push` key gets **pushed** by a polecat following the gate exactly as written.

A rail expressed only in prose is enforced by nothing.

A sweep of one live city found **21 open beads in that state across four rigs**. Fifteen had halted anyway — the polecats read the prose and complied with intent over mechanism. That is good behaviour, not enforcement; it is one differently-prompted agent away from an unintended push. Two of the four rigs are customer repos, and an unintended push to a customer repo has already caused an unintended prod deploy once.

## The fix

A push gate should fail **closed**. Absent metadata is not consent.

| `auto_push` | Bead prose | Outcome |
|---|---|---|
| `false` | not read | halt at branch-ready — unchanged |
| absent | asserts a no-push rail | **halt at branch-ready and escalate** (`halt_reason=no_push_rail_unresolved`) |
| absent | no rail | push + refinery handoff — unchanged |
| any other value | not read | push + refinery handoff — explicit consent |

The escalation mail tells the rail's owner exactly how to resolve it (`--set-metadata auto_push=false|true`, then re-sling), so the halt is a round trip, not a dead end.

### Why the detection is deliberately narrow

- **DESCRIPTION + NOTES only.** Comments are discussion, not standing instructions, and they self-match on any bead that merely *talks about* no-push rails.
- **The push-_to-main_ phrase is stripped before matching** (the phrase, not the whole line). "Do not push to main" is a different rail — every polecat already obeys it, since it pushes a feature branch and never main — and half the beads in a city quote it. Matching on it would make the gate cry wolf, and a gate that cries wolf gets routed around. Stripping only the phrase keeps any *other* rail asserted on the same line visible to the scan.
- **An empty scan result halts too** (`${RAIL_HIT:-1}`): `grep -c` always prints a count when it runs, so an empty result means the scan itself broke, and an undetermined rail is treated as an asserted one.

### Closing the gap at the authoring end

The gate is a backstop, not a substitute. It costs a round trip and a human read every time it fires; it can only see DESCRIPTION and NOTES, so a rail that lives in a comment is invisible to it; and **it covers only the polecat's own `submit-and-exit`**. A polecat that dies mid-work is recovered by the witness's orphan salvage (`mol-witness-patrol.toml:301,313`), which pushes the branch without consulting `auto_push` or the prose at all — that path is untouched by this PR and is filed as its own follow-up. So on a bead that must not reach the remote, the metadata is the record that survives the crash. The mayor's bead-filing gotchas now require a prose rail to be recorded as metadata **in the same write**, and warn that the `.auto_push // "-"` jq idiom reads the legitimate value `false` as absent.

## Verification

Run against a live city, not just fixtures:

- All six beads where the gate could still fire (`det-aur`, `fph-aijq`, `fph-7grj`, `sb-fmly`, `sb-n2qj`, `gci-omd`) now **halt**.
- A bug report reading *"mirror-sync commits locally and never pushes"* does **not** halt. That was a real false positive during development (`never push` matched `never pushes`); it is now a test case.
- Sweep of 142 open beads across nine rigs: hits land on the no-push cohort, not on ordinary work.

`gastown/tests/test_polecat_push_gate.sh` (new, CI auto-runs `gastown/tests/test_*.sh`):

- pins the **ordering** — the ambiguity halt must appear before `git push origin HEAD`;
- **lifts the shipped regexes out of the formula** so the test cannot drift from the gate it tests;
- covers both directions of the rail / no-rail split, including the standing "do not push to main" phrasing.

Mutation-tested: removing the ambiguity halt fails the test, and loosening `never push([^e]|$)` back to `never push` fails it on the false-positive case.

Local gates green: `tomllib` parse, `bash -n` on the rendered gate, `validate_registry.py --require-git`, all five `gastown/tests/test_*.sh`, `shellcheck --severity=style` on the new test, and `gc lint gastown` unchanged at 35 pre-existing findings (0 new).

### DIFF
--- a/gastown/agents/mayor/prompt.template.md
+++ b/gastown/agents/mayor/prompt.template.md
@@ -141,6 +141,26 @@ coordination with no owning repository belongs at city scope.
 
 **Rule**: Think "X needs Y", not "X comes before Y". Verify with `gc bd blocked`.
 
+**A rail expressed only in prose is enforced by nothing.** If you write "no
+push", "HALT branch-ready", or "the mayor publishes" into a bead, record the
+decision as metadata in the SAME write. Prose is read by an agent exercising
+judgement; the push gate is a shell test on `metadata.auto_push`.
+
+- WRONG: description says "branch + HALT branch-ready, mayor publishes" — and nothing else
+- RIGHT: that description, plus `gc bd update <id> --set-metadata auto_push=false`
+
+`mol-polecat-work` fails closed on the mismatch — a bead whose prose asserts a
+no-push rail with no `auto_push` key halts at branch-ready and escalates rather
+than pushing. That is a backstop, not a substitute: it costs a round trip and a
+human read every time, and it can only see DESCRIPTION and NOTES, so a rail
+that lives in a comment is invisible to it.
+
+**Rule**: if a rail changes what the polecat DOES, it belongs in metadata. Use
+`--set-metadata` (never bare `--metadata`) so `branch` and `gc.routed_to`
+survive the write. Reading it back, prefer
+`if has("auto_push") then .auto_push else "absent" end` over `.auto_push // "-"`
+— the `//` idiom treats the legitimate value `false` as absent.
+
 ## Responsibilities
 
 - **Work dispatch**: Assign work to polecats for issues, coordinate batch work on epics
--- a/gastown/agents/mayor/prompt.template.md
+++ b/gastown/agents/mayor/prompt.template.md
@@ -149,17 +149,63 @@ judgement; the push gate is a shell test on `metadata.auto_push`.
 - WRONG: description says "branch + HALT branch-ready, mayor publishes" — and nothing else
 - RIGHT: that description, plus `gc bd update <id> --set-metadata auto_push=false`
 
+The vocabulary is closed and case-folded: `false`/`no`/`0` halt, `true`/`yes`/`1`
+push. Anything else is not read as consent — it halts and escalates, the same as
+metadata the gate cannot decode. A `null` value is the exception, because it is
+JSON's own spelling of "nothing recorded": it is treated as an ABSENT key, not as
+a halt, so it falls through to the prose scan and a rail-less bead still pushes.
+Record one of the six words; do not record `null` expecting a halt.
+
 `mol-polecat-work` fails closed on the mismatch — a bead whose prose asserts a
 no-push rail with no `auto_push` key halts at branch-ready and escalates rather
 than pushing. That is a backstop, not a substitute: it costs a round trip and a
-human read every time, and it can only see DESCRIPTION and NOTES, so a rail
-that lives in a comment is invisible to it.
+human read every time, it can only see DESCRIPTION and NOTES, so a rail that
+lives in a comment is invisible to it, and it only covers the polecat's own
+submit-and-exit. A polecat that DIES mid-work is recovered by the witness's
+orphan salvage, which pushes the branch without consulting `auto_push` or the
+prose at all — so on a bead that must not reach the remote, the metadata is the
+record that survives the crash, and even it is not enforced on that path.
 
 **Rule**: if a rail changes what the polecat DOES, it belongs in metadata. Use
 `--set-metadata` (never bare `--metadata`) so `branch` and `gc.routed_to`
-survive the write. Reading it back, prefer
-`if has("auto_push") then .auto_push else "absent" end` over `.auto_push // "-"`
-— the `//` idiom treats the legitimate value `false` as absent.
+survive the write. Read it back with the shape NORMALISED before the key test:
+
+```bash
+WORK_JSON=$(gc bd show <id> --json)
+if [ -z "$WORK_JSON" ]; then
+  echo "unreadable"   # the ledger read failed; retry. NOT the same as "absent".
+else
+  printf '%s' "$WORK_JSON" | jq -r '
+    if (type != "array") or (length == 0) or ((.[0] | type) != "object")
+    then "unreadable"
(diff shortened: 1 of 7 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277292839

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277292908

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277310084

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @edwarddavison! This PR closes a real fail-open in `mol-polecat-work`'s `submit-and-exit`: an **absent** `auto_push` key was not `"false"`, so control fell straight through to the push, and a bead whose description said *"HALT branch-ready, the mayor publishes, no push"* got pushed by a polecat following the gate exactly as written. Your sweep found 21 open beads in that state across four rigs — two of them customer repos, where an unintended push has already caused an unintended prod deploy once. Turning that into a gate that fails **closed**, with an escalation that tells the rail's owner how to resolve it so the halt is a round trip rather than a dead end, is the right fix. The 858-line test suite that **lifts the shipped regexes out of the formula** so the test cannot drift from the gate it tests is the strongest part of the change.

### Original PR Review

Specific gaps fixed:

- **The "push to main" exclusion was over-broad and line-granular.** The original `grep -Eiv` dropped the *whole line* whenever it mentioned pushing to main, and matched a bare `origin`. A line that asserted a second, real rail alongside the standing "do not push to main" boilerplate therefore lost that rail entirely — seven measured wordings fell through to a push. The final patch replaces the line drop with a destination-narrowed **phrase** strip, so anything else asserted on the same line stays visible to the scan. All seven wordings now halt, and all fifteen original `test_rail_detection` rows keep their verdicts.
- **The suite could abort silently.** The regex lift was anchored to a line pattern under `set -e`: when the pattern moved, the assignment failed, the suite exited early, and the `|| fail` guards below it never ran — leaving a bare rc=0 that proved nothing. The lift is now `tomllib`-anchored to the single bash fence containing `git push origin HEAD` inside `submit-and-exit`, requires exactly one match, and is non-fatal so the guards stay reachable. Deleting the strip stage now produces a named failure instead of zero bytes of output.

Review findings addressed:

- Both gating findings above are fixed in `b1d360e`, and each was validated with a mutation control rather than by inspection: removing the strip stage and inverting the rail-hit halt each turn the suite red with a named message.
- `b1d360e` also carries the reviewers' non-gating suggestions: the prose alternation now includes `(must|should) not push` and the full closed value vocabulary `auto_push=(false|no|0)`, so the prose channel accepts the same words the metadata channel does; the strip's protected boundary class now includes `/`, so a rail naming a nested branch keeps its verb; `test_gate_structure` is scoped to the submit-and-exit block and asserts a mail send **per fail-closed halt** instead of a whole-file total, which removes a future false-red; and the mayor's read-back recipe gained the probe's no-record/empty guards plus an explicit `null` → absent mapping, so a failed ledger read is no longer indistinguishable from "no key".

- `mc-0yf6q`: the witness's orphan salvage (`mol-witness-patrol.toml:301,313`) still pushes a salvaged branch without consulting `auto_push` or the prose, so a polecat that *dies* mid-work is recovered by a path this gate does not cover. It is pre-existing and identical at this PR's merge base, and was deliberately kept out of this diff rather than expanding the change into an untouched formula.
- `mc-5zivo`: the bare-`--notes` regression assertion is scoped to `mol-polecat-work`'s `submit-and-exit` only. The class is closed across gastown today (verified zero executable occurrences), but nothing yet keeps it closed against drift in other formulas.

One maintainer edit to the PR description, for accuracy: its "why the detection is deliberately narrow" bullet still described the old line-dropping behaviour, and the "backstop, not a substitute" paragraph claimed broader enforcement than the gate actually has. Both now describe the shipped phrase strip and scope the claim to the polecat's own `submit-and-exit`, matching `mayor/prompt.template.md:160-164`. Everything else in the description is yours, unchanged.

### Maintainer Changes
`b1d360e` — fix(gastown): strip the push-to-main phrase, and make the rail scan fail closed (7 files changed, 637 insertions(+), 77 deletions(-)).

Your three commits are preserved with authorship intact. The branch was rebased onto current `main`; the resulting patch is byte-identical by patch-id to the reviewed one.

### Final Review Status
Ready for the merge queue: final head `b1d360e` has passing required GitHub checks, and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/gascity-packs/actions/runs/34297514596

### Review Iterations
2 review passes. Pass 1 returned request_changes with the two gating findings above; the fix commit landed, and pass 2 approved with all three reviewer lanes plus the synthesizer independently re-running the five `gastown/tests/test_*.sh` suites and two mutation controls in scrubbed scratch worktrees.

---

PR #277, opened 2026-08-06
decompose: stamp gc.routed_to on work-units so they are claimable

csauer02-personal-user:
## decompose: stamp gc.routed_to on decomposed work-units so they are claimable

Decomposed implementation-convoy work-units are created via ad-hoc `bd create` in the decompose step, bypassing the formula-graph materializer's routing stamp. They land `bd ready` with `gc.routed_to` empty — and pool demand/claim keys exclusively on `gc.routed_to` — so they are structurally invisible to every pool until a human `gc sling` stamps them. That manual sling at every stage transition is the drag this fixes.

**Change:** a coded post-decompose gate (`route-workunits.sh`, run as a `steps.check`) stamps `gc.routed_to` (+ `gc.execution_routed_to`) = the **declared** `implementation_target` var onto each convoy work-unit, resolved via `gc.build.implementation_convoy_id` → `gc convoy status` members. The target is a copied declared value (no judgment-in-Go, no role hardcode); scoped to explicit convoy membership (no over-reach); idempotent (skips already-routed members); fail-loud if the target/convoy are undeclared or any member stays routeless (no silent failure). The decompose→route-workunits→implement `needs` chain guarantees the stamp completes before the drain's claim-scan.

**Wiring completeness:** every concrete build formula that `extends build-base` re-declares its own `implement`/`implement-same-session` steps (composition is per-step-id replacement), so each was repointed to `needs = ["route-workunits"]` (build-basic, compound-build, gstack-build, superpowers-build; bmad threads it via implementation-readiness). The `build-from-*` family gates via `prepare-convoy`. Zero orphans.

**Size:** ~477 production (formula TOML + prompt/shell assets, zero Go) + test. **Verification:** a `needs`-chain test asserts the transitive chain per concrete formula (non-vacuous: fails on the pre-fix wiring). Pack asset suite 0 net-new failures. Two independent adversarial reviews: both green (the wiring-completeness gap was caught + closed in review).

Note (transparency): the `CONCRETE_BUILD_FORMULAS` needs-chain guard covers the 5 `build-base` extenders, not the `build-from-*` family (benign — those don't re-declare `implement`; optional belt-and-suspenders).

### DIFF
--- a/bmad/formulas/bmad-build.formula.toml
+++ b/bmad/formulas/bmad-build.formula.toml
@@ -105,7 +105,7 @@ timeout = "5m"
 [[steps]]
 id = "implementation-readiness"
 title = "BMAD implementation readiness"
-needs = ["decompose"]
+needs = ["route-workunits"]
 metadata = { "gc.run_target" = "bmad.readiness-reviewer" }
 description_file = "../assets/workflows/bmad-build/implementation-readiness.md"
 
--- a/compound-engineering/formulas/compound-build.formula.toml
+++ b/compound-engineering/formulas/compound-build.formula.toml
@@ -91,7 +91,7 @@ description_file = "../assets/workflows/compound-build/plan-review.md"
 [[steps]]
 id = "implement"
 title = "Drain Compound Engineering work"
-needs = ["decompose"]
+needs = ["route-workunits"]
 condition = "{{drain_policy}} == separate"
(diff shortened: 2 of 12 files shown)


julianknutsen:
The drain step and implementation is responsible for handling the routing/execution of the work beads. If there is a bug there, let's fix it instead of pre-routing work.

Issue #252, opened 2026-07-29
Wisp reconciliation queries omit --include-infra, leaking one wisp per restart for every patrol role


The1nk:
Wisp reconciliation queries in the `gastown` pack omit `--include-infra`, so they can never find an existing wisp. Every patrol agent concludes "no wisp exists" and pours a fresh one, leaking one wisp per restart.

Observed against pack commit f69ec02.

## Root cause

Wisp/molecule roots are **infra** beads. `gc bd list` excludes infra beads unless `--include-infra` is passed. The reconciliation queries filter on `--type=molecule` but omit that flag, so they return an empty set regardless of what is actually open.

The queries themselves are otherwise correct — this is purely a visibility problem.

## Reproduction

Live rig, bead `ta-wisp-b4e` open and assigned to `<rig>/gastown.witness`:

```console
$ gc bd list --assignee=<rig>/gastown.witness --status=open --type=molecule --limit=0 --json | jq length
0

$ gc bd list --assignee=<rig>/gastown.witness --status=open --type=molecule --limit=0 --include-infra --json | jq -r '.[].id'
ta-wisp-b4e
```

## Impact, observed twice in one morning

**witness** — startup reconciliation ran the documented query, got an empty result, and poured `ta-wisp-jtw` on top of the already-open `ta-wisp-b4e`. Two live patrol wisps for one witness.

**refinery** — independently reproduced 20 minutes later while working a normal bead: `ta-wisp-fr7` at 03:24:47Z, then `ta-wisp-58z` at 03:26:39Z. Two open patrol wisps, same cause.

Both agents self-corrected only because a human noticed. The recipe re-leaks on every restart.

## Affected call sites

16 sites across 6 files — the pattern is inherited by every patrol role, not just the witness:

| File | Lines |
|---|---|
| `gastown/formulas/mol-witness-patrol.toml` | 534 |
| `gastown/formulas/mol-refinery-patrol.toml` | 122, 245, 346, 438, 1146, and prose at 1109 |
| `gastown/formulas/mol-deacon-patrol.toml` | 463 |
| `gastown/agents/witness/prompt.template.md` | 169, 170, 206, 213 |
| `gastown/agents/refinery/prompt.template.md` | 72, 117 |
| `gastown/agents/deacon/prompt.template.md` | 92, 94 |

Worth noting `mol-witness-patrol.toml:162` already passes `--include-infra` for session beads, so the flag is clearly known — the molecule queries just missed it.

## Fix

[...]

PR #278, opened 2026-08-06
fix(gastown): add --include-infra to wisp reconciliation queries in patrol formulas (#252)

edwarddavison:
Closes the formula-side half of #252.

**Rebased 2026-09-11 onto `2716505`.** This PR previously conflicted, because
#325 fixed the witness half independently while it sat. Rather than resolve the
conflict, the witness hunks have been **dropped**: see *What changed in the
rebase* below.

## Problem

Wisp roots are **ephemeral infrastructure beads**. `gc bd list` hides infra beads
unless `--include-infra` is passed, so a `--type=molecule` reconciliation query
without the flag returns `[]` regardless of how many wisps are actually open.
Every patrol iteration concludes "no wisp exists", pours a fresh one, and orphans
the wisp it was standing on — **one leaked wisp per iteration, per patrol role,
town-wide.**

## Reproduction (live rig)

The rig had 4 open wisps — 2 on the witness, 2 on the refinery — themselves the
product of this duplicate-pour loop, and all 4 invisible to the query the
formulas actually run:

```console
$ gc bd list --assignee=gas-city/gastown.witness --status=open --type=molecule --limit=0 --json | jq length
0

$ gc bd list --assignee=gas-city/gastown.witness --status=open --type=molecule --include-infra --limit=0 --json | jq length
2
```

Confirmed the hidden beads are exactly the wisps (`ephemeral=true`,
`issue_type=molecule`), and that **the flag alone is sufficient** — with
`--include-infra`, `gc bd list --type=molecule` returns the *same id set* as
`gc bd query 'ephemeral=true'`:

```console
$ gc bd query --json 'ephemeral=true' | jq -r '[.[]|select(.status=="open")]|map(.id)|sort|join(" ")'
gci-wisp-5an gci-wisp-8kq gci-wisp-bvr gci-wisp-d2g

$ gc bd list --status=open --type=molecule --include-infra --limit=0 --json | jq -r 'map(.id)|sort|join(" ")'
gci-wisp-5an gci-wisp-8kq gci-wisp-bvr gci-wisp-d2g
```

## What changed in the rebase

**The witness hunks are gone.** `8243d049` (#325, merged 2026-09-08) added
`--include-infra` to `mol-witness-patrol.toml` and to the witness prompt while
this PR was open. Re-applying either would have been a no-op at best and a
conflict at worst, so both were dropped. `mol-witness-patrol.toml` and
`gastown/agents/witness/prompt.template.md` are **untouched by this PR** — that
file is upstream's now.

**The refinery site count moved from 5 to 6.** The original PR was written
against an older main. The sites were re-measured against `2716505` rather than
re-applied by line number; a sixth `CURRENT_WISP` call site had appeared in
`mol-refinery-patrol.toml` in the interim.

**The regression test is a re-add, not a revert.** #325 rewrote
`test_gastown_pack_assets.sh`, including adding its own witness-scoped
`--include-infra` guard. This PR adds the refinery and deacon equivalents on top
of that version, in the same idiom, rather than restoring the original PR's
pack-wide assertion — which would now fail on `#266`'s unlanded prompt work.

## Changes

| File | Sites |
|---|---|
| `gastown/formulas/mol-refinery-patrol.toml` | 6 `CURRENT_WISP` reconciliation sites + 1 prose reference |
| `gastown/formulas/mol-deacon-patrol.toml` | 1 `CURRENT_WISP` reconciliation site |
| `gastown/tests/test_gastown_pack_assets.sh` | 2 new guards + 2 scope-note corrections |

The prose reference in the refinery's `patrol-summary` step is fixed as well as
the code. It is the line that tells the next session how to find its
predecessor, so a bare form there teaches the blind query onward.

The deacon formula carries the identical defect and **no PR has ever covered
it** — #266 took the prompts, #325 took the witness formula. Happy to split that
hunk out if you would rather keep this scoped to the refinery.

## Deliberately out of scope: the prompt templates

`gastown/agents/refinery/prompt.template.md` (2 sites) and
`gastown/agents/deacon/prompt.template.md` (2 sites) still carry the bare form.
Those are the prompt half of #252 and belong to **#266**, which is still open.
Fixing them here would collide with that PR, so the two new guards are
**formula-scoped**, and the existing witness and boot guards' scope notes have
been corrected — they currently claim the refinery and deacon *formulas* are the
outstanding gap, which this PR closes. Pack-wide census after this PR: 18/22
flagged, the 4 remaining being exactly #266's.

## Test

`test_refinery_wisp_queries_pin_include_infra` and
`test_deacon_wisp_queries_pin_include_infra` assert that every
`gc bd list --type=molecule` line in the respective formula carries the flag.

Verified as real regression tests — both arms, on the pre-fix tree:

```console
# main's formulas verbatim, only the new test file dropped in:
FAIL: refinery --type=molecule wisp queries must pass --include-infra (0/7 do)

# then with ONLY the refinery formula fixed, so the deacon guard is reached:
FAIL: deacon --type=molecule wisp queries must pass --include-infra (0/1 do)
```

Both pass after the fix. All **five** `gastown/tests/test_*.sh` suites pass under
`ci.yml`'s exact loop; both edited formulas parse under `tomllib`.

## Overlap with #264 — please note merge order

#264 (also mine) rewrites five of these same `CURRENT_WISP` sites in
`mol-refinery-patrol.toml` to resolve wisps via `gc bd query`. **Whichever lands
second needs a trivial rebase** — the two approaches are semantically equivalent
(id sets proven identical above), and #264's rewrite supersedes these lines
entirely if it lands. Flagging rather than resolving it here, since that is the
maintainer's call.

One correction worth flagging: #264's regression test asserts *"`gc bd list`
never returns ephemeral beads regardless of `--type`"*. That premise is wrong —
as shown above, `--include-infra` makes `gc bd list` return them exactly. #264's
fix still works, but its stated justification does not, so that assertion is
deliberately not adopted here.

### DIFF
--- a/gastown/formulas/mol-deacon-patrol.toml
+++ b/gastown/formulas/mol-deacon-patrol.toml
@@ -542,7 +542,10 @@ Handle any urgent messages that arrived during patrol. Archive the rest.
 # or the deacon gains a second slot.
 CURRENT_WISP=${GC_BEAD_ID:-${GC_TRIGGER_WORK_BEAD_ID:-}}
 if [ -z "$CURRENT_WISP" ]; then
-  CURRENT_WISP=$(gc bd list --assignee="$GC_AGENT" --status=in_progress --type=molecule --limit=1 --json | jq -r '.[0].id // empty')
+  # --include-infra is required: wisp roots are ephemeral infrastructure
+  # beads, which gc bd list hides by default. Without it this resolves empty
+  # and the deacon pours a duplicate wisp instead of burning this one.
+  CURRENT_WISP=$(gc bd list --assignee="$GC_AGENT" --status=in_progress --type=molecule --include-infra --limit=1 --json | jq -r '.[0].id // empty')
 fi
 
 NEXT=$(gc bd mol wisp mol-deacon-patrol --root-only --var binding_prefix='{{binding_prefix}}' --json | jq -r '.new_epic_id // empty')
--- a/gastown/formulas/mol-refinery-patrol.toml
+++ b/gastown/formulas/mol-refinery-patrol.toml
@@ -119,7 +119,7 @@ wisp, burn the current wisp, then request a restart:
 ```bash
 CURRENT_WISP=${GC_BEAD_ID:-}
 if [ -z "$CURRENT_WISP" ]; then
-  CURRENT_WISP=$(gc bd list --assignee="$GC_AGENT" --status=in_progress --type=molecule --limit=1 --json | jq -r '.[0].id // empty')
+  CURRENT_WISP=$(gc bd list --assignee="$GC_AGENT" --status=in_progress --type=molecule --include-infra --limit=1 --json | jq -r '.[0].id // empty')
 fi
 NEXT=$(gc bd mol wisp mol-refinery-patrol --root-only --var target_branch={{target_branch}} --var rig_name={{rig_name}} --var binding_prefix={{binding_prefix}} --json | jq -r '.new_epic_id // empty')
 if [ -z "$NEXT" ]; then
@@ -268,7 +268,7 @@ Target: $TARGET"
   # bead must not end the merge lane.
   CURRENT_WISP=${GC_BEAD_ID:-}
   if [ -z "$CURRENT_WISP" ]; then
-    CURRENT_WISP=$(gc bd list --assignee="$GC_AGENT" --status=in_progress --type=molecule --limit=1 --json | jq -r '.[0].id // empty')
+    CURRENT_WISP=$(gc bd list --assignee="$GC_AGENT" --status=in_progress --type=molecule --include-infra --limit=1 --json | jq -r '.[0].id // empty')
   fi
   NEXT=$(gc bd mol wisp mol-refinery-patrol --root-only --var target_branch={{target_branch}} --var rig_name={{rig_name}} --var binding_prefix={{binding_prefix}} --json | jq -r '.new_epic_id // empty')
   if [ -z "$NEXT" ]; then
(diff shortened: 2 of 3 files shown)


edwarddavison:
Independent reproduction on another rig (gas-city), and one downstream finding for whoever bumps the pin.

**Repro corroborates this PR's premise.** Live wisp `gci-wisp-agu` (`issue_type=molecule`, `ephemeral=true`, `in_progress`, assigned to `gas-city/gastown.refinery`):

```
$ gc bd list --assignee="gas-city/gastown.refinery" --status=in_progress --type=molecule --limit=5 --json
[]
$ gc bd list --assignee="gas-city/gastown.refinery" --status=in_progress --type=molecule --include-infra --limit=5 --json
gci-wisp-agu
```

`--include-infra` matches `gc bd query "ephemeral=true AND status=… AND assignee='…'"` exactly, across `in_progress` and `open`, for both the refinery and the witness. The refinery was holding three wisps at the time (one `in_progress`, two `open`) — the leak in progress.

**Downstream finding: `gascity` CI will break at pin bump, not now.** `examples/gastown/gastown_test.go` reads pack content through the `gascity-packs` Go module, so it is green today. Two assertion styles, only one of which survives:

- `TestWitnessStartupAndNoIdleReconcileWisps` asserts the *prefix* `gc bd list --assignee="$GC_AGENT" --status=in_progress --type=molecule` — **unaffected**, since `--include-infra` is inserted after `--type=molecule`.
- The no-idle prompt-render order check asserts the *full* line, `--limit=1 --json` and all:
  ```go
  `CURRENT_WISP=$(gc bd list --assignee="$GC_AGENT" --status=in_progress --type=molecule --limit=1 --json | jq -r '.[0].id // empty')`
  `OPEN_WISPS=$(gc bd list --assignee="$GC_AGENT" --status=open --type=molecule --limit=0 --json | jq -r '.[].id')`
  ```
  These **will fail** once the pin moves — inserting `--include-infra` before `--limit` breaks the exact-substring match.

So the pin bump needs a companion `gascity` commit updating those two strings. Flagging so it is not hit as a surprise red CI on an unrelated PR. Same applies to #266 and #264.

Refs: gci-853, gci-byp

PR #282, opened 2026-08-07
Gastown dog and patrol formulas mis-handle a missing work-bead id under the claude provider

figgeous:
Three gastown formulas mis-handle a missing work-bead id: the shutdown-dance (the "due process" flow that decides whether to kill a stuck agent) silently operated on an *empty* warrant id, and the patrol formulas fell back to a database query that can pick the wrong item when an agent has several in flight. This change makes the dance resolve the warrant it actually claimed and fail loudly when it cannot, and lets the deacon patrol use the spawn trigger under an explicitly recorded safety precondition.

### Root cause

- Some providers never export `GC_BEAD_ID`. There the id arrives as `GC_TRIGGER_WORK_BEAD_ID` — **the bead that triggered the session, fixed at spawn**. It is *not* advanced by `gc hook --claim`, so in a pool it can name a bead this session never claimed.
- The shutdown-dance snippets used the id unguarded and without strict shell mode, so it expanded to empty — warrant verification and closure ran against an empty id, risking a warrant being misread or mis-closed.
- Naively preferring the trigger is not a safe fix for the dance: under dog-pool contention it would read and **close the wrong bead**. The dance belongs to the warrant *this session claimed*.
- The deacon patrol guarded the id but then always fell through to a "first in-progress molecule" lookup, which is ambiguous when more than one is assigned.

### Fix

- **Shutdown-dance:** normalize the id at the two id-establishing sites (the preamble block and `receive-warrant` step 1) by resolving the session's **claim first**, falling back to the trigger only when nothing is claimed, and aborting loudly when neither exists:
  ```bash
  GC_BEAD_ID="${GC_BEAD_ID:-$(gc hook current --id-only 2>/dev/null)}"
  GC_BEAD_ID="${GC_BEAD_ID:-${GC_TRIGGER_WORK_BEAD_ID:?no work bead id in env}}"
  ```
  Order matters: appending the resolver *after* the trigger instead is inert, because `${A:-${B:-$(C)}}` never evaluates `$(C)` while `B` is non-empty, so a stale trigger would still win.
- **Deacon patrol:** prefer the trigger before the lookup query. This is safe *only* because the deacon is a singleton with a fresh wake (`max_active_sessions = 1`, `wake_mode = "fresh"`), so one process env holds exactly one wisp. That precondition is recorded in the formula and pinned by tests.
- **Refinery patrol:** deliberately **not** changed. It rotates wisps in-session, so a spawn-fixed trigger goes stale mid-loop; the bare `${GC_BEAD_ID:-}` plus live assignee query is the only form that tracks the rotation.

### Tests

`gastown/tests/test_gastown_pack_assets.sh` gains a discriminator test that pins the resolution *form and order* per role: an order-sensitive signature over the dance's two normalization sites, a count pin on the deacon's trigger preference, a negative grep rejecting the inert nested form, and `agent.toml` pins on the deacon's singleton/fresh-wake preconditions. It is registered unconditionally at the runner and executed by CI (`.github/workflows/ci.yml` loops every `gastown/tests/test_*.sh`). All three TOML files parse cleanly; registry validation passes.

---

<details>
<summary>Machine detail</summary>

> **Maintainer note:** this section was rewritten during adoption review. The originally-proposed approach (prefer `GC_TRIGGER_WORK_BEAD_ID` everywhere, including 5 refinery sites) was found to introduce a wrong-bead-close hazard under pool contention and to break the refinery's in-session wisp rotation. The refinery changes were reverted and the dance was changed to claim-first ordering. The description above reflects what actually ships.

**Changed files** (head `354ba3a`, branch `fix/ga-oyp60-gc-bead-id-fallback`, 3 files +148/−1)
- `gastown/formulas/mol-shutdown-dance.toml` (+17) — claim-first normalization at the preamble snippet block and the `receive-warrant` step-1 block (the two id-establishing sites; all later snippets rely on the preamble's "run before any snippet" contract, same as `$target`/`$reason`/`$requester`, so a fresh shell must re-run the preamble). Prose documents the tiebreak order, the fresh-shell rule, and why the appended-resolver form is inert.
- `gastown/formulas/mol-deacon-patrol.toml` (+8/−1) — 1 site: `CURRENT_WISP=${GC_BEAD_ID:-}` → `CURRENT_WISP=${GC_BEAD_ID:-${GC_TRIGGER_WORK_BEAD_ID:-}}`, with the singleton/fresh-wake/no-in-session-rotation preconditions recorded inline.
- `gastown/tests/test_gastown_pack_assets.sh` (+124) — the per-role resolution discriminator described above.
- `gastown/formulas/mol-refinery-patrol.toml` — **not touched** (byte-identical to base).

**Approach / blast radius**
Per-formula resolution rules (provider-agnostic) rather than changing the claude provider to export `GC_BEAD_ID`. Providers that export `GC_BEAD_ID` are unaffected — it wins the chain on every path. The deacon keeps the existing `gc bd list` query as the final fallback, so behavior under providers exporting neither variable is unchanged. The dance change converts a silent empty-id expansion into either a correct claimed-warrant id or a loud abort. No agent config, prompt template, or sibling formula is modified.

**Verification**
- Resolution matrix over the dance form (claimed + foreign trigger → claimed bead; agreement → same bead; nothing claimed → trigger fallback preserves the original provider fix; neither → `:?` aborts non-zero; exported `GC_BEAD_ID` wins untouched).
- Discriminator mutation coverage: reordering a site, the nested form, an added trigger-preferring deacon line, converting a refinery site, and flipping the deacon to two slots each turn the suite red; the restored tree is green.
- `python3 -c "import tomllib; tomllib.load(...)"` on all three formulas → OK; `python3 validate_registry.py registry.toml` → ok; `go test .` → ok; full `gastown` asset suite green at head.

**Post-merge verification**
- Next shutdown-dance run under a provider that does not export `GC_BEAD_ID`: the dog's transcript should show `gc bd show <real-warrant-id>` (the id this session claimed) in the receive-warrant step, and warrants should close with reasons naming the actual claimed warrant bead. Signal: no empty-id invocations and no `MALFORMED_WARRANT` closes against empty ids.
- Next deacon patrol burn-and-repour: `CURRENT_WISP` should resolve directly. Signal: the deacon stops invoking the `gc bd list --assignee=... --limit=1` fallback query on the resolve step when the trigger env is present.

</details>

### DIFF
--- a/gastown/formulas/mol-deacon-patrol.toml
+++ b/gastown/formulas/mol-deacon-patrol.toml
@@ -534,7 +534,7 @@ Handle any urgent messages that arrived during patrol. Archive the rest.
 
 **3. Resolve current wisp and pour next iteration BEFORE burning:**
 ```bash
-CURRENT_WISP=${GC_BEAD_ID:-}
+CURRENT_WISP=${GC_BEAD_ID:-${GC_TRIGGER_WORK_BEAD_ID:-}}
 if [ -z "$CURRENT_WISP" ]; then
   CURRENT_WISP=$(gc bd list --assignee="$GC_AGENT" --status=in_progress --type=molecule --limit=1 --json | jq -r '.[0].id // empty')
 fi
--- a/gastown/formulas/mol-deacon-patrol.toml
+++ b/gastown/formulas/mol-deacon-patrol.toml
@@ -534,6 +534,12 @@ Handle any urgent messages that arrived during patrol. Archive the rest.
 
 **3. Resolve current wisp and pour next iteration BEFORE burning:**
 ```bash
+# Preferring the spawn trigger is safe here only because the deacon is a
+# singleton with a fresh wake (agents/deacon/agent.toml: max_active_sessions = 1,
+# wake_mode = "fresh"), so one process env holds exactly one wisp and the
+# trigger cannot name a foreign molecule. Do not copy this order into a pooled
+# role, and do not keep it here if this formula ever rotates wisps in-session
+# or the deacon gains a second slot.
 CURRENT_WISP=${GC_BEAD_ID:-${GC_TRIGGER_WORK_BEAD_ID:-}}
 if [ -z "$CURRENT_WISP" ]; then
   CURRENT_WISP=$(gc bd list --assignee="$GC_AGENT" --status=in_progress --type=molecule --limit=1 --json | jq -r '.[0].id // empty')
--- a/gastown/formulas/mol-refinery-patrol.toml
+++ b/gastown/formulas/mol-refinery-patrol.toml
@@ -117,7 +117,7 @@ If context feels heavy (multi-hour session, large recent diffs, or you notice
 yourself taking shortcuts or summarizing prematurely), pour and assign the next
 wisp, burn the current wisp, then request a restart:
 ```bash
-CURRENT_WISP=${GC_BEAD_ID:-}
+CURRENT_WISP=${GC_BEAD_ID:-${GC_TRIGGER_WORK_BEAD_ID:-}}
 if [ -z "$CURRENT_WISP" ]; then
   CURRENT_WISP=$(gc bd list --assignee="$GC_AGENT" --status=in_progress --type=molecule --limit=1 --json | jq -r '.[0].id // empty')
 fi
@@ -240,7 +240,7 @@ gc bd update $WORK \
 5. Clean up temp branch: `git checkout {{target_branch}} && git branch -D temp`
 6. Pour next patrol iteration before burning:
 ```bash
-CURRENT_WISP=${GC_BEAD_ID:-}
+CURRENT_WISP=${GC_BEAD_ID:-${GC_TRIGGER_WORK_BEAD_ID:-}}
 if [ -z "$CURRENT_WISP" ]; then
   CURRENT_WISP=$(gc bd list --assignee="$GC_AGENT" --status=in_progress --type=molecule --limit=1 --json | jq -r '.[0].id // empty')
 fi
@@ -341,7 +341,7 @@ TARGET=$(gc bd show $WORK --json | jq -r '.[0].metadata.target // "{{target_bran
    - Clean up: `git checkout "$TARGET" && git branch -D temp`
    - Pour next patrol iteration before burning:
      ```bash
-     CURRENT_WISP=${GC_BEAD_ID:-}
+     CURRENT_WISP=${GC_BEAD_ID:-${GC_TRIGGER_WORK_BEAD_ID:-}}
      if [ -z "$CURRENT_WISP" ]; then
        CURRENT_WISP=$(gc bd list --assignee="$GC_AGENT" --status=in_progress --type=molecule --limit=1 --json | jq -r '.[0].id // empty')
      fi
@@ -433,7 +433,7 @@ Work bead: $WORK
 Existing PR: $EXISTING_PR
 Branch: $BRANCH
 Target: $TARGET"
-  CURRENT_WISP=${GC_BEAD_ID:-}
+  CURRENT_WISP=${GC_BEAD_ID:-${GC_TRIGGER_WORK_BEAD_ID:-}}
   if [ -z "$CURRENT_WISP" ]; then
     CURRENT_WISP=$(gc bd list --assignee="$GC_AGENT" --status=in_progress --type=molecule --limit=1 --json | jq -r '.[0].id // empty')
   fi
@@ -1141,7 +1141,7 @@ If auto_land = "false": skip this entirely.
 
 **2. Resolve the current wisp, pour next iteration, then burn this wisp:**
 ```bash
-CURRENT_WISP=${GC_BEAD_ID:-}
+CURRENT_WISP=${GC_BEAD_ID:-${GC_TRIGGER_WORK_BEAD_ID:-}}
 if [ -z "$CURRENT_WISP" ]; then
   CURRENT_WISP=$(gc bd list --assignee="$GC_AGENT" --status=in_progress --type=molecule --limit=1 --json | jq -r '.[0].id // empty')
 fi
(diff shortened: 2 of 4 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277292696

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277292765

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277310039

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @figgeous! This PR fixes a real and genuinely nasty class of bug: the shutdown-dance — the "due process" flow that decides whether to kill a stuck agent — was expanding an unset `GC_BEAD_ID` to the empty string and then running warrant verification and closure against it. You caught that under providers that never export `GC_BEAD_ID` the dance was operating on *nothing*, silently, on the one code path whose whole job is to be careful before killing something. That matters because it is a fail-silent bug in a safety flow, and because it was invisible from the passing side: nothing was red, the dance just quietly did the wrong thing.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- **Preferring the spawn trigger is not safe for the dance.** `GC_TRIGGER_WORK_BEAD_ID` is the bead that *triggered* the session, fixed at spawn, and `gc hook --claim` never advances it. In the dog pool that means the trigger can name a bead this session never claimed — so the original chain could read and **close the wrong warrant**. The final patch resolves the session's claim first (`gc hook current --id-only`) and falls back to the trigger only when nothing is claimed. The ordering is the fix, not the resolver: the appended form `${A:-${B:-$(C)}}` never evaluates `$(C)` while `B` is non-empty, so a stale trigger would still have won.
- **The refinery could not take this change at all.** The refinery patrol rotates wisps in-session — it pours the next wisp before burning the current one — so a spawn-fixed trigger goes stale *within* a single session. Propagating the fallback to its five resolution sites would have made it burn an already-burnt wisp and leak the live one, once per iteration. Those five sites were reverted; the refinery is now byte-identical to base.
- **The deacon's exemption is now recorded and enforced.** The deacon *may* prefer the trigger, but only because it is a singleton with a fresh wake (`max_active_sessions = 1`, `wake_mode = "fresh"`), so one process env holds exactly one wisp. That reasoning is written into the formula and pinned by tests, instead of living in someone's head.
- **Prose matched to code.** The formula's description of where normalization happens now names the two id-establishing blocks and states the fresh-shell rule, rather than implying every snippet normalizes.
- **The behavior change now has real test signal.** A per-role discriminator test pins the resolution *form and order*, with a negative grep that structurally rejects the inert nested form.

Review findings addressed:
- Wrong-bead-close hazard from trigger-first ordering at both dance normalization sites — fixed in `354ba3a`.
- Refinery in-session rotation regression from propagating the fallback — reverted in `354ba3a`.
- Formula prose asserting a normalization invariant the file did not satisfy — reworded in `354ba3a`.
- No test coverage for a behavior change — discriminator test added in `354ba3a`. It was checked by mutation rather than by inspection: reordering a site, the nested form, an added trigger-preferring deacon line, converting a refinery site, and flipping the deacon to two slots each turn it red, and the restored tree is green.
- The PR description's root cause and machine detail still described the original approach — corrected by a maintainer edit to the description on this PR.

Non-blocking follow-ups (shared for visibility — **not** completed in this PR, each filed so it is not lost):
- `mc-0z9tz` — *Test Evidence Quality, minor:* the new test justifies the refinery guard with the label "pooled", but the refinery and the deacon carry identical singleton/fresh-wake config, so that label cannot distinguish the two roles it is used to distinguish. The real discriminator is in-session wisp rotation. Wording only; the guard itself bites correctly today.
- `mc-592qb` — *Test Evidence Quality, minor:* the dance signature pins two exact strings, so a *novel-form* trigger resolution added to a third dance block would leave the suite green. A cardinality pin on `GC_TRIGGER_WORK_BEAD_ID` references would close that last channel.
- `mc-k9wjk` — *Test Evidence Quality, minor:* the census comment says three role-surfaces; the measured count is five, with two singleton prompt templates outside the net and their exemption undocumented. Both are correct today — this is coverage and documentation, not behavior.
- `mc-oxm3m` — *Test Evidence Quality, minor:* the deacon comment names three preconditions but only the two `agent.toml` ones are pinned. The third — "do not keep this if the formula ever rotates wisps in-session" — is the one no config pin can catch.
- `mc-1h9vo` — *cross-repo (gastownhall/gascity), minor:* `gc hook current --help` still claims pool shells never receive the trigger variable, and its usage example recommends the nested chain this PR's prose now warns against. Worth fixing so the runtime's own help stops handing out the inert pattern.

### Maintainer Changes
`354ba3a fix(gastown): resolve the claimed warrant before the spawn trigger, and pin the resolution discriminator`

```
 gastown/formulas/mol-deacon-patrol.toml   |   6 ++
 gastown/formulas/mol-refinery-patrol.toml |  10 +--
 gastown/formulas/mol-shutdown-dance.toml  |  18 ++++-
 gastown/tests/test_gastown_pack_assets.sh | 124 ++++++++++++++++++++++++++++++
 4 files changed, 150 insertions(+), 8 deletions(-)
```

Net across both commits: 3 files changed, 148 insertions(+), 1 deletion(-) — `mol-refinery-patrol.toml` nets out to no change. Your commit is preserved with your authorship; the branch was rebased onto current `main` and re-verified there.

### Final Review Status
Ready for the merge queue: final head `354ba3a` has passing required GitHub checks, and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/gascity-packs/actions/runs/34286044279

### Review Iterations
3 review passes performed. Iterations 1 and 2 requested changes — first on the refinery propagation and the prose invariant, then on the resolution ordering and the missing test pin — and iteration 3 approved after both gates were discharged together and each fix was verified by execution rather than by reading the commit message.

---

PR #283, opened 2026-08-07
feat(slack-mini): Socket Mode transport for ingress-free installs

nonathaj:
## Problem

slack-mini only takes inbound over the Slack Events API, which needs a public HTTPS Request URL — the README points at Tailscale Funnel. That rules the pack out entirely on networks that cannot accept inbound connections: corporate firewalls, laptops, anywhere a tunnel is blocked. That is the situation this change came out of.

## What this does

Adds Socket Mode as an alternative inbound transport. The adapter opens an outbound WebSocket to Slack instead of receiving webhooks, so it needs only egress to `slack.com`.

`SLACK_APP_TOKEN` (an `xapp-` app-level token with `connections:write`) is the entire switch:

- **set** → Socket Mode. `LISTEN_PUBLIC` is never bound, and `SLACK_SIGNING_SECRET` is no longer required.
- **unset** → the HTTP transport, behaving exactly as before, signature verification included.

The event payload inside a Socket Mode envelope is byte-identical to the webhook body, so `bridgeEvent` is reused verbatim — `socketmode.go` is transport only.

## Design notes

**Authenticity.** Socket Mode has no signing secret and no HMAC. The trust boundary is the app-token-authenticated connection the adapter itself opens. Because every bridged message is stamped with `SLACK_WORKSPACE_ID` as its account id, the socket path additionally drops events whose `team_id` does not match — without it, an app installed in a second workspace would be filed under the wrong account. The HTTP path is untouched and still verifies signatures.

**Reconnect is the steady state.** Slack recycles connections routinely. Failed dials back off 1s→30s. On Slack's ~10s advance disconnect warning the supervisor opens the replacement while the old connection keeps draining and acking, rather than tearing down and losing what is in flight.

**Liveness is a ping, not a read deadline.** The first cut bounded reads at 90s as an idle watchdog, which was wrong: Slack's keepalives are WebSocket pings answered inside the library without ever waking a read, and a healthy connection carries data only when someone mentions the bot. It tore down working connections every 90s. Second commit replaces it with a 30s application-level ping; a connection is dropped only when a ping goes unanswered. Both the bug and the fix are covered by tests.

**Envelopes are acked before bridging**, and bridging happens off the read loop so a slow gc cannot stall acks behind it.

## Dependency

Adds `github.com/coder/websocket` v1.8.15, which declares no requires of its own — `go.sum` stays two lines. This is the adapter's first dependency; the sibling `slack-full/cli` module already carries cobra, so pack binaries taking deps seemed established rather than novel. Happy to hand-roll RFC 6455 on stdlib instead if the maintainers prefer the adapters stay dependency-free — the `discord` pack's gateway service is precedent for that. One concrete argument for the library: it dials via `net/http`, so `HTTPS_PROXY`/`NO_PROXY` are honoured, which matters on exactly the restricted networks this feature targets.

## Manifest

One manifest cannot serve both transports (Slack refuses to deliver to a Request URL while Socket Mode is enabled), so this ships `manifest/app-socket.json` — identical to `app.json` but with `socket_mode_enabled: true`. The README now leads with a transport-choice table.

## Testing

`go vet` + `go test -race` green. 16 tests, including two that drive a real WebSocket via `websocket.Accept` rather than a stub: a full ack→bridge round trip, and one that kills the first connection and asserts the adapter reconnects and keeps bridging.

Validated against a live workspace on a network with no inbound ingress:

- `@`-mention → `envelope received` → acked → `POST /extmsg/inbound 200` → reached the target session.
- One connection held **over three hours with zero reconnects** after the keepalive fix. Before it, the same setup churned a reconnect every 90 seconds.
- HTTP mode re-checked for regressions: still binds `:8775`, still answers `/healthz`.

## Note for maintainers (not addressed here)

While testing, a delivered message sat unrouted in gc until an `extmsg` binding was created by hand. Tier 1 sets `explicit_target` and ships no bind verb (`bind-room`/`bind-dm` are Tier 2), so a stock slack-mini install can look silent even when the bridge is working. That looks like a separate gap — possibly just a docs step — and I have deliberately left it out of this PR, which is scoped to the transport.

### DIFF
--- a/slack-mini/CHANGELOG.md
+++ b/slack-mini/CHANGELOG.md
@@ -5,6 +5,37 @@ All notable changes to slack-mini are documented in this file.
 The format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+
+- **Socket Mode transport** (`adapter/socketmode.go`), an ingress-free
+  alternative to the public Events API listener: the adapter opens an
+  outbound WebSocket to Slack instead of receiving webhooks, so the pack
+  runs on networks that cannot accept inbound connections (corporate
+  firewalls, laptops, anywhere a tunnel is blocked).
+  - Selected by `SLACK_APP_TOKEN` (an `xapp-…` app-level token with the
+    `connections:write` scope). When set, `LISTEN_PUBLIC` is never bound
+    and `SLACK_SIGNING_SECRET` is no longer required — Socket Mode has no
+    request signatures, so the trust boundary is the app-token-authenticated
+    connection the adapter itself opens.
+  - Supervised reconnect with capped exponential backoff (1s → 30s). On
+    Slack's advance disconnect warning a replacement connection is opened
+    while the old one drains and keeps acking, so mentions are not dropped.
+  - Envelopes are acked before bridging; events from a workspace other than
+    `SLACK_WORKSPACE_ID` are dropped rather than filed under the wrong
+    account.
+  - Dialled through Go's default HTTP client, so `HTTPS_PROXY`/`NO_PROXY`
+    are honoured.
+- `manifest/app-socket.json`, the Socket Mode variant of the app manifest
+  (identical to `manifest/app.json` but with `socket_mode_enabled: true`).
+
+### Changed
+
+- The adapter now depends on `github.com/coder/websocket` v1.8.15, which
+  has no dependencies of its own. The HTTP transport is unchanged and still
+  verifies request signatures exactly as before.
+
 ## [0.1.0] — Tier 1 extraction
 
 Initial release. slack-mini is Tier 1 of the Slack pack family — the
--- a/slack-mini/CHANGELOG.md
+++ b/slack-mini/CHANGELOG.md
@@ -22,19 +22,30 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - Supervised reconnect with capped exponential backoff (1s → 30s). On
     Slack's advance disconnect warning a replacement connection is opened
     while the old one drains and keeps acking, so mentions are not dropped.
-  - Envelopes are acked before bridging; events from a workspace other than
-    `SLACK_WORKSPACE_ID` are dropped rather than filed under the wrong
-    account.
+    Every redial is paced, including the warned one — a 1s pause fits inside
+    the ~10s warning, so an ordinary refresh still overlaps — and the delay
+    escalates only for connections that fail to last 30s, so a server that
+    accepts and immediately drops connections cannot be redialled in a
+    tight loop.
+  - Envelopes are acked before bridging.
   - Dialled through Go's default HTTP client, so `HTTPS_PROXY`/`NO_PROXY`
     are honoured.
 - `manifest/app-socket.json`, the Socket Mode variant of the app manifest
-  (identical to `manifest/app.json` but with `socket_mode_enabled: true`).
+  (`manifest/app.json` with `socket_mode_enabled: true` and its own display
+  description; nothing else differs).
 
 ### Changed
 
 - The adapter now depends on `github.com/coder/websocket` v1.8.15, which
-  has no dependencies of its own. The HTTP transport is unchanged and still
-  verifies request signatures exactly as before.
+  has no dependencies of its own.
+- Events from a workspace other than `SLACK_WORKSPACE_ID` are dropped
+  rather than filed under the wrong account. This guard sits in the shared
+  bridge, so it covers **both** transports: neither establishes which
+  workspace an event belongs to (Socket Mode has no signature, and the HTTP
+  path's HMAC proves only that Slack sent the request for this app), while
+  every bridged message is stamped with `SLACK_WORKSPACE_ID` as its account
+  id. Otherwise the HTTP transport is unchanged and still verifies request
+  signatures exactly as before.
 
 ## [0.1.0] — Tier 1 extraction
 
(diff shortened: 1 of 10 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277292554

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277292623

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277309856

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @nonathaj! This PR adds a Socket Mode transport to `slack-mini`, so the pack can run against Slack with no public ingress at all — an operator sets `SLACK_APP_TOKEN` and the adapter dials out over a WebSocket instead of waiting on an inbound Request URL. That removes the tunnel/reverse-proxy prerequisite that made `slack-mini` awkward to adopt for local and private-network cities, and it does so additively: the HTTP receiver is untouched behind the same selector, the shared bridge is reused verbatim, and the whole transport is cleanly revertible. The second commit in this PR, which caught your own 90s read deadline tearing down healthy idle connections, is the kind of self-correction that makes a new transport trustworthy.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- **The idle-liveness regression test could not fail on the bug it was written for.** `TestPumpSocketSurvivesIdleConnection` waited 500ms against the 90s deadline it existed to catch, so it also passed against the pre-fix code, and deleting the `go keepaliveSocket(...)` wiring left the whole suite green. The cause was structural rather than careless: every connection timer was read directly at its call site, so no test could reach the timing layer at all. The final patch introduces a `socketTimings` struct threaded through `runSocketMode` → `runSocketConnection` → `pumpSocket`; production keeps its real constants via `defaultSocketTimings()` while tests drive the same code paths at millisecond scale, and `defaultSocketTimings()` is itself pinned to literal durations so a fast test suite cannot hide drift in the production values.
- **Two reconnect paths had no pacing.** The disconnect-warning path redialled with zero delay, and `backoff` reset to 1s on every successful dial, so a server that accepts a connection and immediately drops it was redialled at network speed, bounded only by Slack's own throttling. Both arms now pace, and `settleBackoff` resets the delay only once a connection has survived `socketMinLifetime`.
- **The workspace-identity guard covered only one transport.** It was added on the socket path, justified by a comment stating the HTTP path proves workspace identity via the signing secret — the HMAC actually proves Slack sent the request for *this app*, not which workspace the event belongs to. The check moved into `bridgeEvent`, the funnel both transports feed, with no wire change: `team_id` was already decoded on `slackEventEnvelope` at the base commit.

Review findings addressed:
- **Test Evidence Quality — minor, high confidence (synthesis):** the idle-liveness regression test was inert and the keepalive wiring was unpinned. Fixed in `52475e3`, with each new test verified red against a mutation of the code it covers before being accepted.
- **Error Handling & Resilience — minor, high confidence (two reviewer lanes, independently):** the disconnect-warning reconnect had no delay and the backoff reset on every successful dial. Fixed in `52475e3`.
- **Contract & Interface Fidelity — minor, high confidence (one lane, mechanically confirmed in synthesis):** the workspace guard was socket-only and the comment justifying that asymmetry was inaccurate. Fixed in `52475e3`.
- **Change Impact / Blast Radius — nit, high confidence:** `README.md` still advertised "a single-file adapter" after the file split, contradicting the pack's own build docs. Fixed in `52475e3`.
- **Architectural Consistency — nit, high confidence:** the CHANGELOG and README said the two manifests differ only in `socket_mode_enabled`; they also differ in `display_information.description`. Fixed in `52475e3`.

- `mc-dzksm`: **Readability of the new connection-lifecycle functions.** `runSocketMode` and `pumpSocket` measure above the complexity budget for code a diff authors. The maintainer fixup removed the duplicated backoff blocks (`sleepFor`/`nextBackoff`/`settleBackoff`), but extracting `runSocketMode`'s reconnect-lifecycle arms — the change that would bring both functions back under budget — was deliberately deferred as a larger restructuring. Worth noting that a meaningful share of the measurement is ordinary Go error-guard nesting, so this is a readability improvement rather than a defect. Tracked separately and not addressed in this PR; it did not gate the merge.

### Maintainer Changes
One maintainer commit was pushed to this PR branch:

`52475e3 fix(slack-mini): pin the Socket Mode timing layer and share the workspace guard` — 6 files changed, 519 insertions(+), 61 deletions(-), entirely under `slack-mini/`.

Two things worth flagging explicitly, since both are deliberate deviations from the findings as written:

- Hoisting the workspace guard into `bridgeEvent` means this PR no longer leaves the HTTP transport bit-for-bit unchanged. The CHANGELOG entry was corrected rather than left standing, the guard is now documented as covering both transports, and the README's `SLACK_WORKSPACE_ID` row warns that a wrong value presents as a silent bot. An absent `team_id` still passes, since Slack sends one on every `event_callback`.
- The warned-path reconnect sleeps the *current* backoff rather than the minimum. That is safe because the lifetime gate landed alongside it: an ordinary Slack refresh has lasted well past `socketMinLifetime`, so the delay resets to 1s first and still fits inside Slack's ~10s advance warning. Only a connection warned moments after opening carries an escalating delay — which is the connection-limit churn scenario the finding itself identified as most plausible.

Your commits are preserved with their original authorship. The branch was rebased onto current `main` (`a54cb74`) before the push, with an unchanged patch-id.

### Final Review Status
Ready for the merge queue: final head `52475e3` has passing required GitHub checks, and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/gascity-packs/actions/runs/34220733576

### Review Iterations
One review pass across three reviewer lanes plus synthesis, reaching approve with no blocker or major finding in any category. Every line anchor in all three lane reports was re-verified against the tree, and two lane claims that the idle case was already covered were corrected by measurement rather than carried forward. The five minors and nits were then applied in a single maintainer fixup and revalidated with `gofmt -l`, `go build ./...`, `go vet ./...`, and `go test -race -count=1 ./...` on the rebased head before pushing.

---

PR #285, opened 2026-08-08
fix(gascity): assert work-item dependency orientation at decomposition time

superlzyguy:
## Problem

A decomposition step that declares a strict sequential chain of work items can materialize every dependency edge inverted. The failure mode: `gc bd dep add <a> <b>` gets read temporally ("a comes before b") instead of by its actual contract ("a depends on b"), so each EARLIER work item ends up depending on its successor and the implementation convoy drains the chain back-to-front.

Observed twice in a production Gas City deployment running `build-basic`:

- A 6-item chain (WI-1 → … → WI-6, "each depends on its predecessor" per the artifact) came out complete but fully inverted: WI-1 depended on WI-2, … WI-5 on WI-6, and WI-6 had no dependencies at all. The plan-wide verification sweep (WI-6) drained FIRST at index 0/6 with nothing to verify, exhausted its 3-attempt budget, closed `gc.outcome=fail` / blocked-precondition, and filed a duplicate follow-up bead. The work item owning a live data re-pull ran LAST, so its consumers priced from a stale snapshot — harmless only because the re-pull happened to show zero drift.
- A second run's 4-item drain executed in exact reverse of declared dependency order — same inversion.

The `[steps.check]` gate validates only the artifact (schema / front matter / coverage); nothing asserts that the created edges match the declared order. The gascity pack's decompose instructions never mention dependency wiring at all — while the older gastown pack prompts carry an explicit WRONG/RIGHT warning about this exact argument-order trap (`gc bd n phase2 phase1`, not `phase1 phase2`), so this is a known failure mode that the newer pack lost.

## Fix

Creation-time assertion, per the incident analysis: assert edge orientation when materializing a declared sequence — not at drain time, and not by adding more edges.

- **Instructions** (`build-basic`, `decomposition-base`, `build-from-decompose-base` decompose steps): spell the command contract — first argument WAITS, second is what it waits for — with a RIGHT/WRONG pair, require one `gc bd dep add` per consecutive pair for chains (later item named first), and require a machine-readable dependency table in the Work Items section (`| ID | Bead | Depends On |`, rows in execution order).
- **New `gascity/assets/scripts/validate_decomposition_edges.py`**: verifies the declaration (Depends On may only reference earlier rows — a forward reference means the table is not in execution order) and the live graph via `gc bd dep list --json` (each bead depends on its declared prerequisites; no earlier bead depends on a later one through a ready-blocking type: `blocks` / `waits-for` / `conditional-blocks`, mirroring gascity's `readyBlockingDependencyTypes`). Failures print the exact `gc bd dep add` / `gc bd dep remove` repair commands so the bounded repair loop can rewire deterministically instead of guessing from prose.
- **`build-artifact-valid.sh`**: after schema validation passes, runs the edge check for `gc.build.decomposition.v1` artifacts.

## Compatibility

- Artifacts that declare no dependency table (methodology packs with other task shapes, e.g. superpowers) skip live verification and validate exactly as before; declaring the table opts the artifact into strict orientation checking.
- The dependency table deliberately has no Status column: `validate_build_artifact.py` treats any ID/Status table as the coverage matrix, and both the new instructions and the edge checker guard against that collision.
- Schema `gc.build.decomposition.v1` is untouched.

## Testing

- `python3 -m pytest gascity/tests -q` — 184 passed, including 8 new tests: table parsing/skip behavior, forward-reference rejection, Status-column rejection, correct chain passes live verification, the observed inverted-chain shape fails with the right repair commands, missing bead fails, non-blocking dep types ignored, coverage-matrix non-interference.
- Full CI pytest sweep (`tests contributing/tests gascity/tests discord/tests github/tests slack-full/tests slack-channel/tests pr-pipeline/tests profiler/tests`): 1179 passed, 8 pre-existing skips.
- `bash -n` on the modified check script.

### DIFF
--- a/gascity/assets/scripts/checks/build-artifact-valid.sh
+++ b/gascity/assets/scripts/checks/build-artifact-valid.sh
@@ -97,11 +97,34 @@ for candidate in \
 done
 [ -n "$VALIDATOR" ] || fail "validate_build_artifact.py not found beside $SCRIPT_DIR or under GC_WORK_DIR"
 
-if OUTPUT="$(python3 "$VALIDATOR" --schema "$SCHEMA" --path "$ARTIFACT_PATH" 2>&1)"; then
-  echo "build artifact valid: schema=$SCHEMA path=$ARTIFACT_PATH"
-  exit 0
+if ! OUTPUT="$(python3 "$VALIDATOR" --schema "$SCHEMA" --path "$ARTIFACT_PATH" 2>&1)"; then
+  echo "build-artifact-check: schema=$SCHEMA path=$ARTIFACT_PATH failed validation" >&2
+  printf '%s\n' "$OUTPUT" >&2
+  exit 1
 fi
 
-echo "build-artifact-check: schema=$SCHEMA path=$ARTIFACT_PATH failed validation" >&2
-printf '%s\n' "$OUTPUT" >&2
-exit 1
+# Decomposition artifacts additionally bind live bead edges to the declared
+# work-item order: a sequential chain wired backwards drains back-to-front, so
+# assert edge orientation at creation time rather than at drain time.
+if [ "$SCHEMA" = "gc.build.decomposition.v1" ]; then
+  EDGE_CHECK=""
+  for candidate in \
+    ${GC_WORK_DIR:+"$GC_WORK_DIR/gascity/assets/scripts/validate_decomposition_edges.py"} \
+    "$(dirname "$VALIDATOR")/validate_decomposition_edges.py"; do
+    if [ -n "$candidate" ] && [ -f "$candidate" ]; then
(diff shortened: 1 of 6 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277292334

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277292407

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277309812

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277329471

PR #287, opened 2026-08-08
fix(gastown): honor rig default_branch via --base in worktree-setup

AJBcoding:
## Problem

`worktree-setup.sh` silently falls back to `origin/HEAD` when creating a new agent worktree branch. A rig pinned to an integration branch (`default_branch` in rig config) gets its polecat/refinery worktrees cut from `main` instead, so agents debug a tree that does not match the deployed binary. Tracked internally as gas-e6r.

## Change

- **`gastown/agents/polecat/agent.toml`, `gastown/agents/refinery/agent.toml`** — append `--base {{.DefaultBranch}}` to the `pre_start` invocation (these are the only two agents that invoke the script).
- **`gastown/assets/scripts/worktree-setup.sh`** — accept `--base <branch>` / `--base=<branch>` as an order-independent trailing flag in both positional signatures. Resolution for an explicit base: `refs/remotes/origin/<base>` (after a best-effort fetch), then `refs/heads/<base>` (a local-only integration branch is a valid pin). If neither resolves: **hard error, no fallback to origin/HEAD**. Unknown arguments are now rejected (exit 2) since this script decides which commit an agent's whole worktree is cut from. With no `--base`, behavior is unchanged except the last implicit case (origin/HEAD unset) now announces itself on stderr instead of silently using the current checkout.

## ⚠️ Deploy ordering

This pack change must not reach a city before a `gc` binary that expands `{{.DefaultBranch}}` in `pre_start` templates (gastownhall/gascity `c49e900`, gas-e6r). Older binaries keep the raw template string via `expandSessionSetup`'s graceful fallback, and the strict script correctly rejects the literal `{{.DefaultBranch}}` — hard-failing every agent spawn. Order: gc binary first, then this pack.

## Verification

16-case harness against real origin+clone repos (local-only base branch, pushed base branch, missing base → loud failure with no worktree left behind, unexpanded-template rejection, no `--base` → origin/HEAD unchanged, flag ordering, `--base=value`, idempotent re-run, legacy positional signature with and without `--base`, unknown-arg rejection, staged-path preserve/restore):

- patched script: **16/16 pass**
- current script: fails 11 of the same checks — every requested base silently resolves to the default branch


AJBcoding:
@quad341 Ready for review and merge when you have a moment — I can't request reviewers directly (outside contributor), so flagging here.

**Verification: 16/16 pass** on a harness that builds real origin+clone repos and exercises every base-selection path: local-only base branch, pushed base branch, missing base → loud failure with no worktree left behind, unexpanded-template rejection, no `--base` → origin/HEAD behavior unchanged, flag ordering, `--base=value`, idempotent re-run, legacy positional signature with and without `--base`, unknown-arg rejection, and staged-path preserve/restore. The pre-change script fails 11 of the same checks — every requested base silently resolves to the default branch.

**Downstream production validation, same-day:** the gc-side prerequisite (pre_start `{{.DefaultBranch}}` expansion, `c49e900`) is deployed in a downstream city binary (gascity deploy train 2026-08-08, `9b84906d6` lineage) and symbol-verified there — on that deployment this pack change is the only missing half, and the patched script is the exact one the 16-case harness validated. For consumers whose gc predates the expansion, the ⚠️ deploy-ordering note in the PR body applies as written.

sjarmak (changes requested):
The script work here is solid and I verified it rather than taking it on trust. Running the patched `worktree-setup.sh` against real throwaway clones: `--base <existing-remote-branch>` cuts the worktree from that branch's tip instead of `origin/HEAD`; `--base <nonexistent-branch>` hard-fails with a message naming both refs it checked and leaves no half-created worktree behind; a missing `--base` value exits 2 cleanly; and the no-`--base` path is byte-for-byte the old fallback, so rigs without `default_branch` are unaffected. `set -eu` posture holds throughout and I found no unquoted expansion that could turn an empty variable into a bad `rm`, `cd`, or worktree target. No role branching in the script.

The blocker is the wiring, not the script.

**`{{.DefaultBranch}}` does not exist in the context that renders `pre_start`.**

`pre_start` is expanded through `expandSessionSetup(cfgAgent.PreStart, setupCtx)` where `setupCtx` is a `SessionSetupContext`. On current `gastownhall/gascity` `main` that struct carries exactly nine fields: `Session`, `Agent`, `AgentBase`, `Rig`, `RigRoot`, `CityRoot`, `CityName`, `WorkDir`, `ConfigDir`. There is no `DefaultBranch`. It exists only on `PromptContext`, which renders `.md` prompt templates, not `pre_start`.

The failure mode is worse than a clean rejection. `expandSessionSetup` does not fail per-token: on any `Execute` error it discards the rendered buffer and falls back to the entire raw command string (`result[i] = raw`). `pre_start` then runs that literal text through `sh -c`, so what actually happens is:

```
sh: 1: {{.ConfigDir}}/assets/scripts/worktree-setup.sh: not found
exit=127
```

The script is never invoked, so its careful `--base` validation never gets a chance to report anything. That surfaces as `pre_start[0]: <err>` and blocks session startup for every polecat and refinery on any city running a currently-released binary. Both agent.toml files take the identical change, so they break identically rather than divergently, which is at least consistent.

The PR body names a companion `gc` commit (`c49e900`) that adds the field. That commit is not an ancestor of `gastownhall/gascity` `main` (`compare` reports diverged, ahead 35, behind 379), and no open PR there references it. `gastown/pack.toml` has no minimum-`gc`-version gate, so nothing stops an older binary from pulling this pack version.

Route: land the Go-side `SessionSetupContext.DefaultBranch` field in `gastownhall/gascity` first, confirm it is in a released binary, then merge this. If the pack format grows a binary-version gate, that would work too. Happy to review the Go-side change when it is up.

Minor, worth doing since this is the commit that first pipes external config into that argv position: `default_branch` is API-patchable with no format validation beyond a trim, so a value starting with `-` would be read by `git` as an option. Adding `--` separators (`git -C "$RIG_ROOT" fetch origin -- "$BASE"`, and likewise on the `show-ref --verify` calls) closes that cheaply.

PR #288, opened 2026-08-08
fix(gastown): accept a local-only base branch in mol-polecat-work

AJBcoding:
## Problem

The `workspace-setup` step of `mol-polecat-work` hard-requires `{{base_branch}}` to exist on origin, at four sites: worktree creation (`git worktree add --detach origin/{{base_branch}}`), branch creation (`git checkout -B ... origin/{{base_branch}}`), the rejection-resume rebase, and both `fork_sha` recordings.

For a rig pinned to an unpushed integration branch (e.g. `integration/deploy-20260804`, which exists only as a local ref), every one of those refs is missing: the fetch is a no-op, the step fails at worktree creation — before ever reaching the branch-creation block — and an agent recovering with the only fetchable ref lands on `origin/main`. That is the gas-e6r wrong-base outcome reached by a second route (tracked as gas-cqh, sibling of the #287 script fix).

## Fix

Step 1 now resolves the base **once**, in the same order as `worktree-setup.sh --base` in #287, so the two paths agree on what a base is:

1. `refs/remotes/origin/<base>` after the fetch — preferred, avoids local drift;
2. `refs/heads/<base>` — a local-only integration branch is a valid pin (announced with a note);
3. neither → **STOP loudly** (`gc runtime drain-ack; exit 1`, the step's existing STOP idiom) instead of substituting a different base — silent fallback is what made gas-e6r invisible.

All later blocks in the step use the resolved `$BASE_REF` — worktree add, branch creation, rejection rebase, and both `fork_sha` recordings — each guarded by `: "${BASE_REF:?...}"` so a block re-run before resolution fails with the remedy named rather than a confusing empty-string git error.

The original branch-from-freshly-fetched-remote rationale is preserved verbatim in intent: when the base exists on origin, the branch is still cut from the just-fetched remote-tracking ref, never a stale local copy (whose already-merged commits with stale hashes the refinery rebase rejects as spurious duplicates).

## Verification

A 24-check harness extracts the exact ```bash blocks from this file (placeholders rendered, `gc` stubbed), syntax-checks them, and executes them against real origin+clone repos:

- **A — base on origin, local copy stale**: resolves to `origin/<base>`; branch cut from the freshly-fetched remote tip, not the stale local ref; `fork_sha` records the remote tip.
- **B — base local-only** (the failing rig shape): resolves to the local branch; worktree add, branch creation, `git rebase "$BASE_REF"`, and `fork_sha` all succeed; branch provably NOT redirected to `origin/main`.
- **C — base nowhere**: nonzero exit, STOP message names the base, no fallthrough, drain-ack fired.
- **D — guard**: creation block without step 1 fails with the remedy message; no branch created.
- **E — control**: the **unpatched** blocks against the local-only shape hit `fatal` at worktree add and branch creation and leave no branch — the bug reproduced.

Patched: 24/24 pass. TOML re-parses cleanly (`tomllib`).

## Relationship to #287

Formula half of the same defect family; #287 fixes the worktree-setup script path. Independent files, no conflict — they can land in either order or together. Unlike #287, this change has **no deploy-ordering hazard**: it introduces no new template variables (only the existing `{{base_branch}}`), so it is safe to vendor into cities running any gc binary.

### DIFF
--- a/gastown/formulas/mol-polecat-work.toml
+++ b/gastown/formulas/mol-polecat-work.toml
@@ -73,7 +73,7 @@ Every check is idempotent — safe to re-run after crash/restart.
 2. `metadata.target` on the parent convoy chain
 3. the rig repo's default branch
 
-**1. Fetch latest and derive your work bead:**
+**1. Fetch latest, derive your work bead, and resolve the base ref:**
 ```bash
 git fetch --prune origin
 CONVOY_STATUS=$(gc convoy status {{convoy_id}} --json)
@@ -92,6 +92,32 @@ fi
 gc bd update "$WORK_BEAD_ID" --unset-metadata handoff_stage
 ```
 
+Then resolve what `{{base_branch}}` means in this clone, remote-first.
+A base that exists on origin is used via its freshly-fetched
+remote-tracking ref. A base that exists only locally (an unpushed
+integration branch) is still a valid pin — use the local branch. If
+neither exists the base configuration is wrong: STOP rather than
+substituting a different base, because silently falling back to the
+repo default is how a rig pinned to an integration branch got polecats
+branched off main (gas-e6r).
+
+```bash
+if git show-ref --verify --quiet "refs/remotes/origin/{{base_branch}}"; then
+    BASE_REF="origin/{{base_branch}}"    # preferred: freshly fetched, no local drift
+elif git show-ref --verify --quiet "refs/heads/{{base_branch}}"; then
+    BASE_REF="{{base_branch}}"           # local-only base (unpushed integration branch)
+    echo "note: base branch {{base_branch}} is local-only — it does not exist on origin"
+else
+    echo "STOP: base branch {{base_branch}} exists neither on origin nor locally."
+    echo "Fix the rig/bead base configuration; do not substitute a different base."
+    gc runtime drain-ack
+    exit 1
+fi
+```
+
+Every later block in this step uses `$BASE_REF` — never assume
+`origin/{{base_branch}}` exists.
+
 **2. Ensure worktree exists.**
 
 Check if `metadata.work_dir` already records your worktree path:
@@ -108,8 +134,9 @@ to create a new one.
 
 **If no worktree** — create one scoped to the bead, not the agent:
 ```bash
+: "${BASE_REF:?base ref not resolved — run step 1 first}"
 WORKTREE_PATH=$(pwd)/worktrees/"$WORK_BEAD_ID"
-git worktree add "$WORKTREE_PATH" --detach origin/{{base_branch}}
+git worktree add "$WORKTREE_PATH" --detach "$BASE_REF"
 cd "$WORKTREE_PATH"
 ```
 Record immediately so restarts and witness recovery can find it:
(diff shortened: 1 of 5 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277292050

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277292188

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277309726

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277316668

julianknutsen:
## Maintainer Adoption Review

### Original PR Review

Thank you, @AJBcoding — this is a genuinely good catch. `mol-polecat-work` assumed the base
branch always exists on `origin`, and every git operation in the formula reached for
`origin/{{base_branch}}` directly. On a rig whose base branch is local-only that assumption
doesn't announce itself: the ref simply doesn't resolve, and the failure surfaces later and
somewhere else. Your diagnosis of the root cause — and the remote-first / local-fallback
resolution you proposed for it — is what the whole change is built on.

Review ran four iterations, and each one went deeper into the same theme rather than off into
polish: the fix was correct where you put it, and the work was making it hold everywhere else
in the formula too.

**Specific gaps fixed:**

- The resolved base didn't survive a step boundary. `BASE_REF` was resolved in step 1 but
  `self-review` and the `submit` recovery text still reached for the bare ref, so the fix held
  for one step and lapsed for the rest. Now the resolved value is stamped onto the work bead as
  `base_ref` and read back downstream, which is what makes the promise true for the whole
  formula.
- The refinery hand-off was still open. A local-only base could reach the refinery, fail to
  resolve its target, and land in the "FAILED (conflicts)" arm that runs `delete-source` /
  `reopen-source` — a destructive path for what is really a precondition problem. Polecat now
  halts that work at `submit-and-exit` with `halt_reason=base_branch_local_only`, hands the bead
  back cleanly, and mails the witness instead.
- `affected_tests_command` was respecified against the durable resolved base and now fails
  closed on an unresolvable range, rather than quietly testing nothing.
- The both-refs-exist case is handled explicitly, with an ancestry check that stops on
  local-ahead or divergence instead of silently preferring `origin`.
- A structural lint was added so this class of bug can't come back: no bare
  `origin/{{base_branch}}` git-ref use outside the resolution block, with the STOP arm's
  presence pinned too. A literal-fragment pin alone catches deletion but stays green against
  inversion or neutralization — this one catches all three.

**Review findings addressed:**

- Threading the resolved base across every step via a durable carrier (blocker) — fixed.
- Closing the refinery hand-off without using the destructive reopen path (blocker) — fixed.
- `affected_tests_command` failing closed on an unresolvable range (major) — fixed.
- The structural bare-ref lint, including the STOP arm (major) — added.
- Explicit both-refs-exist ancestry handling (major) — fixed.
- A `BASE_REF` guard on the step-3 block, mirroring the existing guard in block 4 (major) — added.
- Making the refinery invalid-upstream STOP durable and non-blocking, with mail escalation
  naming the work and target beads (major) — fixed.
- Bringing `polecat_base_ref_problems` back under the complexity band by pure extraction rather
  than by rewriting it (major) — done, and verified as a pure extraction against the pre-change
  behaviour.

**Non-blocking follow-ups (shared for visibility — NOT completed in this PR):**

- `mc-im55u`: two fail-closed branches in the `polecat_base_ref_*` lint family are unpinned —
  deleting either leaves the suite green. One is a genuine fail-open (a formula that never
  assigns `BASE_REF` would report zero problems); the other was measured and does not fail open.
  Both are green on the pre-extraction code too, so neither is a regression from this PR.
- `mc-nhh8k`: the new stop-arm extractor returns `X | None` while the section's two established
  extractors return `(data, problems)`. Defensible as written — a convention note, not a defect.
- `mc-qblsb`: the new `base_branch_local_only` halt is counted by the push-gate test but not
  covered by its escalation and ordering assertions. All those properties hold today; none is
  pinned. See "Maintainer Changes" below for why this was left out rather than fixed here.

### Maintainer Changes

Your commits are preserved as authored; the maintainer work is a single follow-on commit.

The base moved twelve commits while review was running, and four of those touch the same files
this PR does. The branch no longer rebased or merged cleanly, so the conflicts were resolved by
hand. Three of the five conflicting hunks were straightforward keep-both. The other two were
real disagreements worth naming:

- **#290** landed a branch-content gate in the same step, and it measured the branch with a bare
  `origin/{{base_branch}}` — exactly the pattern your new lint exists to reject. Resolved toward
  your invariant rather than away from it: #290's gate now reads the carried `$BASE_REF`, and
  the three coupled literal pins were updated to match. On a branch with a remote base the count
  is identical; on a local-only base it stops reporting a good branch as unmeasurable. The lint
  was not weakened.
- **#273** landed a rule that every `gc bd update` in this step must use `--append-notes`,
  because a bare `--notes` replaces the whole field — it destroyed a project lead's instruction
  text twice on a live rig. Your `base_branch_local_only` halt predates that rule and used a bare
  `--notes`; it now appends.

**This PR therefore touches a fifth file it didn't before**,
`gastown/tests/test_polecat_push_gate.sh`, which arrived with #273 and did not exist when you
opened this. It pins the exact number of halts in the step, and your new halt is a legitimate
fifth — so the count moved from four to five and its explanatory comment now names your halt.
There is no way to land both changes without touching that file. Its `HALTS` table was
deliberately left alone: adding an entry would assert more than either PR shipped, which isn't a
conflict resolution, so it's filed as `mc-qblsb` instead of done quietly here.

Full validation at the merge head: the pack test suite, the registry check, and all five Gastown
shell suites pass. The one failing test, `test_every_universal_waiver_entry_is_still_reported`,
reproduces identically on the untouched base branch and is unrelated to this PR.

### Final Review Status

Approved. No required changes outstanding.

CI: https://github.com/gastownhall/gascity-packs/actions/runs/34388985835

### Review Iterations

Four. Iterations 1 and 2 raised blockers and majors, all addressed; iteration 3 directed a
complexity reduction by pure extraction; iteration 4 found nothing further requiring change and
approved.

---
_Adopted via /adopt-pr workflow. Original contributor commits preserved._

PR #289, opened 2026-08-08
fix(gastown): stop the deacon paging CRITICAL for a city with no Dolt data plane

AJBcoding:
## Problem

`mol-deacon-patrol`'s dolt-health step evaluates its threshold table unconditionally, and the first row escalates **CRITICAL to the mayor** on `server.reachable == false`.

A city whose bead ledger lives on another backend (MySQL, Postgres) has no managed Dolt runtime at all — there is no data plane to report on, and there never will be. On such a city `gc dolt health --json` reports `applicable: false` with a reason, and the remaining fields are zero-valued placeholders: `server.reachable` is `false` because **no Dolt server exists**, not because one is down. Evaluating the table on that payload pages an unactionable CRITICAL every patrol cycle.

## Change

One step description in `gastown/formulas/mol-deacon-patrol.toml`: check `applicable` before the threshold table, and when it is false, log `Dolt health: not applicable (<reason>)` and move on — no escalation, no dog nudge. The `applicable` field is added to the documented payload shape, and the Step 2 heading now states the precondition.

## Backward compatible — safe to land in either order

An older `gc dolt health` omits the field entirely. `jq` yields `null`, `null == false` is false, so a payload without `applicable` is evaluated exactly as it is today. This can land before or after the producer half.

## Pairs with the gascity-side fix

Tracked as gas-e05. Before that fix, `gc dolt health` on a non-Dolt city exited **78 with no payload at all** (it dies while sourcing `runtime.sh`, which resolves the port at source time), so this step could not distinguish "Dolt is not in use here" from "healthy" from "the check itself is broken" — commit bloat, stale backups, zombie servers and orphan databases went unreported with no signal that monitoring had gone blind. The producer now emits a well-formed report and exits 0; this is the consumer half that reads it.

Verified on the MySQL city that filed the bug: `gc dolt health --json` went from exit 78 with no payload to exit 0 with `applicable: false`, `reason: "city beads backend is mysql; no managed Dolt runtime to probe"`.

### DIFF
--- a/gastown/formulas/mol-deacon-patrol.toml
+++ b/gastown/formulas/mol-deacon-patrol.toml
@@ -393,6 +393,7 @@ gc dolt health --json
 ```
 
 Parse the JSON output (HealthReport schema):
+- `applicable`: false when this city has no Dolt data plane at all
 - `server`: running, reachable, pid, port, latency_ms
 - `databases[]`: name, commits, open_beads
 - `backups`: dolt_freshness, dolt_age_seconds, dolt_stale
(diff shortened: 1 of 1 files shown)


AJBcoding:
@quad341 Ready for review when you have a moment — flagging here since I can't request reviewers (outside contributor).

This one is a **safety fix ahead of its own producer**: it is backward compatible by construction (a payload without `applicable` is evaluated exactly as today), so it can land in any order. Landing it *first* is strictly safer — it closes the false-page window before the gascity-side change starts emitting the new field.

Verified against the MySQL city that filed the underlying bug: `gc dolt health --json` went from exit 78 with no payload at all to exit 0 with `applicable: false` and a reason naming the backend.

Unrelated to #287, which is still open — happy to take review on either in whichever order suits you.

PR #290, opened 2026-08-08
fix(gastown): gate the polecat handoff on the branch having commits

atbrace:
## Problem: a polecat can stamp a refinery handoff having produced nothing

`mol-polecat-work`'s `submit-and-exit` step has a branch-**shape** gate, a clean-state check, a push gate and a push-verification. **None of them detect an empty branch.** On a branch created at the base tip and never committed to, every gate passes:

| gate | on an empty branch |
|---|---|
| branch-shape gate | passes — the branch *is* named `polecat/<bead-id>` |
| final clean-state verification | passes — no commits means a clean tree |
| `git push origin HEAD` | **succeeds**, and creates the ref at the base tip |
| push verification (`git ls-remote` == local HEAD) | passes — the ref exists and matches |

So the bead is reassigned to the refinery and moves forward through the pipeline **looking like success**.

Observed in the wild: `metadata.branch=polecat/<id>` existed on origin pointing at a *former main tip*, with `git log origin/main..origin/polecat/<id>` empty and an empty diff. The only thing that caught it was the refinery's own no-work-on-branch check — one gate slot later, after the handoff had already been stamped and a merge-queue slot consumed. **Without that downstream check it merges as a no-op and the bead closes with the work undone.**

This is worse than a worker that simply idles: an idle worker is *visible* (the bead sits unclaimed and a coordinator notices), whereas a phantom handoff advances the bead into a state that reads as completion.

Reproduced minimally:

```
$ git checkout -b polecat/sys-xxxxx origin/main
$ git rev-list --count origin/main..HEAD   # 0
$ git status --porcelain | wc -l           # 0
$ git push origin HEAD                     # succeeds, ref created
$ git ls-remote origin refs/heads/polecat/sys-xxxxx   # == local HEAD
```

## Change

Add a **branch-content gate** before the push: require at least one commit ahead of `origin/{{base_branch}}`. Zero commits is not a handoff.

The gate deliberately does **not** merely abort. "No code change is warranted" is a legitimate outcome for an investigation bead, and if the only modeled exit is a handoff, an agent that reaches that conclusion has nowhere honest to go — which is a plausible way a phantom handoff gets produced in the first place. So the gate returns the bead open and unassigned with `halt_reason=no_commits`, clears `gc.routed_to`, and requires notes stating what was investigated and why no change is warranted. The conclusion gets *reported* instead of laundered into a merge.

The `verify` step already told the agent that `git log origin/<base>..HEAD` "must show your commits" — that was advice in prose. This makes it a gate that fails closed.

## Validation

- TOML parses; all 11 bash blocks in `submit-and-exit` pass `bash -n`.
- Predicate exercised against a scratch repo reproducing the phantom preconditions above: `rev-list --count` is `0` for the empty branch and `1` once a commit exists, while every pre-existing gate passes in both cases.

### DIFF
--- a/gastown/formulas/mol-polecat-work.toml
+++ b/gastown/formulas/mol-polecat-work.toml
@@ -364,6 +364,49 @@ BODY
 This is a belt-and-suspenders check — self-review should have caught this,
 but we never push with untracked work left behind.
 
+**2b. Branch-content gate (fails closed — non-negotiable).**
+
+The branch-shape gate proves the branch is *named* right; it does not prove
+it *contains* anything. A branch created at the base tip and never committed
+to passes every other gate here: the name matches, the tree is clean, and
+`git push` of zero commits succeeds and creates the ref, so push verification
+sees origin == HEAD. The handoff is then stamped and the bead moves forward
+looking like success — a phantom handoff. Only the refinery's own
+no-work-on-branch check catches it, one gate slot later; without that check
+it merges as a no-op and the bead closes with the work undone. An idle worker
+is visible, this is not.
+
+```bash
+COMMITS_AHEAD=$(git rev-list --count "origin/{{base_branch}}..HEAD")
+if [ "$COMMITS_AHEAD" -eq 0 ]; then
+    echo "BRANCH CONTENT GATE FAILED"
+    echo "  branch:  $EXPECTED_BRANCH"
+    echo "  commits ahead of origin/{{base_branch}}: 0"
+    echo ""
+    echo "There is nothing to hand off. Do NOT push and do NOT reassign to the"
+    echo "refinery — an empty branch is a phantom handoff, not a completion."
+    echo ""
+    echo "If you HAVE done work, it is uncommitted or on another branch: go back,"
+    echo "commit it on $EXPECTED_BRANCH, and re-run submit-and-exit."
+    echo ""
+    echo "If the honest answer is that no code change is warranted (an"
+    echo "investigation that concluded 'no', a bead already satisfied by main),"
+    echo "that is a legitimate outcome — but it is reported, not handed off."
+    gc bd update "$WORK_BEAD_ID" \
+      --status=open --assignee="" \
+      --set-metadata gc.routed_to="" \
+      --set-metadata halt_reason=no_commits \
+      --notes "$(cat <<'BODY'
+Halted at submit: branch has zero commits ahead of base. No handoff stamped.
+<State what you investigated and why no code change is warranted, or what
+blocked you from producing one. A coordinator reads this to decide next steps.>
+BODY
+)"
+    gc runtime drain-ack
+    exit 1
+fi
+```
+
 **3. Push gate — FAIL CLOSED, then push:**
 
 Absent metadata is not consent. A rail expressed only in prose is enforced by
--- a/gastown/formulas/mol-polecat-work.toml
+++ b/gastown/formulas/mol-polecat-work.toml
@@ -376,36 +376,79 @@ no-work-on-branch check catches it, one gate slot later; without that check
 it merges as a no-op and the bead closes with the work undone. An idle worker
 is visible, this is not.
 
+The predicate is deliberately commit-count-only, measured against the
+`origin/{{base_branch}}` tip: it catches the phantom handoff — a branch with
+nothing on it — at the cheapest point, before any push. It is weaker than the
+refinery's canonical `branch_has_real_change`, which also requires a non-empty
+`git diff` against the merge-base and therefore additionally catches a net-zero
+branch (commit plus revert). That split is intentional: the refinery stays the
+backstop for net-zero work. Do not "fix" only one side of the pair.
+
+`git rev-list` failing is not the same answer as "zero commits", and neither is
+a licence to hand off: a count that is not a plain number halts under its own
+`halt_reason=content_gate_error` rather than falling through to the push. That
+is why the count is matched with `case` instead of `[ "$COMMITS_AHEAD" -eq 0 ]`,
+which evaluates false — and so fails open — on an empty value.
+
 ```bash
-COMMITS_AHEAD=$(git rev-list --count "origin/{{base_branch}}..HEAD")
-if [ "$COMMITS_AHEAD" -eq 0 ]; then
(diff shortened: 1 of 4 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277291908

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277291977

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277309566

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277316590

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @atbrace! This PR closes a real phantom-handoff hole in the polecat
submit-and-exit path: it refuses to stamp a handoff when the branch has no commits ahead of base,
so a session that did no work can no longer hand off as if it had. That matters because the handoff
stamp is what downstream dispatchers and the witness trust — a phantom one is believed, and the work
silently evaporates. You also put the gate in exactly the right place: after the clean-state check
has committed any leftovers, so it cannot fire on work that merely had not been committed yet, and
before the push that would create the ref.

### Original PR Review
Decision: approve (the first pass requested changes; everything it asked for is now in).

Specific gaps fixed:
- **The count could fail open.** `git rev-list --count` was read with a plain `[ "$COMMITS_AHEAD"
  -eq 0 ]`, so when the count could not be measured at all — an empty or non-numeric result from a
  `git` error — the comparison fell through and the gate waved the handoff past. A tool error is a
  suspect, not a licence to merge. The final patch drops the `-eq` test entirely and validates with
  `case`, distinguishing "branch is empty" (`no_commits`) from "could not measure" 
  (`content_gate_error`) and failing closed on both. The review verified this by extracting the gate
  block and *executing* it against stubbed `git`/`gc` on four inputs: a `rev-list` error, zero, a
  healthy count of three, and a malformed `"  12x "`. The healthy path performs zero `gc` calls; the
  other three all halt.
- **The halt lost its escalation.** The new gate ran ahead of the refinery's existing
  `halt_false_completion`, which nudges mayor and witness — but the new halt only parked the bead
  open, unassigned and unrouted, which no dispatcher tier serves and witness orphan recovery skips.
  Net effect: the gate caught the bug and then dropped the human-routed visibility the old path had.
  The final patch nudges mayor and witness after the park, using target forms byte-identical to the
  refinery sibling it front-runs, and reports a failed nudge rather than swallowing it.
- **The gate was deletable with zero CI signal.** Nothing in the suite would have gone red if the
  whole fence were removed. The final patch pins it from both directions: command-literal pins in
  the build-workflow contract dict, plus a `tomllib`-anchored positional validator that asserts the
  gate sits between the clean-state commit and the push, along with its halt markers. The pins check
  order and adjacency, not just presence — a catch-all `case` arm slipped between the two existing
  arms would silently kill `no_commits` while leaving the gate looking alive, and that mutation now
  reds the suite.

Review findings addressed:
- All of the above, plus three non-gating items, landed in maintainer fixup `8c86fa3`:
  - *Error handling / resilience (major, high confidence):* fail-open count → `case`-validated,
    fails closed with a distinct reason per cause.
  - *Change impact / blast radius (major, high confidence):* escalation restored on the halt path.
  - *Test evidence quality (gating requirement, high confidence):* the contract pins and the
    positional validator, so deleting the gate now reddens CI.
  - *Readability / documented choice (minor):* the gate header now records that the count-only
    predicate is deliberate, names the refinery as the net-zero backstop, and warns against fixing
    only one side of the pair.
  - *Debuggability (minor):* the remedial operator text no longer points at foreclosed recoveries —
    the unreachable "it is uncommitted" arm and the dead "re-run submit-and-exit" advice are gone.
  - *Test evidence quality (minor, mutation-measured):* a second hardening round on the validator
    itself — the `case`-dispatch pin now fixes arm *order* rather than just arm presence, and the
    halt's `exit 1` is pinned for the first time, re-anchored to the escalation loop's `done`. Worth
    naming because without that `exit 1` the fence would end rc=0 and the agent would walk on to the
    push it had just refused.

- `mc-bsubz`: *maintainability, minor* — the validator's halt-path window is cut from the count to
  the push, which overshoots the gate's own `fi` and spans the neighbouring `auto_push_false` halt.
  The specific instance this caused is already closed in this PR (the `exit 1` pin is anchored to
  something unique to this gate), but bounding the window structurally would remove the whole class,
  so a *future* bare fragment cannot be satisfied vacuously by the sibling's copy.
- `mc-0jc9c`: *operability, nit* — `halt_reason` now has three producer values and still zero
  in-repo readers. Both review passes adjudicated this acceptable, because the mayor/witness nudges
  are the live observability channel and carry the reason in their message text; the metadata is an
  audit breadcrumb. Worth a reader or a documented write-only note eventually.

One review suggestion was deliberately **not** taken: a smaller one-fragment version of the
validator hardening. It closes the two sabotage-shaped mutations but leaves the most plausible
*accidental* one green while marking the finding resolved, so the fuller two-entry form was landed
instead.

### Maintainer Changes
`8c86fa3 fix(gastown): make the branch-content gate fail closed, escalate, and be testable` —
4 files changed, 320 insertions(+), 20 deletions(-). On top of your commit the full adopted range is
4 files changed, 345 insertions(+), 2 deletions(-). Your commit is preserved with its original
authorship; the maintainer fixup also carried the change set through a conflict with the
`submit-and-exit` rewrite that landed in #273 (the halt note switched to `--append-notes`, now an
invariant with its own test) and a rebase onto current `main`.

### Final Review Status
Ready for the merge queue: final head `8c86fa3` has passing required GitHub checks — `Check`,
`Analyze (go)`, `Analyze (python)`, and `CodeQL` all green — and the merge-ready metadata is in
place. The formula text, the contract pins, and the validator are one unit and land together, which
this PR's final head does. I am marking this PR `status/merge-ready` so the merge queue can pick it
up.

CI: https://github.com/gastownhall/gascity-packs/actions/runs/34315680464

### Review Iterations
2 review passes performed: the first requested changes on the two majors above plus test evidence;
the second confirmed each fix by execution and mutation testing rather than by inspection, and
approved.

---

PR #291, opened 2026-08-09
fix(gastown): make refinery find-work survive its own claim

AJBcoding:
## The bug

`mol-refinery-patrol`'s `find-work` step queries `--status=open`, so it can never see the work it just claimed. Claiming a bead moves it to `in_progress`, which the filter excludes. The bead strands in `in_progress` with its branch metadata intact and nothing merged, while the step **truthfully reports an empty queue** and the patrol idles.

This is a leak with a rate, not a one-off: one stranded bead per claim, in every rig running this pack. Measured in a production rig, the stranded count went **1 -> 2 in 30 minutes**, with `--status=open` returning nothing the whole time and the branches provably unmerged rather than lost.

It is also silent in the worst way — the refinery looks healthy-idle and the beads look green, so nothing surfaces the backlog.

## The fix

```diff
- --assignee=$GC_AGENT --status=open
+ --assignee=$GC_AGENT --status=open,in_progress
```

The query is **already narrowed by `--assignee=$GC_AGENT`**, so this widens it to the refinery's *own* claims only — it never picks up another agent's in-flight work.

Re-finding your own claim is the intended resume path, not a duplicate merge: `merge-push`'s merge-state gate already closes a bead as `already_merged` if a previous pass landed it. To be precise about what that gate implements: it is **ancestry-based** (plus a `fork_sha` commit count), not patch-equivalence. A branch that landed externally as a squash would miss it, fall through to the diff-based false-completion guard, and halt for escalation rather than re-merging — fail-safe, but not the seamless resume an earlier draft of this line implied.

The comma-separated form is deliberate — repeating `-s/--status` silently overwrites the previous value instead of unioning them, which `gc bd list --help` calls out explicitly.

## Why the prompt changed too

`gastown/agents/refinery/prompt.template.md`'s "Find assigned work" quick-reference row taught the **same** blind query. Fixing only the formula would have left the cheat sheet teaching the bug.

## Verification

- Reproduced against a live store: `--status=open` returns empty while `--status=in_progress` returns the stranded work beads, same assignee, branch metadata intact.
- The fixed query returns a bead that `--status=open` could not see.
- **Wisp safety:** the shipped query **retains** the `--has-metadata-key=branch` filter — what follows is an extra robustness experiment, not a description of the change. Even with that filter dropped, the widened query still matched only real work beads and no patrol wisp. The verified belt is tier exclusion: wisp roots are `ephemeral=true` and this query does not pass `--include-infra`, so widening the status cannot make the refinery adopt its own wisp. (Review correction: `--has-metadata-key=branch` is not an independent second belt — no bead in the measured store carries a `branch` key at all, so that probe cannot discriminate on its own. Tier exclusion is the guard that actually holds.)
- `gastown/tests/test_gastown_pack_assets.sh` passes; `tests/test_no_bare_bd_commands.py` passes; formula TOML parses with all 9 steps intact.

## Relationship to #287

Independent — no file overlap (#287 touches the `agent.toml`s and `worktree-setup.sh`). Mergeable in either order.

### DIFF
--- a/gastown/agents/refinery/prompt.template.md
+++ b/gastown/agents/refinery/prompt.template.md
@@ -325,7 +325,7 @@ alert the witness, not `gc mail send`.
 |------------|----------------|
 | Pour next wisp | `gc bd mol wisp mol-refinery-patrol --root-only --var target_branch={{ .DefaultBranch }} --var rig_name={{ .RigName }} --var binding_prefix={{ .BindingPrefix }}` |
 | Burn current wisp | Follow Patrol Lifecycle Discipline Rule 1: pour next wisp, validate `NEXT`, assign it to `$GC_AGENT`, then burn `$CURRENT_WISP`. Never run a standalone burn. |
-| Find assigned work | `gc bd list ${GC_RIG:+--rig="$GC_RIG"} --assignee="$GC_AGENT" --status=open` |
+| Find assigned work | `gc bd list ${GC_RIG:+--rig="$GC_RIG"} --assignee="$GC_AGENT" --status=open,in_progress` — both statuses, comma-separated: a bead you already claimed is in_progress, and a query filtered to open alone can never see it again |
 | Snapshot event position | `gc events --seq` |
 | Wait for assignment | `gc events --watch --type=bead.updated --after=$SEQ` |
 | Read work metadata | `gc bd show $WORK --json \| jq '.[0].metadata'` |
--- a/gastown/formulas/mol-refinery-patrol.toml
+++ b/gastown/formulas/mol-refinery-patrol.toml
@@ -169,9 +169,9 @@ needs = ["check-inbox"]
 description = """
 **Config: event_timeout = {{event_timeout}}**
 
-Search for work beads assigned to you with branch metadata:
+Search for work beads assigned to you — unclaimed, or already claimed by you:
 ```bash
-WORK=$(gc bd list ${GC_RIG:+--rig="$GC_RIG"} --assignee=$GC_AGENT --status=open \
+WORK=$(gc bd list ${GC_RIG:+--rig="$GC_RIG"} --assignee=$GC_AGENT --status=open,in_progress \
   --exclude-type=epic --has-metadata-key=branch --limit=1 --json | jq -r '.[0].id // empty')
 ```
 
@@ -180,6 +180,26 @@ database. Without it, `gc bd list` routes to the HQ database and never
 sees rig-scoped work beads — the `assignee` field alone does not route
 by namespace. For HQ-only refineries `$GC_RIG` is empty and the flag
 expands to nothing.
+
+`--status=open,in_progress` is what makes this query survive its own claim.
+Claiming a bead moves it to `in_progress`, so a filter of `open` alone can
+never see the work this patrol just took: the bead strands in `in_progress`
+with its branch metadata intact and nothing merged, while this step truthfully
+reports an empty queue. It is a leak with a rate — one stranded bead per
+claim, in every rig running this pack (measured in a production rig: the
+stranded count went 1 -> 2 in 30 minutes, with `--status=open` returning
+nothing the whole time and the branches provably unmerged rather than lost).
+
+Both statuses stay narrowed by `--assignee=$GC_AGENT`, so this widens the
+query to *your own* claims only — it never picks up another agent's in-flight
+work. Wisp roots are `ephemeral=true` and this query does not pass
+`--include-infra`, so widening the status cannot make the refinery adopt its
+own patrol wisp. Use the comma-separated form: repeating `-s/--status`
+silently overwrites the previous value instead of unioning them.
+
+Re-finding your own claim is the intended resume path, not a duplicate: a bead
+you claimed but did not land is unfinished work, and the merge-state gate in
+merge-push already closes it as `already_merged` if a previous pass landed it.
 
 If found: read the bead metadata to confirm it has a branch:
 ```bash
--- a/scripts/gascity_pack_inference_gate.py
+++ b/scripts/gascity_pack_inference_gate.py
@@ -117,7 +117,7 @@ GASTOWN_BUILD_WORKFLOW_CONTRACTS = {
         "gc runtime drain-ack",
     ),
     "mol-refinery-patrol": (
-        "gc bd list ${GC_RIG:+--rig=\"$GC_RIG\"} --assignee=$GC_AGENT --status=open",
+        "gc bd list ${GC_RIG:+--rig=\"$GC_RIG\"} --assignee=$GC_AGENT --status=open,in_progress",
         "git rebase origin/$TARGET",
         "{{typecheck_command}}",
         "{{lint_command}}",


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277290523

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277309405

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277316512

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @AJBcoding! This PR fixes a silent, self-inflicted work leak in the refinery patrol: `mol-refinery-patrol`'s `find-work` step queried `--status=open`, so the moment it claimed a bead — moving it to `in_progress` — it could never see that bead again. The work stranded with its branch metadata intact and nothing merged, while the step *truthfully* reported an empty queue and the patrol idled. That combination is the worst kind of bug to catch in production: the refinery looks healthy-idle and the beads look green, so nothing surfaces the backlog. Your reproduction against a live store, watching the stranded count go 1 → 2 in thirty minutes with `--status=open` returning nothing the whole time, is what made it undeniable.

Two things about the change itself are worth calling out, because they are the parts a hurried fix would have gotten wrong. You widened the query to the refinery's *own* claims only — it is already narrowed by `--assignee=$GC_AGENT`, so it never picks up another agent's in-flight work. And you used the comma-separated `--status=open,in_progress` form deliberately, because repeating `-s/--status` silently overwrites rather than unions, which `gc bd list --help` warns about. You also fixed the "Find assigned work" cheat-sheet row in the refinery prompt template in the same change, on the reasoning that fixing only the formula would have left the cheat sheet teaching the bug. That instinct was right, and it turned out to be the more interesting half of the review.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- **The enforcement pin could not tell the fix from the bug.** `scripts/gascity_pack_inference_gate.py` pins the refinery find-work command as a contract fragment, and the pinned literal still ended `--status=open` — which is a *prefix* of your corrected text. The validator is presence-only (`if fragment not in text`), so the pin was satisfied by both the fixed and the buggy form. Concretely: the gate would have stayed green if someone reverted your query line. This was the one gating item, and it is the reason there were two review passes rather than one.
- **The fix for it had to be the full literal, not the short one.** The obvious remedy — pinning just the token `--status=open,in_progress` — was measured and found *dead*: green on a revert **and** green on outright deletion of the query line, because your own new prose in `mol-refinery-patrol.toml` also contains that token, keeping the pin satisfied. The full command literal occurs exactly once in the formula and is genuinely discriminating. That distinction was measured rather than assumed, and it is why the fixup looks fussier than it needs to.

Review findings addressed:
- The pin repin landed as `3f51d0e` (`scripts/gascity_pack_inference_gate.py`, +1/−1). Verified end-to-end through the real validator and the shipped suite, on the current rebased tree: with the formula's query line reverted to `--status=open`, `tests/test_gascity_pack_inference_gate.py` now fails on exactly `test_validate_gastown_orchestration_contract_accepts_current_pack` (1 failed / 62 passed); restored, 63 passed. So a future revert of your fix is now something CI will actually notice. The full focused set — `tests/test_gascity_pack_inference_gate.py` plus `tests/test_no_bare_bd_commands.py` — is 65 passed on the final head.
- Two overstatements in the PR description were corrected in place (description only, no code): the merge-state gate is ancestry-based rather than patch-equivalence-based, and the wisp-safety bullet read as though the `--has-metadata-key=branch` filter had been removed when the shipped query retains it. Worth knowing for the second one: the review found that filter is not actually an independent second safety belt — no bead in the measured store carries a `branch` key at all, so the probe cannot discriminate on its own. The guard that genuinely holds is tier exclusion, since the query does not pass `--include-infra` and wisp roots are `ephemeral=true`. Your shipped prose in the formula already framed it that way; only the PR description overreached.

- `mc-g1wvr` (minor): the refinery cheat-sheet row at `gastown/agents/refinery/prompt.template.md:323` is the one widened surface with no guard. The filter now lives in three places — formula query, gate pin, and that hand-copied row — and only the first two are coupled, so a revert or a reword of the row alone passes every suite. A validated joint prompt+formula assertion for `test_gastown_pack_assets.sh` is recorded on the bead, along with two measured traps to avoid when writing it.
- `mc-nmaos` (deferred, tracked): both refinery-queue monitors are still `--status=open`-only — the witness `check-refinery` step and the deacon `queue-starvation-check`. A bead stranded in `in_progress` makes both counts zero, so it reads as a healthy idle queue, which is exactly why this leak had to be found by hand. This was deliberately kept out of your PR: the review's own scope instruction was not to touch those two formulas here, and both were verified byte-identical to base. Your change strictly shrinks the invisible population; widening the monitors is its own PR.
- `mc-bjfbp` (minor): the rebase step's `git checkout -b temp` assumes a virgin worktree. Since a stranded claim usually died mid-rebase — precisely when a leftover `temp` branch survives — the resume path your fix enables can land on that collision, and the step's only failure guidance is the conflicts path. Pre-existing mechanism; your change alters which beads reach it. `checkout -B` plus one line distinguishing "branch missing on origin" from a real conflict.
- `mc-5602z` (minor): the deacon formula pours unconditionally while its own prompt teaches reconcile-to-one, so a pour/burn race can accumulate open wisps. Pre-existing, surfaced by the review sweep rather than caused by this PR.
- `mc-nx71t` (nit): the new pin freezes the lone unquoted `--assignee=$GC_AGENT` among the formula's `gc bd list` queries, so a future quoting cleanup will trip the contract gate. No action needed — just change the query line and the pin in one commit if that line is ever touched. Firing on any edit to a pinned line is intended tripwire behavior.
- Also raised and confirmed already tracked, so no new bead: seven sibling own-wisp queries omit `--include-infra`, which is issue #252.

One class-wide limitation is worth stating plainly so the new pin is not over-trusted: these fragment pins are presence-only. Commenting out the pinned line, or adding a second buggy query beside a correct one, still passes. That is true of every fragment pin in this repo and is not something a one-line repin should try to solve.

### Maintainer Changes
One maintainer commit, `3f51d0e` "fix(gastown): repin the refinery find-work contract to the widened status filter" — `scripts/gascity_pack_inference_gate.py`, 1 file changed, 1 insertion(+), 1 deletion(-). The branch was also rebased onto current `main` (`b43abe6`) with your authorship preserved and the patch content unchanged; the final range is 3 files changed, 24 insertions(+), 4 deletions(-). The PR description edits noted above were made directly on this PR and touch no code.

### Final Review Status
Ready for the merge queue: final head `3f51d0e` has passing required GitHub checks — Check, Analyze (go), Analyze (python), and CodeQL are all green on this exact head — and the merge-ready metadata is in place. The rebase onto current `main` moved through commits touching all three files this PR edits, so the focused suite and the pin's revert-detection were re-measured on the final tree rather than inherited from the earlier review. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/gascity-packs/actions/runs/34312752790

### Review Iterations
2 review passes. The first found the pin defect and requested exactly one change; the second confirmed the repin was the discriminating form, in the right location, with nothing else touched, and approved.

---

PR #293, opened 2026-08-09
fix(slack-full): reply-current inherits thread_ts from the latest threaded inbound

aphexcx:
## Problem

`gc slack reply-current` — "reply to the latest inbound" — drops thread context. When the latest inbound routed to a session is a **thread reply**, the reply posts at **channel level** instead of into the thread. Reproduced live twice on 2026-08-09: the session's inbound reminder carried `Message id: 1786291407.960839 (in thread 1786250478.963679)`, the agent answered with `reply-current --conversation-id Cxxx --body-file f`, and the human found the reply in channel history while the thread contained only their own messages.

The thread root was already recorded end to end: the adapter stamps Slack's `thread_ts` into the inbound's `reply_to_message_id`, and gc surfaces it on the transcript record as `ReplyToMessageID`. `reply-current` just never read it.

## Fix

- New `slack_intake_common.find_latest_inbound_thread_for_session` returns `(provider_message_id, thread_root, conversation)` via the existing event→transcript chain. `find_latest_inbound_message_id_for_session` becomes a thin wrapper, so `react`/`upload` callers are untouched.
- `reply-current` now inherits the inbound's thread root **by default** — including when `--conversation-id` names the same conversation explicitly (the repro shape). Guards:
  - an explicit target naming a *different* conversation never borrows the thread anchor (a foreign `thread_ts` would strand the reply);
  - an unthreaded inbound keeps the channel-level reply, byte-for-byte;
  - a failed lookup degrades to channel level with a stderr warning, so `--via adapter` survives a gc outage.
- `--thread-current` now anchors at the thread **root** when the latest inbound was itself a thread reply. Slack threads hang off the parent `ts`; anchoring at a child message strands the reply (Slack API docs: "Avoid using a reply's ts value; use its parent instead").
- New `--no-thread` flag forces a channel-level post; combining it with `--reply-to`/`--thread-current` is rejected.
- The company-context path (`_maybe_company_reply`) is untouched — it already does its own thread anchoring.

## Tests

- gp-i62 repro: threaded inbound + explicit `--conversation-id` → reply carries the inbound's thread root.
- Unthreaded inbound → no `reply_to_message_id` (channel level unchanged).
- Cross-conversation guard, `--no-thread` opt-out + flag-combination rejection, `--reply-to` bypasses the lookup entirely.
- `--thread-current`: root anchoring on threaded inbound, own-ts on plain inbound.
- Lookup-failure degradation (warning + channel level, exit 0).
- Mock-backed event→transcript chain: `GcMock` transcript entries now carry `ReplyToMessageID`, and `find_latest_inbound_thread_for_session` extracts root vs `""`.

Full slack-full suite on this branch: **394 passed**.

https://claude.ai/code/session_01EMTmu76XFdBZRmc3MzZdcB

### DIFF
--- a/slack-full/CHANGELOG.md
+++ b/slack-full/CHANGELOG.md
@@ -8,6 +8,24 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- `gc slack reply-current` now inherits the thread from the latest
+  inbound (gp-i62): a thread-reply inbound's transcript entry carries
+  the Slack `thread_ts` in `ReplyToMessageID`, and the reply anchors
+  there by default — including when `--conversation-id` names the same
+  conversation explicitly, which previously always posted at channel
+  level (live burns 2026-08-09, #gastown + #fundraising). An unthreaded
+  inbound keeps the channel-level reply; an explicit target naming a
+  *different* conversation never borrows the inbound's thread anchor;
+  a failed lookup degrades to channel level with a stderr warning
+  (`--via adapter` must survive a gc outage). New `--no-thread` flag
+  forces a channel-level post. `--thread-current` now anchors at the
+  thread ROOT when the latest inbound was itself a thread reply —
+  Slack threads hang off the parent ts, so anchoring at the child
+  stranded the reply. Retires the fleet-memory workaround of routing
+  threaded replies through `publish-to-channel --thread-ts`.
+
 ### Added
 
 - Accidental-mrkdwn guard on the send path (gp-o42): `gc slack
--- a/slack-full/CHANGELOG.md
+++ b/slack-full/CHANGELOG.md
@@ -25,6 +25,28 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   Slack threads hang off the parent ts, so anchoring at the child
   stranded the reply. Retires the fleet-memory workaround of routing
   threaded replies through `publish-to-channel --thread-ts`.
+  `reply-current` now also reports the anchor it used —
+  `reply_to_message_id` in the result JSON, plus an `inheriting thread
+  <ts> from inbound <mid>` line on stderr when inheritance fires — so a
+  reply that lands in an unexpected thread can be traced to the inbound
+  that donated the anchor. The inherited anchor is the conversation's
+  newest inbound, not provably the message being answered (nothing on
+  the wire identifies which inbound woke the session); the limitation
+  and its `--reply-to` escape hatch are documented in
+  `commands/reply-current/help.md`.
+- `gc slack upload --thread-current` anchors at the thread ROOT too,
+  matching the parity its help text promises with `gc slack
+  reply-current`. It consumed a message-id-only lookup that dropped the
+  thread root, so a file uploaded in answer to a thread reply anchored
+  at the child ts and stranded outside the thread.
+- A *wedged* gc (connection accepted, then stalled) no longer crashes
+  callers whose degrade guards catch `GCAPIError`: the HTTP helper
+  wraps `TimeoutError` — raised bare out of `resp.read()`, an `OSError`
+  rather than a `URLError` — so a stalled lookup degrades like a
+  refused one instead of killing the reply with a traceback. The wrap
+  covers the stall's siblings too: a gc killed mid-response sends an
+  RST and `resp.read()` raises `ConnectionResetError`, so every
+  transport `OSError` now reaches callers in that same currency.
 
 ### Added
 
--- a/slack-full/README.md
+++ b/slack-full/README.md
@@ -65,7 +65,9 @@ Implemented:
 - [x] `gc slack reply-current` — reply to the latest Slack event in the
       current session, by default through gc's `/extmsg/outbound` so
       transcript recording + peer fanout fire (`--via adapter` keeps the
-      old direct-to-adapter path for diagnostics)
+      old direct-to-adapter path for diagnostics). A thread-reply
+      inbound is answered in its thread by default (`--no-thread`
+      forces a channel-level post)
 - [x] `gc slack publish` — publish to a session's saved binding (target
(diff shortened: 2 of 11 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277289869

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277309240

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277316425

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @aphexcx! This PR makes `gc slack reply-current --thread-current` inherit `thread_ts` from the latest threaded inbound, so an agent answering a threaded message replies *inside* that thread instead of dropping a channel-level message next to it. That is the difference between a conversation that stays readable and one where every agent reply detaches from the thread it belongs to — a real day-to-day papercut for anyone running Slack-backed sessions.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- **The inherited anchor was not provably the message being answered.** `find_latest_inbound_thread_for_session` picks the conversation's *newest* threaded inbound, so in a busy shared channel a threaded message from someone else landing in between re-aims the reply into their thread. Binding the pick tighter turned out to need a message identity the wire does not carry — `InboundEventPayload` is `{Provider, ConversationID, Actor, TargetSession, TargetAgent}` with no message id, and nothing hands the command the inbound it is answering. Rather than imply the problem away, the approximation is kept deliberately and made discoverable: the docstring no longer claims "routed to this session", an explicit "Anchor caveat" paragraph names the interloper mechanism, and `commands/reply-current/help.md` states the failure shape plus both escape hatches (`--reply-to`, `--no-thread`).
- **`upload --thread-current` and `reply-current --thread-current` had silently diverged.** `upload` consumed the message-id-only lookup, which drops the thread root, so a file answering a thread *reply* anchored at the child ts and stranded itself outside the thread — while `upload/help.md`'s "same logic as `gc slack reply-current`" claim, true at base, went false with this change. `upload` now anchors at `thread_root or mid`, the identical expression `reply-current` uses, and its argparse help and help.md say so explicitly instead of by reference.

Review findings addressed (all in maintainer commit `c1e3dac`):
- **Error handling / resilience (minor, then nit):** a wedged `gc` that sends headers and then stalls raises a bare `TimeoutError` out of `resp.read()`. It is an `OSError`, not a `URLError`, so it escaped the `except GCAPIError` degrade guard and killed a reply on a path that made no `gc` call at all before inheritance existed. `_request` now wraps it. A follow-up pass found the sibling shape — a `gc` killed mid-response sends an RST and raises `ConnectionResetError`, which also escaped — so a separate trailing `except OSError` arm was added with its own accurate message, kept last on purpose since `URLError` and `TimeoutError` are both `OSError` subclasses and would otherwise lose their specific messages to it.
- **Debuggability (nit):** the success-path re-anchor was silent, which is exactly what would make a misfire of the approximation above unreproducible. Inheritance now logs `inheriting thread <ts> from inbound <mid>` to stderr, and the result JSON reports `reply_to_message_id` on every path.
- **Docs / architectural consistency (nits):** `commands/reply-current/help.md` now covers `--thread-current` and both anchoring arms; the module docstring and the `--thread-current` argparse help no longer open with the retired "routed to this session" phrasing that the rest of the file had just disavowed; and `upload/help.md` states the newest-inbound caveat directly and names `--thread-ts` as upload's exact-anchor escape, instead of carrying the caveat one hop away through the parity sentence.

Every behavior change is pinned by a discriminating test: reverting the timeout wrap, the new `OSError` arm, the upload anchoring, the stderr line, or the JSON field each fails exactly the test that pins it and nothing else. The existing upload tuple-unwrap regression test was repointed at the thread lookup rather than dropped. Full `slack-full` suite: 448 passed.

- `mc-f7u67` — *[performance / state safety, nit]* collapse the double inbound event scan on the no-`--conversation-id` path. Two non-atomic scans mean a new inbound landing between them can make the conversation resolved by the first disagree with the thread picked by the second. Deliberately deferred: it was originally paired with an anchor-binding mechanism that turned out to be unavailable in-pack.
- `mc-5vv4g` — *[correctness, minor]* extend the conversation-match guard to the `--thread-current` arm, which can currently anchor into a conversation other than the one the session last heard from. Pre-existing at base and unchanged in kind by this PR — the review confirmed the anchor within a foreign conversation only moved from the child ts to the thread root. Batched with the out-of-scope half of the phrasing fix above: `slack_chat_react.py` and `commands/react/help.md` still carry the retired wording, and `react` is not among this PR's files.
- `mc-ey2jw` — *[architecture, accepted limitation]* the interloper hazard itself. The durable fix is a gc-core wire change — carry the inbound message id on `extmsg.inbound` and select by equality instead of by recency. The pack's own mock already models a message id, so the pack side is ready for it. Out of this repo's reach, which is why this PR documents the limitation instead.
- `mc-zf3ku` — *[docs, nit]* `template-fragments/slack-v0.template.md:57-67` describes a `--turn-ref`/`posted_ts`/`parked` `reply-current` surface that does not exist in this pack. Verified identical at base and not among this PR's changed files, so out of range here.

### Maintainer Changes
`c1e3dac fix(slack-full): bound the thread-anchor contract and give upload the same anchoring` — 8 files changed, 301 insertions(+), 40 deletions(-). Your commit is preserved unamended as the parent; the maintainer fixup sits on top of it.

### Final Review Status
Ready for the merge queue: final head `c1e3dac` has passing required GitHub checks, and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/gascity-packs/actions/runs/34312319263

### Review Iterations
2 review passes performed. Iteration 1 raised two gating majors — the unbound thread anchor and the `--thread-current` divergence between `reply-current` and `upload` — plus four non-gating recommendations. Iteration 2 verified both gates fixed (including confirming your original commit was not amended, so iteration 1's review of it still stands), took the three remaining nits, and returned approve.

---

PR #294, opened 2026-08-10
fix(gastown): pass --include-infra on every patrol wisp lookup

edwarddavison:
## The bug

Every gastown patrol agent (refinery, witness, deacon) resolves its own wisp with a lookup that cannot see it:

```bash
CURRENT_WISP=$(gc bd list --assignee="$GC_AGENT" --status=in_progress --type=molecule --limit=1 --json | jq -r '.[0].id // empty')
```

Wisps fall in the **infrastructure visibility class**, which `gc bd list` hides by default. Reproduced against a live rig:

```
$ gc bd show gci-wisp-agu --json | jq '.[0] | {id,issue_type,status,assignee,ephemeral}'
{ "id": "gci-wisp-agu", "issue_type": "molecule", "status": "in_progress",
  "assignee": "gas-city/gastown.refinery", "ephemeral": true }

$ gc bd list --assignee="gas-city/gastown.refinery" --status=in_progress --type=molecule --limit=5 --json
[]

$ gc bd list --assignee="gas-city/gastown.refinery" --status=in_progress --type=molecule --include-infra --limit=5 --json
gci-wisp-agu
```

`--include-infra` output matches `gc bd query "ephemeral=true AND …"` exactly, across `in_progress` and `open`, for both the refinery and the witness.

## Why it is silent, and therefore dangerous

The lookup runs inside `$(...)` without pipefail, so `set -e` never sees it. `CURRENT_WISP` comes back empty, the guarded `"could not resolve current wisp; not burning"` branch fires **exactly as designed**, and the stale wisp is left assigned and orphaned instead of burned. One more leaks every cycle.

Measured on that rig before this change, the refinery held three wisps (one `in_progress`, two `open`) — including one a prior session had explicitly confirmed as its clean anchor and which still went unburned. At least the sixth documented recurrence since 2026-06-12; sb-app logged 13+, and two leaked June gate-sweep wisps sat invisible to a default `gc bd list` for eight weeks of triage. It is only ever caught when an operator runs the query by hand instead of trusting the formula.

## Why the previous fix did not fix it

An earlier pass swapped these calls from `--type=wisp` to `--type=molecule`. That fixed a real bug — `wisp` is not a valid gc bd issue type, so the command errored — but not this one, and it made the resolver *look* correct: `--type=molecule` is valid, it matches the wisp's actual type, and it still returns `[]` because the visibility filter runs first.

## The fix

All 15 resolver sites now pass `--include-infra`, plus the predecessor lookup in the refinery `patrol-summary` prose:

```bash
CURRENT_WISP=$(gc bd list --assignee="$GC_AGENT" --status=in_progress --type=molecule --include-infra --limit=1 --json | jq -r '.[0].id // empty')
```

The work-bead queries in the same files are deliberately untouched — only lookups filtering `--type=molecule` are wisp resolvers.

## Test

New `gastown/tests/test_wisp_resolvers.sh`, picked up by the existing "Run Gastown shell tests" CI step. It fails if any patrol prompt or formula resolves a wisp without `--include-infra`, or regresses to `--type=wisp`. It matches invocations only (a resolver line runs `gc bd list` and ends in `--json`), never the prose that warns against those forms.

Negative proof: dropping the flag from a single site fails the test; restoring it passes.

## Verification

- `gastown/tests/*.sh` — 5/5 pass (including the new one)
- `python3 validate_registry.py --require-git` — ok
- `go test .` (root embed module) — ok
- `tests/test_no_bare_bd_commands.py` — both tests pass
- All `gastown/formulas/*.toml` parse under `tomllib`
- End-to-end: the rendered `next-iteration` snippet, executed under `set -euo pipefail` with `GC_AGENT` set to a live refinery, resolves `CURRENT_WISP=[gci-wisp-agu]` — the wisp the unflagged form could not see

## Downstream note

`gascity`'s `examples/gastown/gastown_test.go` asserts the **old** strings verbatim (`TestWitnessStartupAndNoIdleReconcileWisps`, and the no-idle order check in the prompt-render table). Those assertions read pack content through the `gascity-packs` Go module, so they will not break until someone bumps the pin — but they must be updated in the same pin bump. Flagging so it is not discovered as a surprise CI failure.

Refs: gci-853, gci-byp

https://claude.ai/code/session_01UxBWmhHRfH5vfqeuhqvu5q

### DIFF
--- a/gastown/agents/deacon/prompt.template.md
+++ b/gastown/agents/deacon/prompt.template.md
@@ -89,9 +89,9 @@ If `next-iteration` already ran, do not pour again; run `gc hook`.
 ```bash
 CURRENT_WISP=${GC_BEAD_ID:-}
 if [ -z "$CURRENT_WISP" ]; then
-  CURRENT_WISP=$(gc bd list --assignee="$GC_AGENT" --status=in_progress --type=molecule --limit=1 --json | jq -r '.[0].id // empty')
+  CURRENT_WISP=$(gc bd list --assignee="$GC_AGENT" --status=in_progress --type=molecule --include-infra --limit=1 --json | jq -r '.[0].id // empty')
 fi
-ASSIGNED_WISP=$(gc bd list --assignee="$GC_AGENT" --status=open --type=molecule --limit=1 --json | jq -r '.[0].id // empty')
+ASSIGNED_WISP=$(gc bd list --assignee="$GC_AGENT" --status=open --type=molecule --include-infra --limit=1 --json | jq -r '.[0].id // empty')
 if [ -n "$CURRENT_WISP" ] && [ -z "$ASSIGNED_WISP" ]; then
   NEXT=$(gc bd mol wisp mol-deacon-patrol --root-only --var binding_prefix={{ .BindingPrefix }} --json | jq -r '.new_epic_id // empty')
   if [ -z "$NEXT" ]; then
--- a/gastown/agents/refinery/prompt.template.md
+++ b/gastown/agents/refinery/prompt.template.md
@@ -64,12 +64,20 @@ Two rules govern your inter-wisp behavior. Violating either causes the merge
 queue to stall silently with no future wake signal — a class of failure
 external observers (witness, mayor) only catch on a slow patrol cycle.
 
+**Every wisp lookup needs `--include-infra`.** Wisps fall in the
+infrastructure visibility class, which `gc bd list` hides by default, so
+without the flag the resolver returns `[]` for a wisp `gc bd show` proves
+exists — in_progress, assigned to you. It fails inside `$(...)` without
+tripping `set -e`, so `CURRENT_WISP` reads empty, the guarded "could not
+resolve current wisp; not burning" branch fires exactly as designed, and the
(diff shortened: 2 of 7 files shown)


edwarddavison:
Closing as a duplicate — this was authored independently before I found the existing work, and it adds nothing #266 and #278 don't already cover.

- **#266** (open) — the three `agents/*/prompt.template.md` resolvers
- **#278** (open) — the three `gastown/formulas/*.toml` resolvers, the `patrol-summary` prose reference, and a regression test

Together those are the same 15 sites and the same `--include-infra` fix, tracked by #252. #264 is the same bug via `gc bd query "ephemeral=true …"`; the id sets are provably identical, so whichever lands second needs only a trivial rebase.

My independent reproduction corroborates #278's evidence on a different rig (gas-city): `--type=molecule` → `[]`; `--type=molecule --include-infra` → the live `in_progress` wisp, matching `gc bd query "ephemeral=true AND …"` exactly across `in_progress` and `open` for both refinery and witness. The refinery was holding three wisps at the time (one `in_progress`, two `open`).

One finding worth carrying over, posted on #278 as well: `gascity`'s `examples/gastown/gastown_test.go` will fail at pin-bump time, not now.

PR #295, opened 2026-08-10
fix(slack-full): build adapter on macOS — use x/sys/unix.Openat instead of Linux-only syscall.Openat

jeffora:
## The problem

`slack-full/adapter` does not compile on macOS:

```
./confined_open.go:46:24: undefined: syscall.Openat
```

`syscall.Openat` is Linux-only in the Go stdlib. `openBeneath` — the symlink-safe component walk that confines inbound file uploads — has therefore never built on darwin. Confirmed in both directions on `darwin/arm64`, Go 1.26.1: `go build ./...` fails, `GOOS=linux go vet ./...` is clean.

Scope is this module alone. `slack-full/cli`, `slack-channel/adapter` and `slack-mini/adapter` all build fine on darwin.

## The fix

Swap the stdlib `syscall` calls for `golang.org/x/sys/unix`, which implements `Openat` on darwin and linux both.

The algorithm is untouched — same component-by-component walk, same `O_NOFOLLOW` on every component, same `O_DIRECTORY` on interior components, same errno. **Existing tests pass unmodified**, including `TestReadConfinedFileRejectsSymlink`'s `errors.Is(err, syscall.ELOOP)` assertion: `x/sys/unix` returns `syscall.Errno` values, so error identity is preserved.

## The trade — and please tell us if you'd rather have it the other way

This adds the module's **first dependency and first `go.sum`**. `go.mod` had no `require` block and there was no `go.sum` file at all, so the zero-dependency property was literal, not aspirational. That is the strongest argument against this patch and we want to state it plainly rather than gesture at it.

The alternative is stdlib **`os.Root`** (Go 1.24+; the `go` directive here is already 1.25.9). It keeps the module dependency-free and deletes the hand-rolled walk entirely. We did not choose it because it enforces a weaker property than this function promises. Measured on `darwin/arm64`, Go 1.26.1:

| fixture | `os.Root` | `O_NOFOLLOW` walk |
|---|---|---|
| symlink, **absolute** target | rejected — "path escapes from parent" | rejected — ELOOP |
| symlink, **relative** in-root target | **followed, file read** | rejected — ELOOP |
| control, real file | opened | opened |

`os.Root` guarantees *cannot escape the root*. This function's doc comment promises something stronger — *any symlink anywhere beneath root is a hard failure* — and `gc-cby.10` / `gpk-1ta4` filed it specifically to close a TOCTOU window. We also tried `root.OpenFile(rel, os.O_RDONLY|syscall.O_NOFOLLOW, 0)` to recover the strict behaviour: it does not work, `os.Root` does not honour that flag and still follows the relative in-root symlink.

**Worth knowing regardless of which patch you prefer:** the current test suite cannot see that difference. `TestReadConfinedFileRejectsSymlink` builds its symlink with `os.Symlink(filepath.Join(dir, "target.txt"), link)` — an *absolute* target, which `os.Root` rejects for escaping rather than for being a symlink. Change that one line to `os.Symlink("target.txt", link)` and an `os.Root` implementation opens the file and returns its contents while the current one rejects it. If you do prefer `os.Root`, that test needs a relative-target case added, and the doc comment needs rewriting — it currently asserts a property that is contingent on `Openat`.

Happy to swap this PR to the `os.Root` version if you'd rather hold the empty `go.mod`; it is a preference only you can hold.

## Testing

- `go build ./...` on darwin: **passes** (previously failed to compile)
- `GOOS=linux go vet ./...`: clean
- `go test ./...`: **entire suite passes**, no test modified

One caveat on running the suite locally on macOS: three upload-confinement tests fail under the default `TMPDIR` for a **separate, pre-existing** reason unrelated to this patch. `confinedUploadPath` (`main.go` ~1800-1819) compares an `EvalSymlinks`-resolved `realPath` against an *unresolved* root, and on macOS `/tmp` and `/var` are symlinks into `/private`, so any root beneath them fails containment. The failures all report `is outside root` and never reach `openBeneath`. With `TMPDIR` set to a canonical path the whole suite is green. We have deliberately left that out of scope here — happy to file it separately, or fold it in if you'd prefer (the fix is to resolve the root before comparing).

### DIFF
--- a/slack-full/adapter/confined_open.go
+++ b/slack-full/adapter/confined_open.go
@@ -5,7 +5,8 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
-	"syscall"
+
+	"golang.org/x/sys/unix"
 )
 
 // openBeneath opens rel (a Clean, root-relative path containing no
@@ -14,8 +15,19 @@ import (
 // with O_NOFOLLOW set. The parent fd pins the verified directory inode,
 // so a directory swapped for a symlink mid-walk cannot redirect the
 // traversal — the userspace equivalent of openat2(RESOLVE_BENEATH),
-// which stdlib syscall does not expose (and the adapter's
-// zero-dependency go.mod keeps us from importing x/sys for).
+// which stdlib syscall does not expose.
+//
+// The syscalls come from x/sys/unix rather than stdlib syscall because
+// syscall.Openat is Linux-only: on darwin the package builds but the
+// symbol is undefined, so this file did not compile on macOS at all.
+// x/sys/unix implements Openat on both, which keeps one implementation
+// and one security contract across platforms.
+//
+// stdlib os.Root is deliberately NOT used here. It enforces "cannot
+// escape the root", which is weaker than what this function promises:
+// os.Root FOLLOWS a symlink whose target stays inside the root, and it
+// does not honour O_NOFOLLOW passed through OpenFile. Switching to it
+// would silently narrow the guarantee below.
 //
 // O_NOFOLLOW applies to every component, not just the leaf, so any
 // symlink anywhere beneath root is a hard failure (ELOOP / ENOTDIR).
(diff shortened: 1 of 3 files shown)


jeffora:
Closing as a duplicate of #269, which predates this by six days and is strictly broader — same `x/sys/unix.Openat` fix, same reasoning about `os.Root` silently dropping the `O_NOFOLLOW` guarantee, plus it fixes the three `/var` test failures and adds `TestOpenBeneathRefusesSymlinkComponents`.

My apologies for the duplicate — I should have swept open PRs before writing the patch rather than after.

The measurement that motivated the same `os.Root` conclusion here has been moved to #269 as a comment, since that's where it's useful: a table showing that a *relative* in-root symlink is followed by `os.Root` but rejected by the `O_NOFOLLOW` walk, and the fixture line in `TestReadConfinedFileRejectsSymlink` that makes the difference invisible to the current suite.

#269 is the one to review.

PR #297, opened 2026-08-10
feat(slack-channel): Socket Mode transport; fix reply-current --conversation-id

nonathaj:
Two commits. The first ports the Socket Mode transport from #283 (slack-mini) to Tier 2; the second fixes a reply-path bug that is independent of it.

Both are running live against a real workspace on a network with no inbound ingress — see Testing.

## 1. Socket Mode transport

Same problem as #283: the adapter only took inbound over the Events API, which needs a public HTTPS Request URL, so the pack could not run anywhere that cannot accept inbound connections.

`SLACK_APP_TOKEN` (an `xapp-` app-level token with `connections:write`) selects it. When set, `LISTEN_PUBLIC` is never bound and `SLACK_SIGNING_SECRET` is not required; when unset, the HTTP transport behaves exactly as before, signature verification included.

Envelopes are handed to `routeEvent` — the same place the HTTP handler hands off — so channel bindings, handle aliases, the `app_mention` fallback, and per-session identity are shared verbatim rather than reimplemented.

Design carried over from #283: envelopes acked before routing; routing off the read loop so a slow gc cannot stall acks; supervised reconnect backing off 1s→30s; a 30s keepalive ping as the liveness check (a read deadline is wrong here — Slack's keepalives are pings answered inside the library, so silence is the healthy state); and an overlap on Slack's advance disconnect warning so the replacement is dialled while the old connection drains.

Two Tier 2 specifics:

- **Workspace filtering matters more here.** Tier 2 stamps `SLACK_WORKSPACE_ID` as the account id *and* keys channel bindings on it, so a stray workspace's event could otherwise match a binding that is not its own. Events with a mismatched `team_id` are dropped.
- **Interactivity is not carried.** Tier 2 ships interactivity disabled and `/slack/interactions` is an opt-in extra that still needs a public URL, so `interactive` envelopes are dropped with an explanatory log line. Documented in the README's transport table.

Adds `github.com/coder/websocket` v1.8.15 (no dependencies of its own), matching #283. The same offer stands: happy to hand-roll RFC 6455 on stdlib instead if maintainers prefer the adapters stay dependency-free — though the library dials via `net/http`, so `HTTPS_PROXY`/`NO_PROXY` are honoured, which matters on exactly the networks this targets.

## 2. reply-current now accepts --conversation-id

gc's injected reply instruction tells an agent to run:

```
gc slack-channel reply-current --conversation-id <id> --body-file <path>
```

Tier 2 did not accept `--conversation-id`, so the command gc advertises failed. `slack-full` and `discord` both take the flag — Tier 2 was the outlier.

Found the hard way: a mayor agent in a live city hit it, worked out the correct invocation itself, and filed the divergence — after four sessions had concluded Slack was simply down. (Related gc-side issue: gastownhall/gascity#5166, where the injected command's namespace is derived from the provider name. Not addressed here.)

The flag is validated rather than ignored. `reply-current` targets the session's current reply target (its latest inbound, or its single channel binding before any inbound), so a supplied id is redundant by construction — but a mismatch means the caller believes it is answering a conversation the session no longer owns, and silently posting into the current target would put the reply in the wrong channel. A mismatch is 409 naming both ids, and posts nothing.

## Testing

`go vet` + `go test -race` green across the whole Tier 2 suite. New tests: the transport's envelope loop, ack-before-route, drain-on-disconnect, idle-connection regression, keepalive closing a wedged connection, a real-WebSocket round trip via `websocket.Accept`, a reconnect-after-disconnect, and both `--conversation-id` paths (matching accepted; mismatched refused without posting).

Live, gc-supervised, on a network with no inbound ingress:

- `slack-channel` service reaches `ready`, Socket Mode connects, registers through gc's reverse proxy.
- A plain channel message with no `@`-mention is delivered to the bound session — Tier 2's defining behaviour, over the socket.
- One connection held **over three hours with zero reconnects**. Before the keepalive fix (see #283) the same setup churned a reconnect every 90 seconds.
- HTTP mode re-checked for regressions: still binds `:8775`, still answers `/healthz`.

## Note for maintainers

The README lists `GC_API_BASE_URL` among the vars "injected by gc when the adapter runs as a `proxy_process` service". It is not injected, and neither are `GC_CITY_NAME` or `GC_CITY_PATH` — a supervised adapter cannot discover its own city or API endpoint, and defaults to `:9443`. I have left the docs alone here to keep this PR to the transport plus the flag fix, but happy to correct that in either this PR or a follow-up. Raised on gastownhall/gascity#5166 as well.

### DIFF
--- a/slack-channel/CHANGELOG.md
+++ b/slack-channel/CHANGELOG.md
@@ -5,6 +5,45 @@ All notable changes to slack-channel are documented in this file.
 The format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+
+- **Socket Mode transport** (`adapter/socketmode.go`), an ingress-free
+  alternative to the public Events API listener: the adapter opens an
+  outbound WebSocket to Slack instead of receiving webhooks, so the pack
+  runs on networks that cannot accept inbound connections (corporate
+  firewalls, laptops, anywhere a tunnel is blocked). Ported from the
+  slack-mini (Tier 1) transport of the same name.
+  - Selected by `SLACK_APP_TOKEN` (an `xapp-…` app-level token with the
+    `connections:write` scope). When set, `LISTEN_PUBLIC` is never bound
+    and `SLACK_SIGNING_SECRET` is no longer required — Socket Mode has no
+    request signatures, so the trust boundary is the app-token-authenticated
+    connection the adapter itself opens.
+  - Envelopes are handed to `routeEvent`, so channel bindings, handle
+    aliases, the `app_mention` fallback, and per-session identity behave
+    identically to the HTTP path.
+  - Supervised reconnect with capped exponential backoff (1s → 30s), a 30s
+    keepalive ping as the liveness check, and an overlap on Slack's advance
+    disconnect warning so in-flight envelopes are still acked.
+  - Events from a workspace other than `SLACK_WORKSPACE_ID` are dropped
+    rather than filed under the wrong account or matched against another
+    workspace's channel bindings.
+  - Dialled through Go's default HTTP client, so `HTTPS_PROXY`/`NO_PROXY`
+    are honoured.
+  - **Interactivity is not carried over Socket Mode.** Tier 2 ships
+    interactivity disabled and `/slack/interactions` is an opt-in extra
+    that still needs a public URL; `interactive` envelopes are dropped with
+    an explanatory log line.
+- `manifest/app-socket.json`, the Socket Mode variant of the app manifest
+  (identical to `manifest/app.json` but with `socket_mode_enabled: true`).
+
+### Changed
+
+- The adapter now depends on `github.com/coder/websocket` v1.8.15, which
+  has no dependencies of its own. The HTTP transport is unchanged and still
+  verifies request signatures exactly as before.
+
 ## [0.1.0] — Tier 2
 
 Initial release. slack-channel is Tier 2 of the Slack pack family — the
--- a/slack-channel/CHANGELOG.md
+++ b/slack-channel/CHANGELOG.md
@@ -25,10 +25,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
     identically to the HTTP path.
   - Supervised reconnect with capped exponential backoff (1s → 30s), a 30s
     keepalive ping as the liveness check, and an overlap on Slack's advance
-    disconnect warning so in-flight envelopes are still acked.
-  - Events from a workspace other than `SLACK_WORKSPACE_ID` are dropped
-    rather than filed under the wrong account or matched against another
-    workspace's channel bindings.
+    disconnect warning so in-flight envelopes are still acked. The backoff
+    floor is restored only once a connection has lasted 30s, so a server
+    that accepts a connection and drops it immediately — or the
+    connection-limit churn two adapters sharing one app token produce —
+    escalates the delay instead of redialling at the floor forever.
   - Dialled through Go's default HTTP client, so `HTTPS_PROXY`/`NO_PROXY`
     are honoured.
   - **Interactivity is not carried over Socket Mode.** Tier 2 ships
@@ -41,8 +42,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - The adapter now depends on `github.com/coder/websocket` v1.8.15, which
-  has no dependencies of its own. The HTTP transport is unchanged and still
-  verifies request signatures exactly as before.
+  has no dependencies of its own. The HTTP transport still verifies request
+  signatures exactly as before.
+- **Events carrying another workspace's `team_id` are now dropped on both
+  transports**, at the shared `routeEvent` funnel. Previously the HTTP path
+  accepted them: a request signature proves only that Slack sent the event
+  for this app, not that it came from this workspace, so an app installed
+  in a second workspace had that workspace's messages filed under
+  `SLACK_WORKSPACE_ID` and matched against this workspace's channel
+  bindings. An absent `team_id` is still accepted — Slack sends one on
+  every `event_callback`.
 
 ## [0.1.0] — Tier 2
 
--- a/slack-channel/README.md
+++ b/slack-channel/README.md
@@ -45,15 +45,42 @@ flags.
 
 ## Install
 
+## Choose a transport
+
+Inbound messages reach the adapter one of two ways. Pick one before you
+create the app — the choice is baked into the app manifest.
+
+| | **Socket Mode** | **HTTP (Events API)** |
+| --- | --- | --- |
+| Public ingress | none | public HTTPS URL required |
+| How events arrive | outbound WebSocket the adapter opens to Slack | Slack POSTs to your Request URL |
+| Credentials | app-level token (`xapp-…`) | signing secret |
+| Manifest | [`manifest/app-socket.json`](./manifest/app-socket.json) | [`manifest/app.json`](./manifest/app.json) |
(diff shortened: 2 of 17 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277289658

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277309150

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277316339

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @nonathaj! This PR gives `slack-channel` a Socket Mode transport so Tier 2 can run on networks with no inbound ingress, and fixes `reply-current` to accept the `--conversation-id` flag that gc's own injected reply instruction passes. Both matter: the first removes a public HTTPS Request URL as a hard requirement for this tier, and the second closes a gap where the command gc advertises to agents did not exist here — a divergence that reads as "Slack is down" from the inside, which is exactly how it surfaced.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- **The foreign-workspace guard only covered one transport.** Dropping events with a mismatched `team_id` was implemented inside the Socket Mode path, but an HTTP request signature proves only that Slack sent an event *for this app* — not which workspace it came *from*. An app installed in a second workspace would have had that workspace's messages filed under `SLACK_WORKSPACE_ID` and matched against this workspace's channel bindings on the HTTP path. The guard moved to `routeEvent`, the funnel both transports share (mirroring slack-mini's placement), so one check now covers both. An absent `team_id` is still allowed through, as before.
- **Reconnect pacing reset too eagerly.** A connection that Slack accepted and then dropped immediately reset the backoff floor, so the connection-limit churn that two adapters sharing one app token produce would redial at network speed indefinitely. The floor is now restored only once a connection has survived `socketMinLifetime` (30s), and the disconnect-warning arm is paced too — a normal-lifetime connection still resets to the floor, which fits inside the warning's lead and preserves the drain overlap you designed for.
- **The file header overstated the mirror.** `socketmode.go` described itself as mirroring slack-mini's transport with one "only substantive difference" while three things in fact differed. Replaced with an accurate account of what this tier does differently.
- **Understandability budgets.** `runSocketMode` and `pumpSocket` sat above the pack's cognitive-complexity budget, and the new workspace guard had pushed `routeEvent` above the band it occupied at base. All three were brought back by pure extraction into named arms — `superviseConnection`, `paceRedial`, `pumpReadErr`, `startDrain`, `ackEnvelope`, `admitEvent` — with no behavior change and no existing test rewritten to fit. `routeEvent` now sits below where it started.

Review findings addressed:
- **[Security Surface] major (claude, codex)** — foreign-workspace guard moved to the shared inbound funnel, pinned by `TestRouteEventDropsForeignWorkspace` with both a socket arm and an HTTP arm plus a local-team control — `fc8dbfd`.
- **[Error Handling & Resilience] major (claude)** — backoff floor gated on `socketMinLifetime`, and the warned-replacement dial paced — `fc8dbfd`.
- **[Debuggability & Operability] minor (gemini)** — `socketmode.go` header comment corrected — `fc8dbfd`.
- **[Change Impact / Blast Radius] major (claude, codex)** — complexity extraction for `runSocketMode` (29→5), `pumpSocket` (27→14) and `routeEvent` (back under its base band), verified behavior-preserving by an arm-by-arm mutation matrix with a green control — `fc8dbfd`.
- **[Test Evidence Quality] minor (claude)** — keepalive wiring pinned by `TestPumpSocketTearsDownWedgedConnection`: removing the `go keepaliveSocket(...)` call now fails exactly that test while the neighbouring keepalive tests stay green, so the ping is demonstrably a liveness check rather than decoration — `fc8dbfd`.
- **[Contract & Interface Fidelity] minor (claude)** — `reply-current` help text corrected: it had said the target is "always the session's latest inbound", which the no-inbound single-binding fallback makes untrue. The same entry now also spells out that this flag *asserts* the already-resolved target rather than retargeting the way slack-full and discord do, since it shares their name but not their semantics — `fc8dbfd`. The PR description carried the same wording and has been updated to match.
- **[Release Safety] nit (gemini, codex)** — smaller cleanups taken along the way: a README heading level, a Security-posture entry documenting the cross-workspace drop on both transports, and `SLACK_APP_TOKEN` added to `pack.toml`'s inherited-secrets comment — `fc8dbfd`.

- `mc-lr8kp`: [Release Safety] nit, high confidence (gemini) — `CHANGELOG.md` describes `manifest/app-socket.json` as differing from `manifest/app.json` in one field, but the pair actually differs in two (`display_information.description` as well as `settings.socket_mode_enabled`). The manifests are correct and match slack-mini's own pair convention; only the sentence is wrong. Not addressed in this PR — deliberately left for the next time the CHANGELOG is touched, rather than moving the reviewed head for a wording nit.
- `mc-nji4e`: [Architectural Consistency] minor, high confidence (gemini) — nothing on the slack-mini side records that `slack-channel` now carries a copy of this transport, so a connection-lifecycle fix in mini has no marker pointing at the second copy. Not addressed in this PR: the remedy edits `slack-mini/`, outside this PR's diff, and #407 already rewrites that region and is the natural vehicle.
- `mc-w9ez8`: [Resource Lifecycle & Cleanup] minor, high confidence (claude) — `/healthz` is static, so under Socket Mode the process can report healthy while the WebSocket is down, reconnecting, or wedged. Not addressed in this PR: it is a tier-family gap rather than something this PR introduced, and the supervised reconnect and keepalive added here are what make that state knowable in the first place.
- `mc-lwfly`: [Test Evidence Quality] nit, high confidence (all three review lanes) — `TestPumpSocketTearsDownWedgedConnection` (5s bound) and `TestRunSocketModePacesWarnedReconnect` (40ms floor) assert wall-clock behavior and could flake on a pathologically slow runner. Not addressed in this PR: both pins are load-bearing, and the `socketTimings` seam this PR introduces is already the mitigation; no change is warranted until a flake is actually observed.

Two things you raised yourself were checked and settled rather than acted on. The `GC_API_BASE_URL` / `GC_CITY_NAME` / `GC_CITY_PATH` README inaccuracy is real, but it is pre-existing debt rather than anything this PR introduced, and it is tracked upstream at gastownhall/gascity#5166 alongside the injected-command namespace issue — keeping this PR to the transport plus the flag fix was the right call. And your offer to hand-roll RFC 6455 on stdlib was considered and declined, for the reason you gave yourself: `github.com/coder/websocket` dials via `net/http`, so `HTTPS_PROXY`/`NO_PROXY` are honoured, which matters on precisely the networks this transport targets, and it matches what #283 already took for slack-mini. Separately, at-most-once delivery under the socket path was examined and left alone — it is inherited from the HTTP path by design, not a regression introduced here.

### Maintainer Changes
One maintainer fixup commit, `fc8dbfd` "fix(slack-channel): apply review fixes for the Socket Mode transport" — 9 files changed, 627 insertions(+), 133 deletions(-). Your two original commits are preserved beneath it, unmodified.

### Final Review Status
Ready for the merge queue: final head `fc8dbfd` has passing required GitHub checks, and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

The branch was rebased onto current `main` (`5312520`) before the push; the rebase was content-preserving (identical patch-id) and the base move touches none of the files this PR changes. First-hand on the final head: `gofmt -l` clean, `go vet` clean, and the full `slack-channel/adapter` suite green under `-race` (97 tests), plus `gc lint slack-channel`, registry validation, and the pack test suites.

CI: https://github.com/gastownhall/gascity-packs/actions/runs/34344549476

### Review Iterations
3 review passes, each with independent Claude, Codex, and Gemini lanes plus a synthesis. The first raised the three substantive issues above and requested changes; the second confirmed those fixes and asked for the complexity work; the third re-measured every claim first-hand and ran a mutation matrix — each extracted arm was probed to confirm a live test fails when its logic is broken — and approved with nothing gating left.

---

PR #298, opened 2026-08-11
fix(gastown): bound status-line refreshes by process group, not by timeout(1)

AJBcoding:
## Summary

`gastown/assets/scripts/status-line.sh` runs on every tmux status refresh for
every agent, so an unbounded query is paid on the refresh schedule by the whole
fleet. `run_bounded` had two ways to apply no bound at all:

1. **`timeout(1)` is not portable.** It ships with GNU coreutils and is absent
   on macOS (no `gtimeout` either), so `command -v timeout || run unbounded`
   failed open and applied *no* bound. Measured on a macOS fleet host: a 6s `gc`
   stretched a single status refresh to 12-13s, on every render, for every
   agent.
2. **A bound on the direct child does not cover what that child backgrounds.**
   The grandchild also inherits stdout, so a caller reading through a command
   substitution blocks long after the bound expired while the grandchild orphans
   to init. This one bites on Linux too, where `timeout` *is* present — the
   bound applies to `gc` and the backgrounded work sails past it.

`run_bounded` now runs the command in its own process group (`set -m`) writing
to a file rather than to the caller's pipe, polls for the bound, and signals the
whole group on expiry — sweeping it after a clean exit too, since a command can
return promptly and still leave work behind. The bound is `GC_STATUSLINE_BOUND`
seconds (default 2, preserving the previous `timeout 2s` behaviour), and scratch
is removed on any ordinary exit so refreshes do not litter `TMPDIR`.

## Relationship to #207

#207 (`hotfix/statusline-single-flight`, open since 2026-07-15) touches the same
function and I do not want to talk over it. Two honest differences:

- #207 fixes failure mode 1 by shelling out to `perl`, which leaves failure mode
  2 in place: `perl` forks and `TERM`s the direct child, so work that child
  backgrounded still survives and still holds the inherited stdout.
- #207 also carries a **single-flight cache lock**, which this PR does not
  touch. That is a genuinely separate improvement and is worth landing on its
  own merits.

These two changes are complementary, not competing. If maintainers prefer, I am
happy to rebase this on top of #207 (the process-group `run_bounded` replaces
the `perl` branch cleanly and the lock is untouched), or to hold this until #207
lands. Please just say which order you want.

## Safety

- No new dependency: the helper uses only POSIX shell job control plus tools the
  script already relied on. It no longer needs `timeout(1)` at all.
- Signals are scoped to the process group of the command this script itself
  started (`kill -- -$pid`); nothing else on the host is targeted.
- Bound output goes to a per-invocation `mktemp -d` scratch removed on `EXIT`,
  `HUP`, `INT`, `TERM`.
- Rendering is unchanged: counts, cache semantics, city scoping, and the
  always-exit-0 contract all still hold (existing cases cover this).
- Sub-second polling where `sleep 0.1` is supported, so the common fast case is
  not padded out to a whole second on every render.

## Test plan

Two cases added to the existing `gastown/tests/test_gastown_theme_scripts.sh`
(already the home of the status-line coverage, and already run by CI's
`for test_script in gastown/tests/test_*.sh` step):

- `test_status_line_bounds_a_hanging_query_without_timeout_binary` — runs the
  script against a hanging `gc` on a PATH sandbox that deliberately omits
  `timeout(1)`, reproducing the macOS case deterministically on any host. It
  carries a **positive control** asserting the sandbox really does hide
  `timeout`, so the case cannot quietly degrade into "timeout did the bounding".
- `test_status_line_reaps_work_the_query_leaves_behind` — a `gc` that backgrounds
  a long-lived child and returns promptly; asserts both that nothing survives the
  bound and that the refresh is not wedged by the inherited stdout. This one
  fails on the old helper regardless of platform.

Both were confirmed to fail before the fix and pass after (not merely to pass):

```
# new cases vs the pre-fix status-line.sh
test_status_line_bounds_a_hanging_query_without_timeout_binary -> FAIL: hanging query was never bounded where timeout(1) is absent
test_status_line_reaps_work_the_query_leaves_behind            -> FAIL: refresh never completed — caller wedged by a grandchild holding stdout

# whole suite vs the patched status-line.sh
$ bash gastown/tests/test_gastown_theme_scripts.sh; echo $?
gastown theme script tests passed
0
```

Also run:

- `shellcheck -s sh gastown/assets/scripts/status-line.sh` — clean.
- `shellcheck gastown/tests/test_gastown_theme_scripts.sh` — clean.
- The other `gastown/tests/test_*.sh` scripts pass unchanged.
  (`test_witness_heartbeat_check.sh` fails on macOS both before and after this
  change — it uses GNU `date -d` — so it is untouched and unrelated here; it
  passes on CI's ubuntu runner.)

File mode stays `100755`, so the pack content hash keeps the perm the release
tooling derives for `.sh`.

This same change has been running in a production city since 2026-08-09.

### DIFF
--- a/gastown/assets/scripts/status-line.sh
+++ b/gastown/assets/scripts/status-line.sh
@@ -7,7 +7,7 @@
 # Counts are cached with a short TTL so tmux's frequent refresh does not query
 # Beads on every render across every agent. TTL override:
 # GC_STATUSLINE_TTL (seconds). Cache directory override:
-# GC_STATUSLINE_CACHE_DIR.
+# GC_STATUSLINE_CACHE_DIR. Bound override: GC_STATUSLINE_BOUND (seconds).
 
 agent="$1"
 city="${2:-${GC_CITY:-${GT_ROOT:-${GC_DIR:-}}}}"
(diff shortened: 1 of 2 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277289512

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277290157

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277309038

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277316321

PR #310, opened 2026-08-13
fix(checks): route molecule-member reads through gc ready, not a graph-blind collection query

julianknutsen:
## The bug

The three `gascity` review gates found their molecule members with a metadata collection query:

```sh
gc bd list --all --metadata-field "gc.root_bead_id=$ROOT" --json --limit=0 2>/dev/null
```

A metadata collection query carries **no bead id**, so on a city that relocates the graph class `bd` has nothing to route on and **refuses the read (exit 1)** rather than answering from the wrong store. Two of the three call sites also carried an explicit `|| printf '[]'`, and all three carried `2>/dev/null` — so the refusal became an **empty member set**, not an error.

The gate then never saw its verdict key and returned `iterate`, which is indistinguishable from a review that genuinely hadn't finished. The loop ran until Ralph exhausted its attempts.

| script | starved key |
|---|---|
| `design-review-approved.sh` | `design_review.verdict` |
| `gap-analysis-approved.sh` | `gap_analysis.verdict` |
| `implementation-review-approved.sh` | `code_review.verdict` |

## The fix

`gc ready` is the federating reader — city store, rig stores, and the relocated graph store, across both tiers. It takes exactly one `--status` and has no `--all`, so the member set is the union of one leg per status.

The four legs run **concurrently**. A check gate has a 10m budget and one `gc ready` costs ~17s on a loaded city; four sequential legs (~67s) inside a five-iteration poll loop would come uncomfortably close to that budget.

A leg that fails **names itself on stderr and fails the helper**, which propagates under `set -euo pipefail` to a nonzero script exit — the gate iterates instead of silently approving or silently starving. Silent starvation is the defect being fixed; the replacement must not reintroduce it.

## Verification

**Live**, against maintainer-city's relocated graph store (root `gcg-284771409692055`):

- **70 members**, all 70 carrying `gc.step_id`
- sqlite ground truth: **70** — exact match
- **17.27s** (vs ~67s for the sequential form)

**Control + negative**, with a stub `gc` on `PATH`:

| case | rc | members | stderr |
|---|---|---|---|
| control — all legs OK | 0 | 4 | *(none)* |
| negative — `closed` leg refuses | 1 | 3 | `gmol: gc ready failed for status: closed` |

Under `set -euo pipefail` the negative **propagates**: the caller aborts rather than continuing on the short set.

## Scope

Same fix already landed for the workflows-pack gates (`adopt-pr-ci-ready`, `adopt-pr-review-approved`, `code-review-approved`, `design-review-approved`) and for the gc-plan-pack copies (#309).

These three gates are **latent on maintainer-city today** — its graph store has no `design-review-fixes` continuation group and no `design_review`/`gap_analysis` verdict keys — so this closes the gap before the formula next runs rather than repairing a live stall.

The same three scripts also exist, still unfixed, in the detached-HEAD `build-methodology-packs` worktree (`f77fab4`), consumed by `paxel-city`, `paxel-city-cherry`, `maintainer-city-cherry` and `gc-methodology-smoke-city`. None of those cities is currently running; that worktree needs a rebase onto this fix.

### DIFF
--- a/gascity/assets/scripts/checks/design-review-approved.sh
+++ b/gascity/assets/scripts/checks/design-review-approved.sh
@@ -28,7 +28,11 @@ gmol() {   # root_id -> molecule-member JSON array
         echo "gmol: gc ready failed for status: $(tr '\n' ' ' <"$tmp/failed")" >&2
         rc=1
     fi
-    jq -s 'map(select(type=="array")) | add // [] | unique_by(.id)' "$tmp"/*.json || rc=1
+    # unique_by sorts by id, so the union comes back in bead-id order. The
+    # verdict extractors below take `| last`, which must mean "most recently
+    # updated" -- without this re-sort the gate picks a verdict by id hash and
+    # can sit on a stale `iterate` forever while a newer `done` is ignored.
+    jq -s 'map(select(type=="array")) | add // [] | unique_by(.id) | sort_by(.updated_at // "")' "$tmp"/*.json || rc=1
     rm -rf "$tmp"
     return "$rc"
 }
--- a/gascity/assets/scripts/checks/design-review-approved.sh
+++ b/gascity/assets/scripts/checks/design-review-approved.sh
@@ -3,6 +3,36 @@
 
 set -euo pipefail
 
+gmol() {   # root_id -> molecule-member JSON array
+    # `gc bd list --metadata-field` is a collection query carrying no bead id,
+    # so on a city that relocates the graph class bd has nothing to route on and
+    # refuses the read -- and the `2>/dev/null` below turned that refusal into an
+    # empty set, so this gate never saw design_review.verdict and looped until Ralph
+    # ran out of attempts.
+    #
+    # `gc ready` is the federating reader: city store, rig stores and the
+    # relocated graph store, across both tiers. It takes exactly one --status
+    # and has no --all, so the member set is the union of one leg per status.
+    # The four legs are independent reads, so run them concurrently: a check
+    # gate has a 10m budget and each `gc ready` costs ~17s on a loaded city.
+    # A leg that fails is reported on stderr and fails the function rather than
+    # contributing an empty set -- silent starvation is the bug being fixed.
+    local root="$1" tmp st rc=0
+    tmp="$(mktemp -d)" || return 1
+    for st in open in_progress blocked closed; do
+        { gc ready --metadata-field "gc.root_bead_id=$root" --status "$st" --limit 0 --json \
+            >"$tmp/$st.json" || printf '%s\n' "$st" >>"$tmp/failed"; } &
+    done
+    wait
+    if [ -s "$tmp/failed" ]; then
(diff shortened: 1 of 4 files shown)


The1nk:
Verified against `main` at `56016d5` (fresh clone, not a pack cache).

This converted the three `gascity/assets/scripts/checks/*.sh` gates. The same query
class is still present at five other call sites in this repo, and one of them carries
**both** halves of the defect described here — the graph-blind collection query *and*
the `2>/dev/null` that turns a refusal into an empty set rather than an error.

### Both halves, still on main

`gastown/agents/refinery/prompt.template.md:167`

```sh
ORPHANS=$(gc bd list ${GC_RIG:+--rig="$GC_RIG"} --metadata-field gc.routed_to="${GC_RIG:+$GC_RIG/}{{ .BindingPrefix }}refinery" --status=open --json 2>/dev/null \
  | jq -r '.[] | select(.metadata.branch != null) | .id')
```

The consequence here may be worse than the gate loop, because this scan *is* a
fallback. Its own comment (lines 161-165) says:

> Polecats sometimes die between commit and MERGE_READY mail (e.g. controller
> restart, host wake, claim race). Their branch ships but you never see the mail.
> Scan metadata for orphans before the normal patrol — these are real merge
> candidates that need rescuing.

A refused read yields zero orphans, which is indistinguishable from an inbox that
genuinely lost nothing. So the recovery path for lost merge mail reports success
while rescuing nothing, and the polecat's branch is never merged. Same shape as the
starved verdict key: silence that reads as good news.

### Same query class — molecule members by `gc.root_bead_id`, unconverted

- `compound-engineering/agents/ce-code-review-selector/prompt.template.md:144`
- `compound-engineering/assets/workflows/compound-code-review/{target}.conditional-review-gate.md:13`
- `superpowers/assets/workflows/superpowers-brainstorming/{target}.confirm-spec-approval.md:93` and `:98`

All four are `gc bd list --all --metadata-field "gc.root_bead_id=$CLAIMED_ROOT_BEAD_ID" ... --json`,
i.e. literally the query in this PR's description. None has `2>/dev/null`, so a
refusal should surface rather than empty out — but they are member reads, and the
`gc ready` argument in `gmol()` applies to them unchanged.

### Same collection-query shape, different key

- `gastown/formulas/mol-deacon-patrol.toml:344` —
  `gc bd list --status=in_progress --metadata-field gc.routed_to={{binding_prefix}}dog --json --limit=0`

### What we have and have not verified

We have **not** reproduced the refusal. We run `gc` v1.4.0 and do not relocate the
graph class, so for us these sites are latent, not live — we are reporting call-site
parity with a defect you have already diagnosed and fixed elsewhere, not a second
independent bug. Every path and line number above is cited from `main` at `56016d5`,
not from our pin.

The refinery one is the site we would prioritise, on the grounds that its failure mode
is silent non-recovery in a path that only runs when something has already gone wrong.

Happy to open a proper tracking issue for the remaining sites if that is more useful
than a comment on a merged PR — say the word and we will write it up against the
`write-issue` standard rather than leave it here.

PR #313, opened 2026-08-14
fix(slack): reconcile binding and verify delivery

sjarmak:
## Summary

- reconcile `bind-room` with gc's authoritative direct-binding table before writing the pack-local record
- replace a stale direct binding when `--binding-owner` is supplied
- remove a direct binding for group-only rooms so it cannot shadow the group route
- fail before saving local configuration when authoritative reconciliation fails
- document that callers must repeat `--binding-owner` when outbound publication must remain enabled
- treat `chat.postMessage` as write acceptance, not delivery evidence
- read the exact timestamp and posted content back from the intended channel or thread before returning `delivered=true`
- distinguish a valid read with no matching acknowledgement from an unavailable read path

## Why

The pack-local Slack binding and gc's extmsg binding table can disagree. Inbound resolution consults the direct binding first, so a stale record can route a human message to an inactive session even though the visible room group is correct. The command now updates the surface the resolver actually reads and saves its local receipt only after that succeeds.

Outbound publication also treated Slack's successful write response as proof that the intended channel contained the message. The adapter now uses the returned timestamp to read the exact content back through Slack's history API. Missing or mismatched content fails as `readback_unconfirmed`; transport, auth/scope, HTTP, or schema failures fail separately as `readback_unavailable`.

## Verification

- `go test ./... -count=1`, `go test -race ./... -count=1`, `go vet ./...`, and `go build ./...` in `slack-full/adapter`
- `python3 -m pytest -q slack-full/tests/test_slack_chat_publish_to_channel.py slack-full/tests/test_slack_chat_bind_room.py` — 24 passed
- tests cover replacement, group-only unbind, authoritative failure, save ordering, root and threaded readback, exact content, read-path failure classification, response bounds, rewritten mentions, and idempotency
- independent cross-family review `gc-803204` passed after active mutation testing

### DIFF
--- a/slack-full/README.md
+++ b/slack-full/README.md
@@ -57,7 +57,9 @@ for the full rationale.
 Implemented:
 
 - [x] `gc slack bind-dm` — bind a Slack DM channel to one named session
-- [x] `gc slack bind-room` — bind a room to multiple sessions; flags
+- [x] `gc slack bind-room` — bind a room to multiple sessions; requires
+      exactly one of `--binding-owner SESSION` / `--group-only` to
+      declare the room's binding shape; further flags
       `--enable-peer-fanout`, `--allow-untargeted-publication`,
       `--max-peer-triggered-publishes`, `--max-total-peer-deliveries`,
       `--default-handle`, `--handle HANDLE=SESSION` (creates a
@@ -295,13 +297,25 @@ reminder to the other bound sessions so they see what their peer just
 said.
 
 `--binding-owner SESSION` is what makes outbound publishes (and
-therefore `gc slack reply-current --via gc`) actually work. Without
-it, peer fanout still fires on inbound, but `/extmsg/outbound` has
-no `SessionBindingRecord` to resolve the conversation through and the
-publish is rejected. The owner must be one of the participants —
-prefer the session that "owns" the room from gc's perspective. Pass
-the gc session id (e.g. `gc-77139`) when alias resolution semantics
-matter; for stable named sessions, the alias works too.
+therefore `gc slack reply-current --via gc`) actually work: without a
+direct binding, `/extmsg/outbound` has no `SessionBindingRecord` to
+resolve the conversation through and the publish is rejected, even
+though peer fanout still fires on inbound. The owner must be one of
+the participants — prefer the session that "owns" the room from gc's
+perspective. Pass the gc session id (e.g. `gc-77139`) when alias
+resolution semantics matter; for stable named sessions, the alias
+works too.
+
+Every run must declare which of the two shapes it wants: pass
+`--binding-owner SESSION` for a room with an outbound publisher, or
+`--group-only` for a purely group-routed room. `--group-only` removes
+*every* active direct binding for the conversation — including ones
+created by other tooling — so that none can shadow the group route,
+and it prints each removed binding to stderr. Because that removal is
+destructive, it is never the default: a run with neither flag is
+rejected before any API call rather than quietly sweeping the room's
+publisher. Re-running `bind-room` on an owned room therefore has to
+name the owner again.
 
 ## Adapter as a proxy_process service
 
--- a/slack-full/adapter/SETUP.md
+++ b/slack-full/adapter/SETUP.md
@@ -70,9 +70,19 @@ curl -s https://<your-tailnet>.ts.net/healthz
    **OAuth & Permissions** →
    - Bot Token Scopes — add:
      - `chat:write` (post messages)
-     - `im:history` (read DM history for inbound replies)
+     - `im:history` (read DM history for inbound replies **and** to verify
+       outbound ones — see below)
      - `im:read` (open DM channel)
      - `users:read` (resolve display names — optional but useful)
+
+     A history scope is required for every conversation kind you bind, not
+     just DMs: `/publish` confirms each post by reading it back, so a token
+     without history for a bound channel reports every publish as
+     `Delivered:false` / `readback_auth` even though the message is visible in
+     Slack. Add `channels:history` (public channels), `groups:history`
+     (private channels) and `mpim:history` (group DMs) when you bind rooms.
+     `slack-full/schema/apps.schema.json` is the authoritative list for
+     manifest-driven installs.
    - Click **Install to Workspace** at the top, approve.
    - Copy the **Bot User OAuth Token** (`xoxb-...`) — you need it.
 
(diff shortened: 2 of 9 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277289206

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277289275

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277309025

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277316161

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @sjarmak! This PR closes two real holes in the Slack
pack at once: it makes room binding reconcile the *authority* it is replacing
instead of silently leaving a stale direct binding to shadow inbound resolution,
and it turns `/publish` from "the write was accepted" into "the message is
actually there" by verifying delivery with a readback. Both matter because the
adapter's delivery evidence is what the rest of Gas City trusts when it decides
whether a message needs to be re-sent — a publish that reports success without
being observable is exactly the failure that makes retries unsafe.

branch.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- **Delivery evidence compared the wrong two things.** The readback byte-compared
  the text we posted against the text Slack had *stored*, but Slack canonicalizes
  what it stores (entity escaping, auto-linking). Any message containing `&`, `<`,
  `>` or a bare URL would therefore read back as "not the message I sent" and be
  reported undelivered. The final patch compares along two arms instead: a
  `_ref:<hex>_` tail match for keyed publishes (those characters sit outside
  Slack's escape set and are not link-shaped), and a both-sides fold through
  `slackNormalizeText` for everything else.
- **A readback failure after an accepted write left the receipt uncacheable**, so
  a retry would re-*post* rather than re-*verify* — turning one delivered message
  into two. The final patch caches posted-but-unconfirmed receipts under the
  idempotency key and has replay re-verify by readback, releasing the key only on
  a positive absent report.
- **`bind-room` without `--binding-owner` silently destroyed live direct
  bindings.** It now requires the declaration unconditionally and validates it
  before any API call, because gc exposes no conversation-scoped binding read —
  so "only when a binding exists" is not observable from the script, and failing
  closed is the only honest reading. Swept bindings are reported on both stderr
  and `unbound_bindings`, and the flag is propagated through argparse help, all
  four `help.md` examples, and the README.
- **Two invented values on gc's closed `PublishFailureKind` enum.** The readback's
  new failure modes now map onto gc's documented six-value set
  (unconfirmed→`permanent`, unavailable→`transient`, auth→`auth`) with the
  adapter-level detail carried in `Metadata["readback"]`, pinned against the
  literal gc set so a gc-side enum change fails loudly rather than drifting.
- **A not-present readback failed immediately**, which reads Slack's indexing lag
  as a lost message. It now re-checks over a short bounded window inside the
  existing client budget, keeping the immediate failure for a genuine content
  mismatch.

Review findings addressed:
- All five gaps above were resolved by the maintainer fixup `e0f7747`, each with a
  live regression pin rather than a one-off assertion — the mutation probes that
  revert the fix (raw byte compare; delivered-only caching; disabling the release
  branch) all redden the suite.
- *Test evidence quality* (minor, high confidence, review synthesis): the fold's
  behavior on entity-literal payloads was pinned by nothing — the test fake escapes
  unconditionally, modelling a stricter Slack than the documented one, so no fixture
  could observe a change in those semantics. `e0f7747` adds an entity-literal row to
  the compare table and names the fake's escape model in a comment, so the current
  behavior is a pin rather than an accident.
- *Architectural consistency* (nit, high confidence, gemini + a second site found in
  synthesis): the `/publish-file` docs claimed a missing `files:write` scope surfaces
  as `permanent`, but both upload steps classify through `mapSlackError`, which maps
  `missing_scope` to `auth`. Corrected at **both** doc sites in `e0f7747`, not just
  the one originally reported.
- *Error handling* (nit, high confidence, claude): a `config.json` that exists but
  cannot be read escaped `_recorded_binding_owner`'s `except`, so the fail-closed
  refusal arrived as a traceback instead of the curated message naming
  `--binding-owner`. `e0f7747` catches `OSError` alongside `GCAPIError` and extends
  the "unreadable local state" test to a genuinely unreadable path — a directory at
  the config location, which reproduces for every uid, rather than a chmod-000 file.

One reported issue was deliberately *not* acted on. A behavioral-correctness finding
argued that `matchReadbackMessage` should treat entity-literal text as confirming;
measured against the suite, its premise turned out to be the test fake's model of
Slack rather than Slack itself, and its proposed remedy was an exact inversion of the
shipped semantics that left the entire adapter suite green. Changing the matcher to
satisfy the fake would have reintroduced the original bug from the other side, so the
behavior stands as written.

- `mc-71i9u` — *Change impact / blast radius* (minor, medium confidence, gemini): the
  documented 22.4s readback worst case holds for the fresh-publish path but not for
  the replay-then-repost path this fixup introduced, where post + two retryable
  absent-read windows can compound to ~34.8s against gc's 30s adapter budget. The
  follow-up gives the replay re-verify its own smaller window and restates the bound
  the timing test currently names without enforcing.
- `mc-cuq6u` — *Change impact / blast radius* (minor, high confidence, gemini): the
  reconciliation this PR added to `bind-room` was not extended to `bind-dm`, which
  still plain-binds; re-binding a DM therefore still hits a 409 or leaves a stale
  binding shadowing. Deferred as a defensible scope call — it is a contract change to
  a second command, and the failure is visible rather than silent.
- `mc-f1q99` — long-term residuals this PR left unwidened: the `/react` handler's
  `missing_scope` classification drift, the shared `gc_mock.py` lacking bind-room
  routes, and same-key concurrent publishes remaining non-atomic. All pre-existing and
  unchanged here; recorded so they outlive the review artifacts.

### Maintainer Changes
`e0f7747` fix(slack-full): make the publish readback match what Slack actually
stores — 9 files changed, 1367 insertions(+), 179 deletions(-). Your two commits
are preserved as authored; the fixup sits on top and was rebased onto current
`main` with an unchanged patch-id.

### Final Review Status
Ready for the merge queue: final head `e0f7747` has passing required GitHub checks
(Check, Analyze (go), Analyze (python), CodeQL), and the merge-ready metadata is in
place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/gascity-packs/actions/runs/34313373147

### Review Iterations
2 review passes performed. The first found one blocker and four majors across three
independent reviewer lanes; a single maintainer fixup addressed all of them plus three
of the non-gating items, and the second pass confirmed each fix with its own
regression control, verified no test pins were dropped, and approved with no required
changes.

---

PR #317, opened 2026-08-14
fix(gastown/worktree): sync against origin/<branch> explicitly, not implicit tracking

### LINKED ISSUE #299: gastown/worktree-setup.sh --sync silently stops syncing a worktree forever when its branch has no upstream
## Summary

`worktree-setup.sh --sync` can stop syncing a worktree **permanently and
silently**. The failure is swallowed twice — once by `2>/dev/null` and once by
`|| true` — so the directory quietly freezes at whatever commit it was created
at, and nothing in any log, bead, or doctor check reports it.

We hit this in production. It cost us real work, and the failure mode is
unusually nasty for reasons I'll explain below.

## Where

`gastown/assets/scripts/worktree-setup.sh`, currently lines 46-47 on `main`:

```sh
git -C "$WT" fetch origin 2>/dev/null || true
git -C "$WT" pull --rebase 2>/dev/null || true
```

Invoked from `gastown/agents/polecat/agent.toml`:

```toml
work_dir  = ".gc/worktrees/{{.Rig}}/polecats/{{.AgentBase}}"
pre_start = ["{{.ConfigDir}}/assets/scripts/worktree-setup.sh {{.RigRoot}} {{.WorkDir}} {{.AgentBase}} --sync"]
```

## Why it fails

`git pull --rebase` with **no refspec** requires `branch.<name>.remote` and
`branch.<name>.merge` on the checked-out branch. Without them it exits non-zero
with `There is no tracking information for the current branch`.

`2>/dev/null` discards the message. `|| true` discards the exit code. The script
continues and reports success.

## Which branches lack tracking

The script picks its start point from `refs/remotes/origin/HEAD`. When that ref
is missing it falls back to:

```sh
git worktree add <wt> -b <branch>
```

with no start point, which sets **no upstream**. Branches created after `origin`
exists get `-b <branch> refs/remotes/origin/HEAD` and track for free.

So the affected population is: **any agent worktree whose branch was created
before the rig had an `origin`** — which includes every rig set up locally and
given a remote afterwards. Once a branch is in this state it never recovers,
because the only thing that would fix it is the sync that is failing.

## Why this is worse than a stale directory

The polecat `work_dir` is `.../polecats/{{.AgentBase}}`, and `AgentBase`
resolves to the **pool template name**, so with `max_active_sessions = 5` every
pool instance launches into the *same* directory. One frozen worktree therefore
freezes the cwd of the entire pool at once. (Filed separately — that sharing is
its own question; here it is the amplifier.)

On our city this left the shared polecat cwd **31 commits behind**, with an
entire component directory absent from disk that had been merged to `main` weeks
earlier.

The damaging part is the *corroboration*. A worker greps from cwd, finds
nothing, and concludes the code does not exist. The stale bead text says the same
thing. Two apparently independent signals agree, and the agent reasonably
believes both. Our workers repeatedly arrived one step away from re-implementing
code that already existed on `main`. A loud failure would have cost minutes; this
cost us a recurring class of wasted work and a diagnostic detour to find it.

## Diagnosis

The command that reveals what the sync is hiding:

```sh
git -C <worktree> rev-parse --abbrev-ref '@{upstream}'
```

On an affected directory this prints the fatal error `--sync` swallowed.

## Workaround (config only, safe under a live agent)

```sh
git -C <rig-root> branch --set-upstream-to=origin/main <branch>
```

After that the existing `--sync` fast-forwards on every start. No template
change, no pack-cache patch.

## Suggested fix

Any of these would close it; roughly in order of preference:

1. **Set tracking at creation.** In the no-`origin/HEAD` fallback path, once a
   remote exists, set the upstream so `--sync` can work at all.
2. **Make `--sync` state its own failure.** Check `@{upstream}` before pulling
   and emit a warning when absent, rather than attempting a pull that cannot
   succeed. Even a single line on stderr would have made this a five-minute
   diagnosis.
3. **At minimum, stop discarding the exit code** on the sync path. `|| true` is
   defensible for `fetch` (offline is survivable); for the `pull` that defines
   whether `--sync` did anything, it converts a hard failure into a false
   success.

Happy to send a PR for whichever shape you prefer.

## Notes

- Confirmed present on `main` today, read from the repo rather than from our
  local pack cache.
- **Not fixed by #287**, which edits this file and rewrites nearby `fetch` lines
  but does not touch the `pull --rebase` at line 47.
- Related to **#209**. This is arguably a live instance of one of that RFC's
  stated invariants — *"Unexpected source/Git state blocks visibly; it is never
  silently reset, overwritten, adopted, or routed through an old fallback"* —
  being violated by the shipping code today. Reporting it separately because it
  is a concrete bug with a small fix, rather than part of that design change.

rjgeng:
Fixes #299.

`sync_worktree()` ran `git pull --rebase` with no refspec, which
resolves its target via the current branch's
`branch.<name>.remote`/`.merge` config ("upstream tracking"). A worktree
branch created via this script's own no-start-point fallback — taken
whenever `origin/HEAD` isn't configured at creation time, which covers
any rig set up locally and given a remote afterwards — never gets that
config, because `git worktree add -b <branch>` with no explicit start
point doesn't set up tracking. `git pull --rebase` then fails with "no
tracking information for the current branch" on every single `--sync`
call, forever. Both the `fetch` and the `pull` are `2>/dev/null || true`,
so the failure is completely invisible — no log, no bead, no doctor
check reports it. The worktree just freezes at whatever commit it was
created at.

Reproduced the actual git behavior directly in a throwaway repo before
touching any code: `git worktree add -b <branch> refs/remotes/origin/HEAD`
(the normal creation path) does set up tracking, but
`git worktree add -b <branch>` with no start point (the fallback path)
does not.

## Fix

`sync_worktree` now fetches and rebases against the origin branch of the
*same name* explicitly (`git fetch origin "$CUR_BRANCH"` / `git pull
--rebase origin "$CUR_BRANCH"`), reading the branch name directly off
the worktree's own `HEAD` instead of depending on tracking config to
resolve it implicitly. This sidesteps the missing-config problem
entirely rather than patching the config after the fact, and preserves
existing best-effort semantics: no origin counterpart yet is still a
clean no-op, and the already-working tracked-branch case still syncs
correctly.

## Tests

No existing test covered `worktree-setup.sh`'s sync path at all, so this
adds a new file: `gastown/tests/test_worktree_setup_sync.sh`, three
subtests:

- `test_sync_pulls_when_branch_lacks_origin_tracking` — reproduces
  #299's exact production scenario (no-start-point branch, pushed once,
  origin advances) — RED against the unfixed script (confirmed via
  `git stash`), GREEN after.
- `test_sync_is_noop_without_origin_counterpart` — an unpushed branch
  with nothing to sync stays untouched and error-free.
- `test_sync_still_works_with_configured_tracking` — the normal
  already-working tracked case keeps syncing correctly (no regression).

Full `gastown/tests/test_*.sh` suite passes (two unrelated local-only
failures — missing `tomllib` on default `python3`, macOS `date` lacking
GNU's `-d` — are pre-existing environment gaps already documented by the
suite itself, not regressions from this change).

### DIFF
--- a/gastown/assets/scripts/worktree-setup.sh
+++ b/gastown/assets/scripts/worktree-setup.sh
@@ -43,8 +43,20 @@ sync_worktree() {
     if ! git -C "$WT" remote get-url origin >/dev/null 2>&1; then
         return 0
     fi
-    git -C "$WT" fetch origin 2>/dev/null || true
-    git -C "$WT" pull --rebase 2>/dev/null || true
+    # Fetch and rebase against the origin branch of the SAME name explicitly,
+    # rather than relying on `git pull --rebase`'s implicit upstream
+    # (`branch.<name>.remote`/`.merge`). A worktree branch created via the
+    # no-start-point fallback below (no `origin/HEAD` at creation time) never
+    # gets that tracking config, so the implicit form fails with "no tracking
+    # information for the current branch" -- silently, since both git calls
+    # here are best-effort. That failure is permanent: nothing later sets the
+    # missing tracking config, so every subsequent --sync repeats it forever.
+    # Naming the branch explicitly sidesteps the missing config entirely, and
+    # is a no-op (as before) when origin has no branch of this name yet.
+    CUR_BRANCH=$(git -C "$WT" symbolic-ref --short HEAD 2>/dev/null || true)
+    [ -n "$CUR_BRANCH" ] || return 0
+    git -C "$WT" fetch origin "$CUR_BRANCH" 2>/dev/null || return 0
+    git -C "$WT" pull --rebase origin "$CUR_BRANCH" 2>/dev/null || true
 }
 
 branch_name() {
--- a/gastown/tests/test_worktree_setup_sync.sh
+++ b/gastown/tests/test_worktree_setup_sync.sh
@@ -1,0 +1,116 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+ROOT=$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd)
+SCRIPT="$ROOT/gastown/assets/scripts/worktree-setup.sh"
+
+fail() {
+    echo "FAIL: $*" >&2
+    exit 1
+}
+
+git_c() {
+    git -c user.email=a@a.com -c user.name=a "$@"
+}
+
+# new_upstream_and_rig — a bare "upstream" repo plus a rig clone of it, with
+# one commit on main and origin/HEAD already configured (the common case:
+# the rig was cloned from an origin that already existed).
+new_upstream_and_rig() {
+    local base="$1"
+    git_c init -q --bare "$base/upstream.git"
+    git_c init -q -b main "$base/seed"
+    (cd "$base/seed" && git_c commit -q --allow-empty -m init \
+        && git_c remote add origin "$base/upstream.git" \
+        && git_c push -q origin main)
+    git_c clone -q "$base/upstream.git" "$base/rig"
+}
+
+test_sync_pulls_when_branch_lacks_origin_tracking() {
+    local base="$1"
+    local rig="$base/rig" wt="$base/wt-notrack"
+
+    # Reproduce #299's precondition directly: a worktree branch created by
+    # the no-start-point fallback (the path taken when origin/HEAD isn't
+    # configured at creation time -- e.g. a rig set up locally and given a
+    # remote afterwards) has no branch.<name>.remote/.merge config at all.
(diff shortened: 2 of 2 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277289060

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277289800

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277308906

PR #318, opened 2026-08-14
docs(do-work): resolve default branch locally first, network only as fallback

### LINKED ISSUE #305: prepare-worktree resolves the default branch over the network on every run
Follow-up to #228, which correctly stopped basing prepare-worktree worktrees on the launcher's local `HEAD` and switched to the fetched remote default branch.

The remaining rough edge is how the default branch is resolved:

```
DEFAULT_BRANCH=$(git remote show origin 2>/dev/null | sed -n 's/.*HEAD branch: //p')
```

`git remote show origin` contacts the remote. That makes every worktree prepare pay a network round trip, and it pairs badly with the fail-closed branch immediately after it: when the value comes back empty the step refuses rather than falling back, which is the right call for a stale base but means a transient network failure now refuses dispatch instead of degrading.

A local read avoids both costs:

```
DEFAULT_BRANCH=$(git symbolic-ref --short refs/remotes/origin/HEAD 2>/dev/null | sed 's|^origin/||')
```

That ref is written at clone time and refreshed with `git remote set-head origin -a`, so the lookup is local and the subsequent `git fetch --prune origin "$DEFAULT_BRANCH"` still guarantees the base is current. Keeping the fail-closed behavior when the ref is genuinely missing preserves what #228 was protecting against.

Worth confirming before changing: whether any consumer clones with a configuration that leaves `refs/remotes/origin/HEAD` unset, since those would need the `set-head` refresh wired in rather than assumed.

rjgeng:
Fixes #305. Follow-up to #228.

`do-work`'s `prepare-worktree.md` step resolves the remote default
branch with `git remote show origin`, which contacts the remote. Every
worktree prepare pays that network round trip, and it pairs badly with
the fail-closed guard right after it: a transient network failure now
refuses dispatch entirely instead of degrading, since the network call
was the only path to a result.

## Fix

Reads `refs/remotes/origin/HEAD` locally first
(`git symbolic-ref --short refs/remotes/origin/HEAD`, strip the
`origin/` prefix) — that symref is written by a normal `git clone` and
kept current by `git remote set-head origin -a`, so the common case now
resolves at zero network cost.

On the reporter's own flagged caveat (worth confirming whether any
consumer clones with a configuration that leaves
`refs/remotes/origin/HEAD` unset): rather than assume every consumer
clones normally, kept the original network-based command as an explicit
fallback when the local read comes back empty. That's the exact same
"rig set up locally, remote added afterward" scenario `#299` reproduced
this session for a different script, so this doesn't guess it away.
Behavior is strictly a superset of the old command now: free and
identical whenever `origin/HEAD` is set (the common case), unchanged
from before when it isn't. The existing "if it is empty, fail closed"
guard is untouched and now only triggers when both reads fail.

## Verification

No test parses this workflow doc's prose. Directly reproduced: confirmed
the new local-read command produces byte-identical output to the old
network-based command against a normal clone, and confirmed it correctly
returns empty (triggering the fallback) when `origin/HEAD` is unset.
Full `gastown/tests/test_*.sh` suite passes clean.

### DIFF
--- a/gascity/assets/workflows/do-work/prepare-worktree.md
+++ b/gascity/assets/workflows/do-work/prepare-worktree.md
@@ -24,9 +24,15 @@ setup only. Do not edit source files in the launcher checkout.
    `$(pwd)/worktrees/<source-anchor-id>`, based on the up-to-date remote
    default branch — never the launcher's local `HEAD`, which may be behind
    `origin`. If the path is missing:
-   - Resolve the remote default branch (do not hardcode `main`):
+   - Resolve the remote default branch (do not hardcode `main`), reading the
+     local `origin/HEAD` symref first so a normal prepare does not pay a
+     network round trip on every item:
+     `DEFAULT_BRANCH=$(git symbolic-ref --short refs/remotes/origin/HEAD 2>/dev/null | sed 's|^origin/||')`.
+     If that is empty (no local record of the remote's default branch —
+     e.g. a repo whose `origin` was added after local setup rather than via
+     `git clone`), fall back to the network form:
      `DEFAULT_BRANCH=$(git remote show origin 2>/dev/null | sed -n 's/.*HEAD branch: //p')`.
-     If it is empty, fail closed — do not fall back to local `HEAD`.
+     If both are empty, fail closed — do not fall back to local `HEAD`.
    - Fetch it so the base is current:
      `git fetch --prune origin "$DEFAULT_BRANCH"`.
    - Create the worktree detached at the freshly fetched tip:


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277288775

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277288848

rjgeng:
Superseded — the issue this PR targets, #305, is already closed and fixed on `main` by a more complete version: PR #331 (`fffa8fd`) not only reads the local `origin/HEAD` symref first, it also repopulates it via `git remote set-head origin --auto` before falling back to the network, avoiding both the round trip *and* the repeated network fallback on repos where the local symref is missing but recoverable (e.g. `origin` added after local setup rather than via `git clone`). This PR's version jumps straight to the network fallback without that repopulation step. Closing as redundant rather than rebasing toward a strictly weaker duplicate of what's already on main.

PR #320, opened 2026-08-14
fix(gastown/witness): re-verify liveness before delete, content-check merges past rebase/squash

### LINKED ISSUE #286: mol-witness-patrol/recover-orphaned-beads: two silent-destructive defects (stale liveness map deletes live worktrees; --is-ancestor false-negatives on rebase)
Two independent defects in the same step — `recover-orphaned-beads` — of `gastown/formulas/mol-witness-patrol.toml`. Both fail in the silent direction: the witness logs a normal-looking recovery while destroying or duplicating work. Both were caught **live during patrol**, not hypothesised.

Observed at pin `sha:33d3a430a67d1782ad364556cb566bdb01d0afe3` (the sha our `pack.toml` currently pins for `imports.gastown`). Every rig's witness on the box carries both — all six formula paths across two cities are symlinks to the one shared cache file, so this is not city-local and a local patch would be reverted on the next pack sync.

---

## Defect 1 (high, destructive): the liveness snapshot ages inside the loop, so a *live* polecat can be reset

`recover-orphaned-beads` step 1 says (around line 174):

> 1. Build the assignee→state liveness map ONCE per cycle (before the per-bead loop), then resolve each bead's assignee against it.

The map is built once and then consulted throughout the per-bead loop. A polecat that spawns **after** the map is built resolves as `absent`, and per the classification rules (~line 246) `absent` after the exact lookup means *orphaned* — worktree removed, bead reset to pool — while that agent is alive and mid-edit.

**Observed live.** Map built 08:43Z (9 sessions, 26 identity keys) contained no entry for session `de-o2d` / `sound_calculation/gastown.slit`. At 08:44:37Z a bead was `in_progress` under exactly that assignee; at 08:45Z the session was active and mid tool-call. Had the per-bead loop reached that bead a minute later — well inside a normal cycle — the worktree would have been deleted, uncommitted work lost, and a second polecat dispatched onto the same bead. Corroborated from the other side: `de-o2d` spawned 10:43:10 local in an unrelated session-survival measurement, the same event seen independently.

**The existing fail-safe does not cover this.** The guard at ~line 218 refuses only an *empty* map. This map was fully populated and correct for every session that existed when it was built. It was stale by exactly one. A non-empty map is not evidence that a given assignee was resolvable at the moment the bead was read.

**Suggested fix.** Only `absent` triggers destruction, so only `absent` needs to be fresh; every other state means do-not-touch and is safe to cache. Keep the prebuilt map for classification, then re-verify each **candidate orphan** with a fresh single-identity lookup taken after the bead was read, immediately before recovery. That is one extra lookup per candidate orphan, not per bead. A second cheap guard: refuse to recover any bead whose `updated_at` is newer than the map's build time.

Worth fixing in the recipe text as well as the logic — it currently presents map-before-loop as correct without noting that the snapshot ages, which is what makes the window easy to reintroduce.

---

## Defect 2 (medium, duplicative): the merge check false-negatives on rebase, so merged work is re-dispatched

Step 3, "Verify work on main before resetting" (~line 322):

```bash
git merge-base --is-ancestor origin/$BRANCH origin/main
```

ancestor → close the bead; not ancestor → proceed with reset → back to pool → re-dispatched.

That test only holds when the refinery fast-forwards or merges. When the refinery **rebases**, the branch tip is never an ancestor of main even though every line of the work is on main.

**Live evidence.** `origin/polecat/<bead>` = `a6c3d4f`, `origin/main` = `c03d6d5` with single parent `76cb607` (i.e. a rebase). `--is-ancestor` returns false. But `c03d6d5`'s subject names the bead, `git diff --stat 76cb607 c03d6d5` is exactly that bead's 5-file changeset (437 insertions, 13 deletions), and `git diff origin/main a6c3d4f` over those 5 paths is **empty**. The bead's own metadata records `merge_result=merged`, `merged_sha=c03d6d5`. Fully merged; the formula says no.

**This is the normal case, not an edge case.** The precheck was written when main was at an earlier commit ("clean fast-forward"), then another bead merged first and moved main, so this one was no longer a fast-forward by the time it was processed. Any queue with 2+ beads produces this routinely.

Independently reproduced from the refinery side in the same cycle: the refinery's own ancestor guard refused to delete two polecat branches because the pre-rebase commits were not in main, and only proceeded after confirming patch-id equivalence.

**Impact.** An orphaned bead whose work is already merged-by-rebase is reset and re-dispatched. A new polecat re-applies changes already on main — at best a wasted cycle producing an empty diff, at worst a confused re-implementation or a conflicting revert of newer work.

**Suggested fix.** Replace the ancestry test with a **content** test, which is correct for fast-forward, merge, rebase *and* squash: an empty diff over the bead's paths means the work is on main. Patch-id equivalence works too and is what the refinery already fell back to.

---

## Why one issue

Same formula, same step, same failure mode (silent + destructive), and the two fixes touch adjacent logic. Fixing one without the other leaves the step half-safe.

## Why this is time-sensitive rather than cosmetic

The interim mitigation — treat every candidate orphan as suspect and re-verify liveness before any destructive step — lives only in operator mail and in the formula's prose. Because the formula is shared and pinned, every witness has to re-apply that rule **by hand every cycle** until the pin moves. One witness that restarts, recycles context, or simply follows the recipe as written re-opens the destructive window. The mitigation does not persist; only a released fix and a pin bump do.

Happy to test a fix against the pin here and report back — we have both defects reproducing on real cycles. Note we have read-only access to this repo, so we can't open a PR directly; if you'd prefer a patch, say the word and we'll attach diffs against the two blocks above.

rjgeng:
Fixes #286.

Two independent defects in `mol-witness-patrol.toml`'s
`recover-orphaned-beads` step, both caught live during patrol, both fail
in the silent-destructive direction: the witness logs a normal-looking
recovery while destroying or duplicating real work. This is a shared,
pinned pack — a local patch would be reverted on the next sync, so it
needs a real fix here.

**Defect 1 (high, destructive).** The assignee->state liveness map is
built once per patrol cycle, then consulted for the rest of the per-bead
loop. A polecat that spawns *after* the map is built resolves as
`absent`, which the classification rules treat as orphaned — so a live
agent minutes into real work can have its worktree deleted and its bead
reset to pool out from under it. The existing fail-safe only guards an
*empty* map, not a map that's simply one entry stale.

**Defect 2 (medium, duplicative).** The merge check
(`git merge-base --is-ancestor origin/$BRANCH origin/main`) only holds
when the branch's own commit objects are literally reachable from main —
true for a fast-forward or a real merge commit, but false whenever the
refinery lands the branch by *recreating* its commits (rebase onto a
main that moved since the branch was cut, or a squash). That's the
normal case once a queue has 2+ beads, not an edge case. A false
negative here re-dispatches already-merged work — wasted at best, a
confused re-implementation or conflicting revert at worst.

Confirmed both defects still reproduce verbatim against current
`origin/main` (not just the issue's pinned sha — this file has moved 3
commits since that pin; checked each one's actual diff, none touch this
step's logic). For Defect 2, independently reproduced the false-negative
in a throwaway git repo before writing the fix: a branch rebased onto a
main that had already diverged, then landed, correctly shows
`is-ancestor=false` even though its content is fully on main.

## Fix

- **Defect 1**: Step 3b (the one truly destructive point — right before
  worktree deletion + pool reset) now re-verifies the specific candidate
  orphan's liveness with a fresh single-lookup query immediately before
  acting, plus a second guard comparing the bead's own `updated_at`
  against the Step 1 map's build timestamp (`MAP_BUILT_AT`, newly
  captured). Only candidates that already survived Step 1's orphan
  classification pay the extra lookup.
- **Defect 2**: keeps the ancestor check as a cheap first pass, falls
  back to a content test scoped to the branch's own changed files when
  ancestry says no — holds for fast-forward, merge, rebase, and squash
  alike, since it compares tree content instead of commit-object
  reachability.

Fixed both in one commit — same step, same failure shape, and Defect 1's
fix sits directly downstream of Defect 2's `ON_MAIN` determination in the
same Step 3/3b flow.

## Verification

No test suite exercises this file's prose/recipe semantics — only
`test_gastown_pack_assets.sh`'s TOML-parse/structural checks apply,
which pass. In place of RED/GREEN: reproduced Defect 2's false-negative
and the fix's correction in a throwaway repo (plus confirmed no false
positive on a genuinely unmerged branch), `bash -n` on all three new
embedded bash blocks, verified the fresh-liveness jq query against both
a live and an absent assignee, re-ran the full `gastown/tests/test_*.sh`
suite.

### DIFF
--- a/gastown/formulas/mol-witness-patrol.toml
+++ b/gastown/formulas/mol-witness-patrol.toml
@@ -213,7 +213,13 @@ a NEW session with a different ID — the old one never comes back.
    SESSION_COUNT=$(jq -r '(.sessions // []) | length' "$SESSIONS_FILE")
    MAP_COUNT=$(printf '%s' "$LIVENESS_MAP" | jq -r 'length')
    rm -f "$SESSIONS_FILE" "$SESSION_BEADS_FILE"
+   MAP_BUILT_AT=$(date -u +%Y-%m-%dT%H:%M:%SZ)
    ```
+
+   `MAP_BUILT_AT` is this snapshot's timestamp — every bead the per-bead
+   loop below reaches was read from the store *after* this instant, but
+   its own `updated_at` may be newer still if something touched it mid-cycle.
+   Step 3b uses this to catch that case before it destroys anything.
 
    **Fail safe — never orphan on an empty map.** If the map is empty while
    sessions exist, the schema has drifted again (the exact failure in
@@ -323,14 +329,44 @@ starts fresh.
 **Step 3: Verify work on main before resetting.**
 
 Before resetting the bead to pool, check if the branch was already merged
-to main (polecat finished, pushed, but crashed before closing the bead):
+to main (polecat finished, pushed, but crashed before closing the bead).
+
+**Ancestry alone is not enough.** `git merge-base --is-ancestor` only holds
+when the branch's own commit objects are literally reachable from main —
+true for a plain fast-forward or a real merge commit. It false-negatives
+whenever the refinery lands the branch by *recreating* its commits
+(rebase-then-fast-forward onto a main that has moved since the branch was
+cut, or a squash) — main then carries a new commit with a different SHA
+that contains exactly the branch's changes, and ancestry can never see
+that. This is the normal case, not an edge case: any queue with 2+ beads
+produces it as soon as a second bead lands after the first was rebased.
+A false negative here re-dispatches already-merged work.
+
 ```bash
-git log main --oneline | grep -q "$BRANCH"
-# Or check if branch commits are reachable from main:
-git merge-base --is-ancestor origin/$BRANCH origin/main
+# Cheap first pass — catches the plain fast-forward/merge-commit case:
+if git merge-base --is-ancestor origin/$BRANCH origin/main; then
+  ON_MAIN=true
+else
+  # Rebase/squash recreates commits, so a "not an ancestor" result does
+  # NOT mean "not merged" -- fall back to a content test, scoped to the
+  # branch's own changed files, which holds regardless of how it landed.
+  MERGE_BASE=$(git merge-base origin/$BRANCH origin/main 2>/dev/null || true)
+  if [ -z "$MERGE_BASE" ]; then
+    ON_MAIN=false   # no shared history found; treat conservatively as not-yet-merged
+  else
+    CHANGED_FILES=$(git diff --name-only "$MERGE_BASE" origin/$BRANCH)
+    if [ -z "$CHANGED_FILES" ]; then
+      ON_MAIN=false   # branch introduced nothing beyond the merge-base; nothing to confirm
+    elif git diff --quiet origin/main origin/$BRANCH -- $CHANGED_FILES; then
+      ON_MAIN=true    # main's current content for exactly these files matches the branch
+    else
+      ON_MAIN=false
+    fi
+  fi
+fi
 ```
 
-If the work is already on main:
+If the work is already on main (`ON_MAIN=true`):
 - **Close the bead** (work is done, don't re-dispatch). Use `--force`: the bead
   is still assigned to the crashed polecat, so this is a cross-actor close and
   bd's close-authority guard (gastownhall/beads#3734) would otherwise refuse it:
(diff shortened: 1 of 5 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277288564

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277288633

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277308820

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277316001

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277329385

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @rjgeng! This PR stops the witness `recover-orphaned-beads` path from
destroying a bead on the strength of a stale liveness snapshot or an ancestor check that a rebase or
squash-merge has already invalidated, which closes a real data-loss window in an automated recovery
path that runs without a human in the loop.

### Original PR Review
Decision: request_changes resolved to approve.

Specific gaps fixed:
- The new "did this content already land on main?" test collected changed paths into an unquoted
  `$CHANGED_FILES` and passed it to `git diff --quiet ... -- $CHANGED_FILES`. Word-splitting made the
  pathspec miss on any path containing whitespace, so the comparison silently narrowed and the
  destructive verdict fired as a false positive. The final patch collects NUL-delimited names into a
  quoted array instead, which also holds for newline-bearing paths where the originally prescribed
  per-file loop would still have destroyed (`--name-only` C-quotes those).
- The fresh liveness re-check failed **open**: a failed or empty map rebuild left the verdict at its
  destructive value, so "we could not verify" and "we verified it is orphaned" were the same outcome.
  The final patch initialises the verdict to the safe value and promotes it only after a provably
  successful rebuild; every cannot-verify arm now skips, and two escalate.
- The `updated_at` staleness guard read the field off a `.metadata` path that never carries it, so it
  was dead code that could never fire. It now reads the bead top level, and the comment that claimed
  otherwise is gone.
- The pre-destruction verdict guarded 1 of the 3 destructive sites. It is now hoisted above the case
  split so every arm sees it, and it is a genuine code gate at both terminal sites.
- The staleness watermark was stamped *after* the snapshot it guards and the comparison inverted in
  the boundary second. It is now stamped before the queries, pinned once per cycle, and compared
  strictly. The fixup also caught a hazard the checklist had missed: rebuilding the map per candidate
  would have eroded the staleness window to sub-second.
- The change shipped entirely unpinned — both a full revert and a full restoration of the original
  defect stayed green. Deleting a guard line now reddens two independent nets by name, and a new
  executed harness exercises the recovery paths directly.

Review findings addressed:
- All six required changes above landed in maintainer commit `fab436d`, five in the directed form and
  one (the pathspec fix) in a deviation verified to be strictly stronger than what was prescribed.

- `mc-ccu2x` — Debuggability & Operability, minor: Step 3b's `STILL_ORPHANED` re-statement is inert.
  The guard at `mol-witness-patrol.toml:640-645` is echo-only with no `continue`, and the destructive
  block at `:647-653` is a separate ungated fence joined only by prose — so the "all three
  destructive sites re-state the verdict" claim, asserted in three artifacts, is false for the third
  site. Non-gating because the base had no guard there at all, Step 2a's blanket instruction covers
  the action, and the blast radius is a lost checkout (the branch ref and its objects survive).
- `mc-5y9xo` — Test Evidence Quality, minor: no net binds a destructive command to its gate. The pins
  are presence-only, so they catch **deletion** of a guard but never **hoisting** of a command out of
  one. Measured: hoisting the terminal cross-actor force-close out of its `if`, leaving the `if` line
  in place, makes it unconditional with all three nets still green — which is the exact failure mode
  the gate comment says these pins exist to stop.
- `mc-fwphp` — Debuggability & Operability, nit: `scripts/gascity_pack_inference_gate.py:153` names
  `gastown/tests/test_mol_witness_patrol_on_main.sh` as the companion coverage; the shipped file is
  `gastown/tests/test_mol_witness_patrol_orphan_recovery.sh` and no `*_on_main.sh` exists in the tree.

### Maintainer Changes
`fab436d` fix(gastown): fail closed before the witness destroys a bead, and pin every guard —
5 files changed, 830 insertions(+), 82 deletions(-). Branch total against the current base:
5 files changed, 864 insertions(+), 46 deletions(-). The branch was also rebased onto the latest
`main` with the adopted patch content unchanged.

### Final Review Status
Ready for the merge queue: final head `fab436d8` has passing required GitHub checks, and the
merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can
pick it up.

CI: https://github.com/gastownhall/gascity-packs/actions/runs/34547423528

### Review Iterations
2 review passes. Iteration 1 returned request_changes with a six-item checklist; the maintainer fixup
landed every item and iteration 2 re-reviewed the fixup against the checklist, validated each fix
against a restoration of the original defect, and approved. The two minor findings that surfaced in
iteration 2 are the non-blocking follow-ups listed above.

---

PR #322, opened 2026-08-14
fix(gastown/witness): complete crashed submit-and-exit handoff instead of resetting to pool

### LINKED ISSUE #276: mol-polecat-work.toml - killed polecat leaves finished, pushed work invisible as open/unassigned bead
#### NOTE this was written by a sonnet 5 agent as  I was trying to debug this problem.

**Title:** `mol-polecat-work.toml`: push/reassign not atomic — killed polecat leaves finished, pushed work invisible as open/unassigned bead

**Repo:** `gastownhall/gascity-packs` (formula: `gastown/mol-polecat-work.toml`, step `submit-and-exit`)

### Summary

In the `submit-and-exit` step of `mol-polecat-work.toml` (~lines 284–421), `git push origin HEAD` happens (~line 368), then bd metadata update (~line 408) and reassignment to the refinery (~line 421) happen as **separate, independently-interruptible steps afterward**. If the polecat session is killed/recycled/restarted (pool reconciler reclaiming an "idle" slot, host restart, context exhaustion, etc.) between the push succeeding and the reassignment completing, the branch is safely on origin but the bead handoff never finishes. The bead sits open/unassigned, invisible to the refinery, indistinguishable from lost work without manually checking git.

### Impact observed

On 2026-08-05, I had four beads hit exactly this gap: real, completed, pushed work with reverted/open bead status.

All four branches exist on origin with real commits, none merged, none reflected correctly in bd. No data was lost — but confirming that took hours of manual forensics (see below), during which it looked like finished work had been destroyed.

### Root cause

Push, bd-metadata-update, and reassign-to-refinery are three separate steps rather than one atomic unit. Any interruption after step 1 and before step 3 completes leaves the system in a state nothing else detects or corrects.

### Why diagnosis was so costly

- Had to do `git reflog` archaeology across nested worktrees to reconstruct what each killed/recycled polecat session had actually done.
- Had to hand-correlate a JSONL reconciler trace log (UTC timestamps) against the reflog timeline (local timestamps) to establish event ordering.
- Got two wrong intermediate conclusions specifically from UTC/local timezone confusion during that correlation, and had to walk them back.
- **There is no single command/view to ask "does this bead's bd status agree with what's actually on origin for its branch"** 

### Requested fix

Collapse push → bd metadata update → reassign-to-refinery into one atomic block in `mol-polecat-work.toml`'s `submit-and-exit` step, so a kill mid-sequence can't leave the bead stranded.

rjgeng:
Fixes #276.

`mol-polecat-work.toml`'s `submit-and-exit` step records
`metadata.target` (step 5) and reassigns the bead to the refinery
(step 6) as two separate, independently interruptible `gc bd update`
calls. A polecat killed in that exact window has already pushed the
branch and run its final clean-state check — everything is done except
the handoff. The reporter hit this for real on four beads in one day and
spent hours of git-reflog/reconciler-trace archaeology confirming
nothing was actually lost.

## Why this wasn't data loss but was still worth fixing

The witness's existing orphan-recovery logic (Case A: branch on origin
-> not yet on main -> Step 3b reset to pool) already handles this
correctly in the sense that no work is lost — `workspace-setup`'s
rejection-aware resume picks the existing branch back up when a new
polecat gets the reopened bead. But it's wasteful (a fresh polecat spins
up just to re-run checks and a no-op push to reach the same step 6) and,
per the reporter, expensive to *diagnose* — the state looks alarming
(open/unassigned, work "gone") even though it isn't.

## Fix

New Step 3a in `gastown/formulas/mol-witness-patrol.toml`, between
Step 3's "not on main" branch and Step 3b's pool reset. When it fires,
the witness completes the handoff directly — the same reassignment
submit-and-exit's own step 6 would have run, the same wake+nudge as its
step 7, the same worktree cleanup Step 3b already does for the identical
reason (the branch is canonical on origin). Otherwise it falls through to
the existing Step 3b reset, unchanged.

**The discriminator is an explicit stage marker, not `target`.** Step 5
now writes `handoff_stage=target_recorded` in the same `gc bd update`
that records `target`, and Step 3a fires only on
`handoff_stage = target_recorded` **and** a live branch on origin.

`target` cannot carry this signal: `gc sling` accepts it as a documented
mint-time input (`workspace-setup` resolves `base_branch` from
`metadata.target` first), nothing in the tree ever unsets it, and it
survives both `mol-refinery-patrol` rejection paths. Beads that never
reached step 5 already carry it, so keying on `target` would hand
never-submitted or already-rejected work to a refinery whose only merge
gate is tests-pass.

`handoff_stage` is a single-purpose marker with a deliberately narrow
contract — **one writer, three clearers**:

| | Site |
|---|---|
| Written | `mol-polecat-work` `submit-and-exit` step 5 (with `target`, one update) |
| Cleared | `mol-polecat-work` `workspace-setup`, on every fresh attempt |
| Cleared | `mol-refinery-patrol`, both rejection paths |

So its presence means the bead genuinely completed step 5 *and* nothing
has since invalidated that. Every branch-ready halt in the push gate
(`auto_push_false`, `metadata_unreadable`, `no_push_rail_unresolved`)
returns before step 5, as do the push-failure paths, so an unpushed
branch never carries the marker. The branch-on-origin conjunct is a
fail-closed backstop — there is nothing to hand off if the refinery's
`--has-metadata-key=branch` query cannot see the work — and it queries
the fully-qualified `refs/heads/$BRANCH`, because `ls-remote` patterns
match ref *tails* and the bare form stays truthy when a tail-colliding
ref outlives the branch.

The reassignment is a **checked `--force` write**: `--force` because this
is a cross-actor write against the dead polecat's still-live
`in_progress` claim, which `bd` refuses otherwise (the same claim guard
Step 3's close cites, gastownhall/beads#3734); checked because the
success arm is the only path that may report success. On failure nothing
above it has mutated anything, so it falls through to Step 3b rather
than mailing `ORPHAN_HANDED_OFF` and removing the worktree — claiming a
handoff that did not happen would strand the bead `in_progress` under a
dead assignee with its worktree gone.

**Root-cause honesty**: true atomicity across push/update/reassign isn't
achievable in bash against an unpredictable kill point — this doesn't
pretend otherwise. It closes the gap from the other side: make the
witness recognize and complete the one specific interrupted state that's
unambiguously safe to complete, rather than trying to prevent the
interruption. The rest of the crash window (before step 5 runs) is
already covered correctly by the existing path this fix doesn't touch.

## Verification

**New static property test** — `gastown/tests/test_gastown_pack_assets.sh`
(+100 lines), in the pack's established style. It pins the parts of this
contract that silently rot if a later edit moves them:

- **Marker census across the three formulas**: exactly *one* writer of
  `--set-metadata handoff_stage=target_recorded`, the `workspace-setup`
  `--unset-metadata handoff_stage`, and exactly *two* refinery clearers.
  Cardinality, not just presence — a second writer reds the suite.
- **Step 3a's position and call order**: a line-anchored sequence
  signature (`guarded-update delete-source mail wake nudge
  worktree-remove else`) asserting 3a sits between Step 3's on-main close
  and Step 3b, that the mail/wake/nudge/cleanup are all *inside* the
  success arm, and that 3b still follows for fall-through states.
- **The discriminator and the exact-ref `ls-remote` form** as literals.

The test was validated by mutation, not just by passing. Each arm was
run against an isolated copy with a green control before and after:

| Mutant | Class | Suite |
|---|---|---|
| Revert the discriminator to the `target`/`branch_ready` form | inversion | **red** |
| Drop `--force` from the reassignment | deletion | **red** |
| Drop one refinery clearer | cardinality | **red** |
| Add a second marker writer | **addition** | **red** |
| Swap the success/failure arms; move 3a after 3b; hoist the mail out of the success arm | ordering | **red** |
| Demote the discriminator to a comment and substitute `if true` | narration | *green* |

The last row is a known and disclosed limit: `grep -F` pins certify that
a literal is present in the file, not that it is live in the executed
block, so commenting a pinned line out keeps them green. The
line-anchored *sequence* pins do not share that blind spot (a comment
breaks the anchor). Closing it properly needs a harness that extracts
and *executes* the block against a stubbed `gc`, which is filed as its
own hardening item rather than papered over with more substring pins.

**Behavior of the block** was checked by execution too: the failure arm
mutates nothing (update-fails → 0 `delete-source`, 0 mail, 0 wake), and
a live `ls-remote` experiment on git 2.43.0 reproduces the ref-tail
behavior the exact-ref form exists to avoid — deleting
`refs/heads/polecat/tst-1` still returns
`refs/heads/nested/polecat/tst-1` for the bare pattern, and 0 lines for
the `refs/heads/` form.

**Suites**: all `gastown/tests/test_*.sh` pass, plus
`tests/test_gascity_pack_inference_gate.py` and
`gascity/tests/test_formula_assets.py`. Note that
`test_formula_assets.py` globs `*.formula.toml` and so does **not** cover
`gastown/formulas/*.toml` — the live consumer for this pack's formulas is
the pack asset suite above, which is why the new pins live there.

Re-verified after rebasing onto current `main`, which independently
rewrote the same `submit-and-exit` step (#273); the two changes are
additive at the same anchor and both sides' pins were confirmed still
live on the merged text.

Touches the same file as #286 (PR #320) but a non-overlapping section —
new Step 3a here vs. Step 3b/3's ancestor+content check there.

### DIFF
--- a/gastown/agents/polecat/prompt.template.md
+++ b/gastown/agents/polecat/prompt.template.md
@@ -87,7 +87,8 @@ Work beads carry structured metadata for lifecycle tracking and handoff:
 |-------|--------|------|-------------|
 | `work_dir` | polecat (branch-setup) | Early | Absolute path to git worktree |
 | `branch` | polecat (branch-setup) | Early | Source branch name |
-| `target` | polecat (submit) | Late | Target branch (default: {{ .DefaultBranch }}) |
+| `target` | caller (sling) or polecat (submit) | Mint-time or Late | Target branch (default: {{ .DefaultBranch }}). `gc sling` accepts it as a mint-time input and nothing ever unsets it, so its presence is **not** a signal that you submitted |
+| `handoff_stage` | polecat (submit step 5) | Late | `target_recorded` once submit has recorded `target`. This — not `target` — is what says submit ran to completion |
 | `existing_pr` | caller | Before dispatch | Existing PR URL to reuse instead of creating another PR |
 | `pr_url` | refinery | PR handoff | Canonical PR URL recorded after validation |
 | `rejection_reason` | refinery (on failure) | On reject | Why the merge was rejected |
@@ -96,12 +97,21 @@ Work beads carry structured metadata for lifecycle tracking and handoff:
 This enables crash recovery — the witness can find and salvage your work.
 
 **On submission:** You update `branch` (may have changed after rebase),
-set `target`, then reassign to refinery. If `existing_pr` is present, leave
-it for refinery to validate and canonicalize into `pr_url`.
+set `target` and `handoff_stage` in the same update, then reassign to
+refinery. If `existing_pr` is present, leave it for refinery to validate and
+canonicalize into `pr_url`. If you die between those two steps, the witness
+reads `handoff_stage` and completes the reassignment for you instead of
+resetting your finished work to the pool.
 
 **On rejection:** The refinery puts the bead back in the pool with
-`rejection_reason` set and the branch intact. A new polecat picks it up,
-sees the existing branch and reason, and resumes instead of redoing everything.
+`rejection_reason` set, `handoff_stage` cleared, and the branch intact. A new
+polecat picks it up, sees the existing branch and reason, and resumes instead
+of redoing everything. Your own workspace-setup clears `handoff_stage` again
+on every fresh attempt, so a stale marker never survives into new work.
+
+The formulas are the source of truth for this contract:
+`mol-polecat-work` writes the marker, `mol-refinery-patrol` and
+workspace-setup clear it, and `mol-witness-patrol` Step 3a is its only reader.
 
 Read metadata:
 ```bash
--- a/gastown/agents/refinery/prompt.template.md
+++ b/gastown/agents/refinery/prompt.template.md
@@ -230,7 +230,12 @@ Never infer a branch name. If `metadata.branch` is missing, reject the bead.
 
 On rebase conflict or test failure:
 1. Put work bead back in pool:
-   `gc bd update $WORK --status=open --assignee="" --set-metadata rejection_reason="..."`
+   `gc bd update $WORK --status=open --assignee="" --set-metadata rejection_reason="..." --unset-metadata handoff_stage`
+
+   Clearing `handoff_stage` is required, not cosmetic, and belongs in the same
+   update as the pool reset: rejection invalidates the polecat's completed
+   submit, and a bead left carrying the marker is one the witness would hand
+   straight back to you as a crash-recovery handoff.
 2. Branch handling depends on failure type:
    - Conflict: leave branch intact (polecat needs it for rebase)
    - Test failure: delete branch (polecat redoes work)
--- a/gastown/agents/witness/prompt.template.md
+++ b/gastown/agents/witness/prompt.template.md
@@ -89,7 +89,11 @@ bead metadata; do not use template-pattern or fixed-prefix matching.
 branch-setup. For each orphaned bead:
 
 1. **Branch on origin** (`metadata.branch` exists, verified on remote) ->
-   worktree disposable. Delete worktree, reset bead to pool.
+   worktree disposable. If the bead also carries
+   `metadata.handoff_stage=target_recorded`, the polecat finished submit
+   through step 5 and only the reassignment is missing: complete that handoff
+   to the refinery (formula Step 3a) rather than discarding finished work.
+   Otherwise delete worktree, reset bead to pool.
 
 2. **Worktree exists, unpushed commits** ->
    commit any remaining uncommitted work (`git add -A && git commit`),
--- a/gastown/formulas/mol-polecat-work.toml
+++ b/gastown/formulas/mol-polecat-work.toml
@@ -83,6 +83,13 @@ if [ -z "$WORK_BEAD_ID" ]; then
   gc runtime drain-ack
   exit 1
 fi
+# Clear any stale handoff marker before touching the work. A fresh attempt on
+# this bead means a prior submit-and-exit step 5 no longer describes it, and a
+# leftover marker would let the witness's Step 3a hand un-submitted work to the
+# refinery if this attempt also crashes. Unconditional on purpose: a bead can
+# arrive here carrying the marker with no rejection_reason (witness pool reset,
+# operator re-sling), so clearing only on the rejection path is not enough.
+gc bd update "$WORK_BEAD_ID" --unset-metadata handoff_stage
 ```
 
 **2. Ensure worktree exists.**
@@ -658,6 +665,7 @@ git branch -D "$BRANCH"              # Branch is pushed; refinery owns it now
 # --append-notes, NEVER --notes — see the note under this block.
 gc bd update "$WORK_BEAD_ID" \
   --set-metadata target={{base_branch}} \
+  --set-metadata handoff_stage=target_recorded \
   --append-notes "Implemented: <brief summary>"
 ```
 Branch was recorded in workspace-setup and is already in metadata.
@@ -682,6 +690,18 @@ the very notes a bare `--notes` would have just deleted. The identity-stamped
 bead record is also the city's anchor against nudge-text spoofing, so a rail
 that silently erases it is a P1, not a formatting nit. The same rule holds
 anywhere a formula writes a note onto a bead it did not create: append.
+
+`handoff_stage=target_recorded` is the stage marker the witness's Step 3a
+uses to recognize a polecat that crashed between this step and step 6 and to
+complete the handoff instead of respinning the work. It must be written here
+and nowhere else: `target` cannot carry that meaning because `gc sling`
+accepts it as a mint-time input, nothing ever unsets it, and it survives
+refinery rejection. Every path that invalidates "step 5 completed" clears the
+marker — `workspace-setup` on each fresh attempt, and both `mol-refinery-patrol`
+rejection paths. Every branch-ready halt in the push gate above
+(`auto_push_false`, `metadata_unreadable`, `no_push_rail_unresolved`) returns
+before this step, as do the push-failure paths, so an unpushed branch never
+carries the marker.
 
 **6. Reassign to refinery:**
 ```bash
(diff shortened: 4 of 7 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277252283

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277288279

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277288348

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277288422

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277308660

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277315841

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @rjgeng! This PR closes a real gap in the witness's
orphan-recovery logic: a polecat killed between `submit-and-exit`'s step 5
(record `target`) and step 6 (reassign to the refinery) had finished every bit
of its work — branch pushed, clean-state check run — and was still being reset
to pool, sending a fresh polecat through the whole cycle to reach the same step 6.
Your report is what made the fix possible: four beads in one day, plus the
reflog and reconciler-trace archaeology that proved nothing was actually lost.
That distinction — wasteful and alarming to diagnose, but not data loss — is
exactly the framing that let this land as a targeted completion path instead of
a speculative rewrite of the recovery logic.

branch.

### Original PR Review
Decision: approve (request_changes on the first pass, resolved to approve on the second).

Specific gaps fixed:
- **The completion signal was overloaded.** The original Step 3a keyed on
  `metadata.target` plus the absence of `metadata.branch_ready`. But `target`
  isn't a completion signal: `gc sling` accepts it as a documented mint-time
  input, nothing in the tree ever unsets it, and it survives both
  `mol-refinery-patrol` rejection paths — so beads that never reached step 5
  already carry it. The final patch introduces an explicit single-purpose marker,
  `handoff_stage=target_recorded`, with a deliberately narrow contract: one
  writer (step 5, in the same update that records `target`) and three clearers
  (`workspace-setup` on every fresh attempt, and both refinery rejection paths).
  Step 3a now fires only on that marker plus a live branch on origin.
- **The reassignment could report a success that never happened.** Step 3a
  writes across actors against the dead polecat's still-live `in_progress`
  claim, which `bd` refuses without `--force`. The final patch adds `--force`
  *and* checks its exit status: on failure it falls through to Step 3b having
  mutated nothing, rather than mailing `ORPHAN_HANDED_OFF` and removing the
  worktree — which would have stranded the bead `in_progress` under a dead
  assignee with its workspace gone.
- **The contract had no test.** The final patch adds a static property test
  spanning all three formulas — a marker census (one writer, three clearers, by
  cardinality), Step 3a's position between Step 3 and Step 3b, and its call
  ordering — validated by mutation rather than by passing: it goes red on
  inversion, deletion, cardinality, ordering, and on *addition*, which is the
  mutant class this repo's prose-pin suites usually miss.
- **Two smaller behavioral gaps.** `gc session wake` now precedes the nudge
  (mirroring step 7 — the refinery is an on-demand session, and nudge alone
  leaves a stopped one asleep with the handoff unprocessed), and
  `gc workflow delete-source --apply` closes the crashed polecat's workflow
  subtree inside the success arm.

Review findings addressed:
- **Behavioral Correctness** (minor, corroborated by two reviewers and reproduced
  first-hand): `git ls-remote --heads origin "$BRANCH"` matches ref *tails*, so
  the branch-on-origin backstop stayed truthy for a branch that was **not** on
  origin whenever a tail-colliding ref outlived it — reproduced live on git
  2.43.0. Fixed in `32b2339`: the query is now the fully-qualified
  `refs/heads/$BRANCH`, with the coupled test pin updated in the same commit.
- **Contract & Interface Fidelity** (minor): the agent prompt templates still
  taught the pre-fix `target` contract. Fixed in `32b2339` — the polecat contract
  table gains a `handoff_stage` row and a corrected `target` row, the refinery
  rejection sketch now shows `--unset-metadata handoff_stage`, and the witness
  prompt points at the handoff outcome.
- **Error Handling & Resilience** (nit): `gc workflow delete-source` was the one
  command in the success arm with no stated failure policy. Fixed in `32b2339`
  with `|| true` and a comment recording why best-effort is correct there (the
  reassignment has already succeeded, so a failure leaves inert step beads
  behind, never an unhandled handoff).

- `mc-5u7m9`: **Test Evidence Quality** (nit) — the `grep -F` pins certify that a
  literal is *present in the file*, not that it is *live in the executed block*,
  so commenting a pinned discriminator out keeps them green. This is the inherent
  limit of substring prose-pinning across this whole test suite, not something
  introduced here, and the line-anchored sequence pins in this PR don't share the
  blind spot. Closing it properly needs a harness that extracts and *executes*
  the formula block against a stubbed `gc` — worth doing once for the
  polecat/refinery/witness formulas together, which is why it is filed as its own
  hardening item rather than papered over with more substring pins. Not addressed
  in this PR.

### Maintainer Changes
`32b2339` — fix(gastown): gate the witness handoff on an explicit stage marker
and a checked `--force` write (7 files changed, 232 insertions(+), 42
deletions(-)). Full adopted range against the current base: 7 files changed,
246 insertions(+), 8 deletions(-).

While this PR was in review, `main` moved and independently rewrote the same
`submit-and-exit` step (#273, which made that step's `gc bd update` calls use
`--append-notes`). The two changes are additive at the same anchor, so the
conflict was resolved keeping **both** sides in full — nothing from either was
dropped — and both sides' test pins were confirmed still live on the merged
text before pushing. One sentence of this PR's prose was reworded because the
new base turned a single branch-ready halt into three; the invariant it asserts
(an unpushed branch never carries the marker) is unchanged and re-verified.
The PR description above was also rewritten to describe the marker contract that
actually landed, since the original text described the superseded
`target`/`branch_ready` design.

Your commit `1e9d702` is preserved with original authorship.

### Final Review Status
Ready for the merge queue: final head `32b2339` has passing required GitHub
checks, and the merge-ready metadata is in place. I am marking this PR
`status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/gascity-packs/actions/runs/34307491527

### Review Iterations
2 review passes performed: the first requested changes (one blocker on the
unforced, unchecked cross-actor reassignment, plus the overloaded completion
signal and the missing test), the second confirmed every item discharged by
execution — a re-run marker census, a live `ls-remote` experiment, a mutation
matrix in an isolated lab copy, and the full `gastown/tests` suites green at the
reviewed head.

---

PR #325, opened 2026-08-15
fix(gastown/witness): add --include-infra to bd list molecule queries (#324)

### LINKED ISSUE #324: witness wisp reconciliation blind: bd list hides molecule infra without --include-infra
## Bug

The witness startup protocol and `mol-witness-patrol` `next-iteration` step reconcile patrol wisps with:

```bash
gc bd list --assignee="$GC_AGENT" --status=open --type=molecule --limit=0 --json
```

`bd list` excludes `issue_type=molecule` beads as infra by default, and `--type=molecule` does **not** lift that exclusion — so this returns `[]` even when open wisps are assigned.

## Observed impact (live, 2026-08-15)

A witness patrol cycle poured a wisp at 03:55:03Z; the next witness session's startup check saw nothing and poured a second wisp → duplicate. Surplus had to be burned manually.

## Repro

```
$ bd list --status=open --type=molecule --json --limit=0 | jq length
0
$ bd list --status=open --type=molecule --include-infra --json --limit=0 | jq length
2
```

## Affected files (in `gastown/`)

- `agents/witness/prompt.template.md` — 4 `bd list ... --type=molecule` calls in the Startup Protocol (lines ~169, ~170, ~206, ~213): all missing `--include-infra`
- `formulas/mol-witness-patrol.toml` — line ~520 (`next-iteration` reconcile): missing `--include-infra`
- Same files under `examples/gastown/packs/gastown/` (if that copy is also shipped)

## Fix

Add `--include-infra` to those `bd list` calls (verified to resolve the blindness), or switch to `gc bd query --json 'status=open' | jq 'select(assignee+issue_type)'`, which sees molecules.

Found by the bloop rig's witness (bead `bl-pe8` in our city). Happy to PR if welcome.

rjgeng:
Fixes #324.

`gc bd list --type=molecule` (without `--include-infra`) hides molecule wisps by default, so the witness startup protocol and `mol-witness-patrol`'s `next-iteration` reconcile step both see `[]` even when a wisp is actually assigned. Live impact reported: a patrol cycle poured a wisp, the next session's startup check saw nothing, poured a duplicate -- the surplus had to be burned manually. Found by the reporter's own witness in production.

**Fix:** added `--include-infra` to all 5 confirmed call sites (4 in `agents/witness/prompt.template.md` at lines 169, 170, 206, 213; 1 in `formulas/mol-witness-patrol.toml`, shifted slightly to line 534 since the issue was filed against an older `main`). Confirmed all 5 sites and line numbers against current `origin/main` before editing; also checked for an `examples/gastown/packs/gastown/` mirror copy the issue flagged as "if that copy is also shipped" -- it does not exist in this repo, so no extra site there.

**Verification:** no unit-test harness exists for prompt/formula shell content specifically, so verification was exact-match grep against the issue's report before editing (confirmed the bug pattern present), then grep-confirmed all 5 sites carry `--include-infra` after. `gastown/tests/test_gastown_pack_assets.sh` and `tests/test_no_bare_bd_commands.py` pass. Full pytest suite (`tests/`, `gastown/tests/`, `gascity/tests/`): 285 passed, 8 skipped (pre-existing), 9435 subtests, no regressions. `gastown/tests/test_witness_heartbeat_check.sh` fails, but reproduces identically on unmodified `origin/main` (GNU `date -d` vs macOS BSD `date`) -- confirmed pre-existing and unrelated.

**Flagging, not fixed here:** the same `--type=molecule` pattern (without `--include-infra`) also appears in `agents/refinery/`, `agents/deacon/`, `formulas/mol-deacon-patrol.toml`, and `formulas/mol-refinery-patrol.toml`, all on `--status=in_progress` queries. Could not get a live Dolt server running locally to confirm those hit the same infra-exclusion path this issue's `--status=open` repro demonstrated, so left untouched rather than guess. If `bd list --status=in_progress --type=molecule --json` (no `--include-infra`) on a live rig comes back empty when it shouldn't, that's the same bug hitting deacon/refinery's "am I mid-wisp" resume check too, and would be a larger follow-up fix.

---

### DIFF
--- a/gastown/agents/witness/prompt.template.md
+++ b/gastown/agents/witness/prompt.template.md
@@ -166,8 +166,8 @@ nothing).
 # burn the surplus so restarts never accumulate duplicates. Wisp roots are
 # molecules — filter --type=molecule, never --type=wisp.
 WISP_IDS=$(
-  gc bd list --assignee="$GC_AGENT" --status=in_progress --type=molecule --limit=0 --json | jq -r '.[].id'
-  gc bd list --assignee="$GC_AGENT" --status=open --type=molecule --limit=0 --json | jq -r '.[].id'
+  gc bd list --assignee="$GC_AGENT" --status=in_progress --type=molecule --include-infra --limit=0 --json | jq -r '.[].id'
+  gc bd list --assignee="$GC_AGENT" --status=open --type=molecule --include-infra --limit=0 --json | jq -r '.[].id'
 )
 WISP=$(printf '%s\n' $WISP_IDS | sed -n '1p')           # keep one (prefers in_progress)
 for extra in $(printf '%s\n' $WISP_IDS | sed '1d'); do  # burn any surplus
@@ -203,14 +203,14 @@ If `next-iteration` already ran, do not pour again; run `gc hook`.
 ```bash
 CURRENT_WISP=${GC_BEAD_ID:-}
 if [ -z "$CURRENT_WISP" ]; then
-  CURRENT_WISP=$(gc bd list --assignee="$GC_AGENT" --status=in_progress --type=molecule --limit=1 --json | jq -r '.[0].id // empty')
+  CURRENT_WISP=$(gc bd list --assignee="$GC_AGENT" --status=in_progress --type=molecule --include-infra --limit=1 --json | jq -r '.[0].id // empty')
 fi
 # Reconcile queued (open) patrol wisps to exactly one. A prior cycle may have
 # poured a next wisp without burning, or a restart may have raced — keep the
 # first and burn the surplus so wisps never accumulate. Wisp roots are
 # molecules (never --type=wisp, which is not a valid gc bd type and matches
 # nothing).
-OPEN_WISPS=$(gc bd list --assignee="$GC_AGENT" --status=open --type=molecule --limit=0 --json | jq -r '.[].id')
+OPEN_WISPS=$(gc bd list --assignee="$GC_AGENT" --status=open --type=molecule --include-infra --limit=0 --json | jq -r '.[].id')
 ASSIGNED_WISP=$(printf '%s\n' $OPEN_WISPS | sed -n '1p')
 for extra in $(printf '%s\n' $OPEN_WISPS | sed '1d'); do
   gc bd mol burn "$extra" --force
--- a/gastown/agents/witness/prompt.template.md
+++ b/gastown/agents/witness/prompt.template.md
@@ -164,7 +164,10 @@ nothing).
 # Step 1: Reconcile your patrol wisps to exactly one (town ledger, via gc bd).
 # Collect every open/in_progress patrol wisp assigned to you, keep one, and
 # burn the surplus so restarts never accumulate duplicates. Wisp roots are
-# molecules — filter --type=molecule, never --type=wisp.
+# molecules — filter --type=molecule, never --type=wisp. They are also
+# ephemeral, so they live in the wisps tier that gc bd list hides without
+# --include-infra; drop that flag and these queries return [] even when a
+# wisp is assigned, and you pour a duplicate.
 WISP_IDS=$(
   gc bd list --assignee="$GC_AGENT" --status=in_progress --type=molecule --include-infra --limit=0 --json | jq -r '.[].id'
   gc bd list --assignee="$GC_AGENT" --status=open --type=molecule --include-infra --limit=0 --json | jq -r '.[].id'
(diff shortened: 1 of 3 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277252206

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277252434

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277288133

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277288206

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277308574

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @rjgeng! This PR fixes a real, quietly-destructive
bug: the witness patrol's wisp-reconciliation queries called
`gc bd list --type=molecule` without `--include-infra`, and `bd` hides the infra
tier unless that flag is passed (`cmd/bd/list_filter.go` sets `SkipWisps=true`
by default, independent of `--status`). So the reconcile query returned `[]`
*even when a wisp was already assigned*, and the witness poured a duplicate —
one leaked wisp per restart. Five lines of flag, and a recurring resource leak
in the patrol loop goes away. Finding the omission at the query sites rather
than chasing the duplicate wisps downstream is the part that made this easy to
adopt.

branch.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- **The fix was correct but unexplained.** The edited sites carried no note of
  *why* the flag is required, which is how the omission got introduced in the
  first place. All three edited blocks now state the mechanism (wisps are
  ephemeral and live in a tier `gc bd list` hides without the flag) and the
  consequence (query returns `[]`, duplicate wisp poured).
- **Nothing pinned the flag.** The behaviour had regressed once already, so a
  witness-scoped regression test now guards all five query sites. It was
  measured red under every regression mode: flag stripped from a prompt query,
  flag stripped from the formula query, and a query site deleted.
- **The original commit message's scoping rationale was inaccurate.** It has
  been corrected in the permanent record, with the `list_filter.go` source trace
  standing in for the live repro, and the real reason for the witness-only
  boundary (scope discipline, plus the broader #252) restated.

Review findings addressed:
- *The pin test's scoping comment deferred to an unnamed tracker*, making the
  deferral unauditable where someone would actually edit the test. Fixed in
  `8aab19a` — the comment now names #252.
- *The count assertion `[[ "$total" -eq 5 ]]` pinned query cardinality rather
  than the contract*, so a legitimate sixth witness query, or a comment reflow
  landing the counted tokens on one line, would have failed the suite with
  nothing wrong. Fixed in `8aab19a`, relaxed to `-ge 5`. This was picked over
  the alternative proposal on measurement: a 6×3 matrix over six tree states
  showed `-ge 5` keeps every regression mode red (a deleted query drops the
  count to 4, which still fails) while dropping both false positives. The
  `flagged -eq total` assertion — the one that actually guards the contract —
  was left untouched.

- `mc-f6fo1`: two full-line witness-prompt literals in the downstream `gc` repo
  (`examples/gastown/gastown_test.go:2881-2882`) stop matching once this pack is
  released, because the flag is inserted mid-line. Not fixable here, and that is
  a measured claim rather than a judgement call: every syntactically valid
  placement of the flag on that line was tested and all five break a whole-line
  pin. The repin belongs in the other repo and is safest done by relaxing the
  literals to the pre-flag prefix, which matches what the surviving coarser pins
  there already do. An exhaustive scan (74 candidate files, ~11.9k literals)
  confirms those two lines are the complete blast radius.
- `mc-qjnu8`: on a rig running a `gc`/`bd` build predating `--include-infra` on
  `bd list`, these queries now fail loudly on an unknown flag instead of
  silently returning `[]`. A strictly better failure mode, noted as a rollout
  check rather than a defect.
- #252: the sibling refinery and deacon patrol queries still carry the bare
  form (re-measured at this head: 5 flagged, 11 bare). That gap is pre-existing
  and unregressed by this PR, and keeping the fix witness-scoped was the
  unanimous recommendation across both review rounds. I have commented on #252
  to record that #324 is its partial duplicate and that this PR closes its
  witness rows, so the remaining scope is unambiguous.

### Maintainer Changes
One maintainer commit, `8aab19a` *fix(gastown/witness): explain and pin
`--include-infra` on the wisp queries* — 3 files changed, 44 insertions(+),
3 deletions(-). It adds the rationale comments at the edited sites and the
witness-scoped pin test; it changes no executable behaviour of the fix itself.
Your original commit is preserved as `f21c3c5`, rebased onto current `main`.

### Final Review Status
Ready for the merge queue: final head `8aab19a` has passing required GitHub
checks (Check, Analyze (go), Analyze (python), CodeQL), and the merge-ready
metadata is in place. The pack asset suite, `tests/test_no_bare_bd_commands.py`
and `tests/test_gascity_pack_inference_gate.py` were also re-run locally on the
rebased tree after `main` moved. I am marking this PR `status/merge-ready` so
the merge queue can pick it up.

CI: https://github.com/gastownhall/gascity-packs/actions/runs/34199636239

### Review Iterations
2 review passes performed: the first accepted the fix and asked for the two
cleanups above; the second confirmed both landed as scoped, independently
re-derived the underlying `bd` behaviour from source rather than taking the
PR's premise on trust, and re-measured every verification claim.

---

PR #327, opened 2026-08-17
compound-implementation: carry the coordinator steps directly instead of extending implement

swedeinasia-flow:
## Problem

The compound-engineering pack is half-migrated off the base `implement` lane: `compound-work` and `compound-work-item` already extend `do-work` / `do-work-item`, but the `compound-implementation` coordinator still declares `extends = ["implement"]`.

Deployments that have retired `implement` from their catalogs cannot compile any Compound Engineering lane through it. (Context: bare `implement` defaults `push=false` / `open_pr=false`, so slung work closes `pass` with nothing published — a stranding trap that led at least one production city to tombstone the formula at the compiler gate. Any formula extending it is then refused, which currently blocks compound-engineering trials there.)

## Change

Make `compound-implementation` self-contained: inline the five inherited coordinator pieces (launch-input vars and the `prepare`, `wait-for-drain`, `summarize`, `publish` steps) verbatim from `implement`, with the step prompts copied into this pack's assets so `description_file` paths stay pack-relative. The two drain steps were already fully overridden here, so this drops the last dependency on `implement`.

Behavior is byte-identical for deployments that still ship `implement`.

## Verification

Fresh throwaway city importing this branch by sha alongside the `gascity` base pack:

- `gc formula show compound-implementation --json` — compiles, 9 steps
- `gc formula show compound-build --json` — compiles, **106 steps, unchanged** from the pre-change compile of v0.4.0
- `gc formula show compound-work / compound-work-item` — compile (7 / 5 steps)
- cached pack content confirms no `extends` remains in `compound-implementation.formula.toml`

### DIFF
--- a/compound-engineering/assets/workflows/compound-implementation/prepare.md
+++ b/compound-engineering/assets/workflows/compound-implementation/prepare.md
@@ -1,0 +1,33 @@
+
+Validate graph.v2 target state before side effects.
+
+This step is validation only. Do not edit source files in the launcher checkout,
+any existing worktree, or any newly created path. Do not create, modify, or commit source code. Do not run implementation or test-fix loops.
+
+Requirements:
+- identify the current claimed step bead from `CLAIMED_BEAD_ID` in the startup
+  claim output, or `$GC_BEAD_ID` if the claim output did not expose one;
+  hard-fail if neither value is present
+- read the claimed step JSON with `gc bd show "$CLAIMED_BEAD_ID" --json`, then
+  read the current workflow root id from `metadata["gc.root_bead_id"]`
+- read the current workflow root JSON with `gc bd show "$ROOT_ID" --json`
+- resolve the implementation input convoy from the current workflow root
+  metadata key `gc.input_convoy_id`
+- verify the resolved input convoy id matches rendered runtime convoy
+  `{{convoy_id}}`
+- validate that input bead is a convoy or normalized singleton convoy with
+  `gc bd show "<gc.input_convoy_id>" --json` before treating it as implementation
+  work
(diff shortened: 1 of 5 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277252060

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277252129

swedeinasia-flow:
I re-baselined the failing `Check` job against the current repository policy. The three failures are substantive, not transient:

- third-party implementation wrappers are required to declare `extends = ["implement"]`;
- derived-pack prompt assets may not dispatch through `{{pack_root}}/assets/scripts`;
- the resolved methodology contract must preserve the adapter-selectable implementation stage.

This PR intentionally makes `compound-implementation` self-contained and carries coordinator steps directly, so merely rebasing or editing around the assertions would either restore the behavior this PR set out to replace or weaken a repository-wide compatibility contract.

I am therefore not pushing a cosmetic CI workaround. Maintainer direction would help choose the honest next step:

1. keep Compound Engineering as a third-party methodology adapter extending `implement`, in which case this self-contained variant should remain downstream or be retired; or
2. establish a new first-class contract/formula shape for self-contained coordinator steps, then rework this PR against that agreed boundary and update the compatibility tests deliberately.

I can take either path once the intended architecture is clear.

_Codex-unknown-model-unknown-reasoning on behalf of Klas Hesselman_

julianknutsen:
The intention here is to demonstrate how to extend the common base implementation to meet advanced scenarios. If there is a bug, let's fix that instead of making it hermetic.

Issue #307, opened 2026-08-12
10 of 16 registry packs are never exercised against a running gc, and the registry gives users no way to tell


sjarmak:
A user reported that they set up `oversight-rig`, `slack-full`, and `github`
in their city and "not a single one of them was able to land without source
modifications," while reasonably assuming that packs this visible could not be
broken out of the box. Looking at what CI actually covers, the assumption is
the problem rather than their setup.

### 10 of the 16 registry packs are never exercised against a running `gc`

`ci.yml` installs `gc` and lints a fixed list:

```yaml
go install github.com/gastownhall/gascity/cmd/gc@latest
for pack in gascity gascity/roles bmad compound-engineering gstack superpowers profiler; do
  "$GC_BIN" lint "$pack"
done
```

`supported-pack-nightly.yml` exercises another fixed list: `gascity`,
`superpowers`, `compound-engineering`, `gstack`, `bmad`, `gastown`.

Everything else in `registry.toml` is outside both:

```
cass, contributing, discord, github, oversight-rig,
pr-pipeline, runtime-cloudflare, slack-channel, slack-full, slack-mini
```

All three packs in the report are in that list.

This is not the same as untested. `github`, `slack-full`, `slack-channel`,
`discord`, and `pr-pipeline` have pytest suites that CI runs, and the Slack and
cloudflare packs have Go test suites. Those suites test the pack's own logic
against its own fixtures. Nothing runs `gc lint` on them, and nothing stands
them up against a gascity that has moved since the pack was last touched.

`github` was last modified 2026-07-13. `oversight-rig` landed 2026-06-20 with
one addition on 2026-08-08. gascity main moves daily. Nothing in either repo
detects the drift in between.

### The registry does not tell users which is which

`registry.toml` entries carry `name`, `description`, `source`, `source_kind`,
`ref`, `commit`, `hash`, `schema`, `version`. There is no field for support
tier, last-validated gascity version, or known prerequisites. A user reading
the registry sees `slack-full` and `bmad` described identically as far as
support is concerned. One is exercised nightly and one is not exercised at all.

Since the concept of a supported pack already exists (the nightly workflow's
input is literally named `all-supported`), the gap is that the concept is not
expressed anywhere a user looks.

### Two smaller things found in the same pass

`go install ...@latest` resolves to the latest release tag, currently v1.4.0,
not main. Even the validated packs are validated against a tag while users
building from main run something newer.

[...]

Issue #137, opened 2026-06-24
Can't import packs


johnstegeman:
This happens for any pack install

gc import add https://github.com/gastownhall/gascity-packs.git//superpowers

`
gc import add "https://github.com/gastownhall/gascity-packs.git//superpowers": import "pack:superpowers": cached pack "https://github.com/gastownhall/gascity-packs.git//superpowers" is missing pack.toml at /Users/jstegeman/.gc/cache/repos/385410dbb16b5d87f95aedc0733e07c3044a0feadd8e068de0e0136c5cadb2a9/superpowers/pack.toml
`

esciara:
Confirmed — same root cause, reproduced independently (I hit it on the `contributing` pack).

`gc import add … //superpowers` resolves the version against the repo's **git tags**, not the registry's per-pack releases. `gastownhall/gascity-packs` has only two tags (`v0.2.0`, `v0.3.0`); the resolver picks the highest one (`v0.3.0`), but `superpowers/` was added to `main` **after** that tag — `superpowers/pack.toml` is HTTP **404 at `v0.3.0`** and **200 at `main`** — so the checked-out tree has no `pack.toml`, validation fails, and the cache dir is then deleted (which is why the reported cache path doesn't exist afterward).

**Workaround** — pin a `main` commit that actually contains the pack, which bypasses tag resolution:

```bash
# current main commit
git ls-remote https://github.com/gastownhall/gascity-packs.git main
```

```toml
# pack.toml
[imports.superpowers]
source  = "https://github.com/gastownhall/gascity-packs//superpowers"
version = "sha:<that-commit>"
```

then `gc import install`. (For packs that are in the registry, `gc pack registry show <pack>` prints the release commit to use.)

Full root-cause writeup filed against the CLI repo: gastownhall/gascity#3710.

johnstegeman:
Thank you - testing now!

jomadu:
Same failure as this issue, on a different pack (`gascity/roles`). The CLI-side root cause is tracked in gastownhall/gascity#3710 (fix in gastownhall/gascity#3731): version resolution picks the newest repo git tag instead of the registry release, and the newest tag predates the pack.

[...]

PR #329, opened 2026-08-17
fix: five reported pack defects, and the green suites that hid them

### LINKED ISSUE #63: Fix manifest/app.json: empty slash_commands rejected by Slack, missing app_mentions:read scope
## Problem

`slack-full/manifest/app.json` has two bugs that cause Slack app manifest import to fail or the app to malfunction:

1. **`features.slash_commands: []`** — Slack's manifest validator rejects an empty array here. Slack requires either ≥1 item or the key omitted entirely. Importing the manifest as-is fails with a validation error.

2. **Missing `app_mentions:read` scope** — The manifest subscribes to the `app_mention` bot event (`settings.event_subscriptions.bot_events`) but does not include the `app_mentions:read` OAuth scope in `oauth_config.scopes.bot`. Slack rejects the manifest because event subscriptions require the corresponding scope to be present.

## Fix

- Drop `slash_commands` from `features` entirely (or remove the empty array key).
- Add `"app_mentions:read"` to `oauth_config.scopes.bot`.

Corrected `app.json` (minimal diff):

```json
// REMOVE:
"slash_commands": [],

// ADD to oauth_config.scopes.bot:
"app_mentions:read",
```

Both fixes are confirmed against the Slack API manifest validation endpoint.

## Context

Discovered while migrating a gas-city deployment to the slack-full pack (Tier 3). The corrected manifest was used successfully to create an app in the gascityone workspace.

### LINKED ISSUE #263: test: claim-command fixtures leak ambient GC_TEMPLATE — both test_city_claim_command_* fail when the suite runs inside a gc session (green on CI)
## Summary

`gascity/tests/test_formula_assets.py` builds subprocess environments from `{**os.environ}` and defensively clears **four** of the session variables `gascity/commands/claim/run.sh` reads, but not all of them. `GC_TEMPLATE` is missed, so both `test_city_claim_command_*` tests **fail whenever the pack suite is run from inside a gc-managed session** and pass on a clean runner.

CI never sets `GC_TEMPLATE`, so CI is green and this is invisible there. Agents run pack tests from inside gc sessions, which is the normal case rather than the exception — so the tests are red exactly where they are most often run.

## Reproduction

Bare checkout of `main` (verified at `0b95742`), no local changes:

```
$ cd gascity && python3 -m unittest discover -s tests
FAIL: test_city_claim_command_verifies_and_normalizes_claim
FAIL: test_city_claim_command_declares_authoritative_convoy_source_on_explicit_run
Ran 176 tests — FAILED (failures=2)

$ env -u GC_TEMPLATE -u GC_AGENT python3 -m unittest discover -s tests
Ran 176 tests — OK
```

Reproduced independently from two different sessions with different values (`GC_TEMPLATE=mayor` and `GC_TEMPLATE=adjunct`), failing identically — so it is any leaked value, not one specific to a session.

## Cause

```sh
# gascity/commands/claim/run.sh:113
EXPECTED_ROUTE="${GC_TEMPLATE:-${GC_AGENT:-}}"
```

`GC_TEMPLATE` takes precedence over `GC_AGENT`. The fixtures override `GC_AGENT` but never `GC_TEMPLATE`, so an ambient `GC_TEMPLATE` wins and `EXPECTED_ROUTE` becomes the *host session's* template instead of the fixture's `gc.implementation-worker`. `run.sh:215-217` then correctly reports:

```
CLAIM_REJECTED route mismatch for bd-123
```

**`run.sh` is behaving as designed — the fixture is wrong.** No product defect here.

## The underlying issue is the clear-list, not the one variable

The fixtures at `test_formula_assets.py:985` and `:1052` already do:

```python
for key in ("BEADS_ACTOR", "GC_SESSION_NAME", "GC_SESSION_ID", "GC_AGENT"):
    env.pop(key, None)
```

So ambient leakage was a known hazard and a defence exists — it is just incomplete, and nothing keeps it in sync with what `run.sh` actually consumes.

`run.sh` reads eight session variables:

| variable | guarded? | how |
|---|---|---|
| `BEADS_ACTOR` | yes | cleared |
| `GC_SESSION_NAME` | yes | cleared |
| `GC_SESSION_ID` | yes | cleared |
| `GC_AGENT` | yes | cleared |
| `GC_PACK_DIR` | yes | set explicitly |
| `GC_PACK_NAME` | yes | set explicitly |
| **`GC_TEMPLATE`** | **no** | **leaks — this bug** |
| **`GC_BEADS_PROJECT_ID`** | **no** | **leaks — latent** |

`GC_BEADS_PROJECT_ID` is the same bug waiting to happen: `run.sh:271-273` branches on it (`WORK_REF_SKIPPED explicit run has no GC_BEADS_PROJECT_ID`), so an ambient value silently changes which path the test exercises rather than failing loudly.

Repo-wide there are **16 `{**os.environ}` sites across 2 files** (`gascity/tests/test_formula_assets.py`, `tests/test_gc_role_prompt_integration.py`), and **neither file mentions `GC_TEMPLATE` anywhere**.

## Suggested fix

The one-line patch is adding `GC_TEMPLATE` to the clear-list, but that leaves the class open — the next variable `run.sh` learns to read reintroduces it silently.

Two better options:

1. Build the subprocess env from a **minimal explicit dict** (`PATH`, `HOME`, plus exactly what the fixture intends) instead of `{**os.environ}`. Strongest, and makes intent explicit.
2. Clear **all** `GC_*` and `BEADS_*` variables, then set what the fixture needs:
   ```python
   env = {k: v for k, v in os.environ.items()
          if not k.startswith(("GC_", "BEADS_"))}
   env.update({...})
   ```

Either removes the need to keep a hand-maintained list in sync with `run.sh`.

## Notes

Found while adopting #149/#169 onto a fork. Not urgent — no product impact, CI unaffected — but it cost us real time: the red tests initially read as "upstream ships a broken agent-facing command", and `commands/` is auto-discovered by convention, so that misreading nearly led to excluding `gascity/commands/claim` from a downstream pack pin. A green-on-CI / red-in-session split is an expensive shape for exactly that reason.

sjarmak:
Closes #263.

The `GC_TEMPLATE`/`commands/claim/run.sh` mismatch this fixes is the cause of the two `test_formula_assets.py::...city_claim_command...` failures reported there.

---

Five user-reported defects in the shipped packs, and the reason each one's
test suite stayed green while the pack was broken.

Closes #63. Addresses #167. Addresses #307.

Issue #137 is already resolved by the v0.4.0 release; its guard here is regression coverage, not the original fix.

## What was broken

**#63, slack-full.** A user migrating to `slack-full` could not import
`manifest/app.json`. Slack rejected it on two rules: `features.slash_commands`
was `[]` (Slack requires the key absent or non-empty), and the manifest
subscribed to `app_mention` without declaring `app_mentions:read`. The pack's
own `test_manifest.py` asserted `slash_commands` had to exist as a list, so the
test held the defect in place. `manifest/agent-app.json` carried the same empty
array and was not in the report because nobody had tried to import it yet.

**#167, deprecated contract key.** 81 formulas declared
`contract = "graph.v2"`, which raises one blocking `gc doctor` warning per
formula in every consuming city. Measured against a city importing eleven packs
and 97 formulas: 79 warnings before, 0 after. This is a rename, not a behavior
change. `directFormulaCompilerConstraints` maps the legacy field to exactly the
constraint `[requires] formula_compiler = ">=2.0.0"` produces, and consumers key
off `UsesGraphCompiler` over the resulting constraint set rather than the
literal field, so the `gc.formula_contract=graph.v2` bead stamp is untouched.

**#137, registered packs absent from every release tag.** `gc import add
<repo-url>//<pack>` with no version constraint resolves against this
repository's git tags, not `registry.toml`. A pack added to `main` with no newer
tag cut resolves to the newest existing tag, finds no `pack.toml`, fails
validation, and deletes the cache directory it names in the error. The user sees
`cached pack ... is missing pack.toml at <path that no longer exists>`. Eight
registered packs were in that state for weeks. `v0.4.0` closed the gap; the new
test is what keeps the next pack from reopening it.

**The migration to `[requires]` created a version floor nobody declared.** gc
only learned to read that table in v1.3.0 (`Requires *Requirements` first
appears in `internal/formula/types.go` at that tag, and is absent at v1.2.1).
Older gc parses formula TOML leniently, so the table is not rejected there, it
is dropped: a formula using graph-only constructs then fails to compile, and one
that does not silently takes the v1 path, losing graph validation and the bead
stamp. Nine packs now declare `requires_gc = ">=1.3.0"`.

**Be precise about what `requires_gc` buys, because the key looks stronger than
it is.** gc parses it into `config.Pack.RequiresGC` and copies it through `gc
init`, and nothing compares it against the running version:

    grep -rn 'RequiresGC' --include=*.go <gascity checkout>/internal <gascity checkout>/cmd

Every hit is a declaration, a round-trip test, or a copy. So the key is
accurate, greppable, and in the slot designed to hold it, and it starts working
the day a consumer lands. It is not a guarantee gc enforces today. The new test
says that in its docstring rather than implying otherwise. What is enforced is
the coupling inside this repo: a pack shipping a formula that selects the v2
compiler without declaring the floor fails.

**#307, partial: the suites were green for a reason unrelated to the code.**
Running the CI pytest line on a machine with a real Gas City installed produced
14 failures; on a clean CI runner the same line is green. A live Gas City seat
exports `GC_API_BASE_URL`, `GC_TEMPLATE`, `GC_CITY_PATH` and friends into every
process it spawns, and the packs read those ahead of the fixture's own
`city.toml`. Each `setUp` now strips the prefixes its pack reads before pinning
the fixture. `oversight-rig/tests` existed on disk and was missing from the CI
pytest step, so it had never run; added, with a test that keeps the explicit
list in step with the tree.

## Why the tests are evidence about a live build

Each guard below replaces one seam, and each replacement is stated with what it
can and cannot tell you. The standard is in
`pr-pipeline/docs/testing-across-the-layer-boundary.md`, added in this branch:
a fix's test fails on the unfixed live build and passes on the fixed one, with
the fix as the only difference.

| Guard | Seam replaced | Fidelity, and where it stops |
|---|---|---|
| `test_slack_manifest_conformance.py` | none. Reads the real shipped manifests from the tree, discovered by content (any `*.json` with an `oauth_config` key) rather than a path list | The rules are transcribed from the behavior a user's failed import exhibited, not from Slack's published validator. A rule Slack enforces and this file does not is invisible here |
| `test_registry_packs_are_in_a_release_tag.py` | none. Real `registry.toml`, real `git tag`, real `git cat-file -e` against the tag | A fixture version of this test would have passed on every day the repository was broken. It needs tags fetched, so it fails rather than skips on a tagless checkout |
| `test_formula_contract_declaration.py`, `test_pack_gc_version_floor.py` | `UsesGraphCompiler`, reimplemented in `scripts/formula_compiler_requirement.py` | Mirrors `declaresGraphCompilerRequirement`: a walk over the constraint list, true on the first constraint that excludes capability 1.0.0 and admits 2.0.0. Not a Masterminds semver implementation. Forms it does not model raise rather than evaluate, so a new constraint shape fails loudly. Prereleases are refused, not truncated, because a three-integer tuple cannot order `1.0.0-alpha` below `1.0.0` |
| env-prefix guard in `github/tests`, `discord/tests`, `gascity/tests` | "run the module and see what it reads", replaced by an AST enumeration over the guarded modules | Re-derives the variable names from the module ASTs at test time. A hand-list is what drifted the first time. Keys it cannot resolve to a literal are reported, not dropped, so an unresolvable read fails the guard instead of shrinking it |

Verified against the live binary rather than only the fixtures:

- 34 formulas compiled through the real `gc` binary against a throwaway city,
  before and after the #167 migration. 34 identical, 0 differing, 0 failures.
- `gc doctor` on a city importing eleven packs and 97 formulas: 79 blocking
  warnings before, 0 after, with the discovery control that reverting one file
  turns it red again.
- `gc lint` on all nine packs is unchanged by this branch. Eight report ok.
  `gastown` reports the same 14 findings on `upstream/main` and on this branch
  (five `named_session` targets a pool-controlled agent, nine `bd-unknown-flag`
  hits in prompt templates), none mentioning `requires_gc`. Commit 723e950's
  message says eleven; that count was wrong, and 14 is the measurement on both
  trees. Refute with `gc lint gastown | wc -l` from each.

Those 14 `gastown` findings are pre-existing and are not fixed here, because
they are a change to a different pack and belong in their own PR. `gastown` is
in `supported-pack-nightly.yml`, so this is not the uncovered-pack case in
#307, it is a supported pack drifting from the binary with nothing failing.

**Correction, added after this PR was opened.** The paragraph above originally
said all nine `bd-unknown-flag` findings "are real". That was asserted from the
flag list, not measured per finding, and it is wrong. Working through them one
by one in #330:

- Four are real flag errors at real invocation sites. `bd create` takes `-l,
  --labels`, not `--label`; `bd create` has no `--prefix`.
- Three are the linter flagging a documented anti-example: two strikethrough
  cells in a "Common mistake" column, and one line of prose forbidding a flag.
  Those are still worth fixing, because an anti-example built on a flag that
  does not exist teaches the wrong lesson twice, but they were never going to
  be run.
- Two are a defect in the linter, not the pack. `internal/bdflags` keys top out
  at two words, so `bd mol wisp gc --age` matches "mol wisp", reads `gc` as a
  positional, and validates `--age` against the parent instead of the
  subcommand that owns it. `bd mol wisp gc --help` documents `--age` in its own
  examples.

The five `named_session` findings are also linter defects, for a separate
reason set out in #330 and in `tests/gastown_lint_upstream_defects.txt`.

#330 fixes the seven that are pack defects and pins the seven that are not.

## Test plan

    python3 -m pytest -q            # 1372 passed, 8 skipped

Every guard added here is proven in both directions rather than asserted to
work. Red on each of these mutations, green again on restore:

- strip the `[requires]` declaration from a pinned formula
- drop a line from `tests/graph_v2_formulas.txt`
- widen the version regex to truncate prereleases
- evaluate the comma-joined constraint string instead of the list
- restore the OR short-circuit in the constraint parser
- weaken a pack's `requires_gc` to `">=1.2.0"` (and it stays green on
  `">=1.4.0"`, which is a legitimately stricter answer)
- revert either Slack manifest
- drop `GC_` or `GITHUB_INTAKE_` from a suite's strip list
- blind the environ-alias tracker, or bucket AST call sites by the first module

`test_formula_contract_declaration.py` pins the declaring formulas as a set in
`tests/graph_v2_formulas.txt` rather than a floor on the count. A count holds
while one formula loses its declaration and an unrelated one gains it, which is
the realistic shape of the regression: a migration touches many files and misses
one.

## Known and deliberate

A diamond `extends` chain appends the shared grandparent's constraint twice,
here and in gc. Neither deduplicates, and the predicate is an OR over the list,
so the repeat cannot change the answer. Left as parity rather than fixed on this
side, which would make the mirror wrong in a new way.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,7 +41,14 @@ jobs:
         run: python3 validate_registry.py --require-git
 
       - name: Run Python pack tests
-        run: python3 -m pytest tests contributing/tests gascity/tests discord/tests github/tests slack-full/tests slack-channel/tests pr-pipeline/tests profiler/tests -q
+        # oversight-rig/tests existed on disk and was absent from this list, so
+        # its tests had never run in CI. The list stays explicit (pytest
+        # discovery from the root would sweep vendored and fixture trees), and
+        # tests/test_ci_covers_every_test_directory.py fails when a */tests
+        # directory holding test_*.py is missing from it. gastown/tests is
+        # deliberately absent: it holds shell tests only, run by the
+        # "Run Gastown shell tests" step below.
+        run: python3 -m pytest tests contributing/tests gascity/tests discord/tests github/tests oversight-rig/tests slack-full/tests slack-channel/tests pr-pipeline/tests profiler/tests -q
 
       - name: Lint and exercise shared role prompt composition
         run: |
--- a/bmad/formulas/bmad-build.formula.toml
+++ b/bmad/formulas/bmad-build.formula.toml
@@ -7,9 +7,11 @@ epics/stories so it matches the upstream workflow.
 """
 formula = "bmad-build"
 version = 1
-contract = "graph.v2"
 extends = ["build-base"]
 target_required = true
+
+[requires]
+formula_compiler = ">=2.0.0"
 
 [catalog]
 name = "bmad-build"
--- a/bmad/formulas/bmad-code-review-flow.formula.toml
+++ b/bmad/formulas/bmad-code-review-flow.formula.toml
@@ -7,7 +7,9 @@ review lanes, including gap analysis, with synthesis and fix application.
 formula = "bmad-code-review-flow"
 type = "expansion"
 version = 1
-contract = "graph.v2"
+
+[requires]
+formula_compiler = ">=2.0.0"
 
 [vars]
 [vars.implementation_target]
--- a/bmad/formulas/bmad-decomposition.formula.toml
+++ b/bmad/formulas/bmad-decomposition.formula.toml
@@ -3,10 +3,12 @@ BMAD implementation of the decomposition-base methodology contract.
 """
 formula = "bmad-decomposition"
 version = 1
-contract = "graph.v2"
 extends = ["decomposition-base"]
 target_required = false
 internal = true
+
+[requires]
+formula_compiler = ">=2.0.0"
 
 [[steps]]
 id = "decompose"
--- a/bmad/formulas/bmad-fix-loop.formula.toml
+++ b/bmad/formulas/bmad-fix-loop.formula.toml
@@ -3,10 +3,12 @@ BMAD implementation of the fix-loop-base methodology contract.
 """
 formula = "bmad-fix-loop"
 version = 1
-contract = "graph.v2"
 extends = ["fix-loop-base"]
 target_required = false
 internal = true
+
+[requires]
+formula_compiler = ">=2.0.0"
 
 [vars]
 [vars.implementation_formula]
--- a/bmad/formulas/bmad-implementation.formula.toml
+++ b/bmad/formulas/bmad-implementation.formula.toml
@@ -3,10 +3,12 @@ BMAD implementation of the public implement methodology contract.
 """
 formula = "bmad-implementation"
 version = 1
-contract = "graph.v2"
 extends = ["implement"]
 target_required = true
 internal = true
+
+[requires]
+formula_compiler = ">=2.0.0"
 
 [vars]
 [vars.implementation_target]
--- a/bmad/formulas/bmad-planning.formula.toml
+++ b/bmad/formulas/bmad-planning.formula.toml
@@ -3,10 +3,12 @@ BMAD implementation of the planning-base methodology contract.
 """
 formula = "bmad-planning"
 version = 1
-contract = "graph.v2"
 extends = ["planning-base"]
 target_required = false
 internal = true
+
+[requires]
+formula_compiler = ">=2.0.0"
 
 [[steps]]
 id = "requirements"
--- a/bmad/formulas/bmad-review.formula.toml
+++ b/bmad/formulas/bmad-review.formula.toml
@@ -3,11 +3,13 @@ BMAD implementation of the code-review-base methodology contract.
 """
 formula = "bmad-review"
 version = 1
-contract = "graph.v2"
 extends = ["code-review-base"]
 target_required = false
 internal = true
 mode = "report"
+
+[requires]
+formula_compiler = ">=2.0.0"
 
 [vars]
 [vars.implementation_target]
--- a/bmad/formulas/bmad-story-development-item.formula.toml
+++ b/bmad/formulas/bmad-story-development-item.formula.toml
@@ -6,11 +6,13 @@ shared drain lifecycle and serializes item work.
 """
 formula = "bmad-story-development-item"
 version = 1
-contract = "graph.v2"
 extends = ["do-work-item"]
 target_required = true
 internal = true
 single_lane = true
+
+[requires]
+formula_compiler = ">=2.0.0"
 
 [[steps]]
 id = "implement-item"
--- a/bmad/formulas/bmad-story-development.formula.toml
+++ b/bmad/formulas/bmad-story-development.formula.toml
@@ -6,9 +6,11 @@ implementation and review lanes.
 """
 formula = "bmad-story-development"
 version = 1
-contract = "graph.v2"
 extends = ["do-work"]
 target_required = true
+
+[requires]
+formula_compiler = ">=2.0.0"
 
 [[steps]]
 id = "implement"
--- a/bmad/pack.toml
+++ b/bmad/pack.toml
@@ -2,6 +2,14 @@
 name = "bmad"
 version = "0.1.0"
 schema = 2
+# Minimum gc version. The formulas in this pack declare
+# [requires] formula_compiler, a table gc only learned to read in v1.3.0;
+# older gc parses the formula leniently and drops the declaration, so a
+# formula using graph-only constructs fails to compile and one that does
+# not silently takes the v1 path. gc parses this key today and does not
+# yet enforce it -- it is the declared floor, not a guarantee.
+requires_gc = ">=1.3.0"
+
 
 [imports.gc]
 source = "../gascity"
--- a/compound-engineering/formulas/compound-build.formula.toml
+++ b/compound-engineering/formulas/compound-build.formula.toml
@@ -7,9 +7,11 @@ is a finalize-stage behavior.
 """
 formula = "compound-build"
 version = 1
-contract = "graph.v2"
 extends = ["build-base"]
 target_required = true
+
+[requires]
+formula_compiler = ">=2.0.0"
 
 [catalog]
 name = "compound-build"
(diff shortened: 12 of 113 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277251913

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277251982

julianknutsen:
Please take a pass at these merge conflicts and we'll get it back in the pipeline.

PR #332, opened 2026-08-17
fix(gastown): drop refinery's masked sleep_after_idle, and guard every pack against it

### LINKED ISSUE #165: gastown pack: refinery sets both idle_timeout and sleep_after_idle (advisory in every consuming city)
**Gas City version:** gc 1.3.2 (homebrew, darwin/arm64); linked beads library v1.0.4 (via `go version -m`); packs pinned at gascity-packs `0c1fc594`. Where noted, findings re-verified against the unmodified 1.3.3 release binary.
**Environment:** macOS (darwin arm64), multi-rig city (5 rigs + HQ), Dolt-backed beads store, launchd-managed supervisor, custom tmux socket.

- **Where:** `gastown/agents/refinery/agent.toml` (pinned cache
  `3d7862f3...`, also present at proposed re-pin target `0c1fc594` — verify
  at submission time): sets `idle_timeout = "2h"` AND
  `sleep_after_idle = "300s"`.
- **Effect:** every rig that imports gastown draws
  `agent "<rig>/gastown.refinery": idle_timeout and sleep_after_idle are
  both set; idle_timeout takes precedence and sleep_after_idle only applies
  when the session survives the idle_timeout check` — 5 warnings in this
  city alone, on every `gc config show`.
- **Ask:** ship a single idle key on the refinery template. If the
  precedence pair is intentional (sleep as a pre-timeout nap), say so in a
  comment and silence the advisory for that shape; otherwise drop
  `sleep_after_idle`.
- **Local workaround applied (P3.2, 2026-07-03):** city.toml
  `[[patches.agent]]` per refinery with `sleep_after_idle = ""` — empty
  string unsets the pack key on gc 1.3.2 (omission merges the pack value
  back; `"0s"` is a literal zero, still warns). Remove the five patches
  when the pack fix lands.
---
*Filed from a downstream Gas City deployment as part of an operator-approved findings bundle (18 items; cross-references like “ask #N” refer to sibling issues in this batch — index in the first comment once all are filed). Happy to provide deeper traces or re-run repros on request.*

sjarmak:
Closes #165.

## The fix

`gastown/agents/refinery/agent.toml` set both `idle_timeout` and `sleep_after_idle`. Gas City reports that pairing as an advisory in every city importing the pack:

```
agent "<rig>/gastown.refinery": idle_timeout and sleep_after_idle are both set;
idle_timeout takes precedence and sleep_after_idle only applies when the session
survives the idle_timeout check
```

Refinery is the only gastown agent that sets both. Mayor, witness and deacon use `1h`; polecat, dog and refinery use `2h`; none of the others sets `sleep_after_idle`. The line arrived in a bulk embedded-pack sync (`2068e84`, #73) with no stated rationale. Dropping it is the change that makes refinery match the rest of the pack, and it removes the advisory.

## The test

`tests/test_pack_composition_warnings.py` discovers every pack that ships agent definitions, stands each one up in an isolated scratch city, and runs the real `gc config show` against it.

That structure is the point. A pack's own tests read its TOML and can tell you the file parses. They cannot tell you what Gas City says when it loads that file into a city, which is the only place a user meets this defect. So the test exercises the shipped pack directory through the shipped binary, not a copy of either.

**It goes red on the real defect.** Restoring the `sleep_after_idle` line, with nothing else changed:

```
1 failed, 7 passed        gastown
```

Removing it again:

```
8 passed
```

Verified against two `gc` builds, since the linter version skew in #330 showed a same-answer-everywhere assumption is not safe here: the released `v1.4.1` and a build from current `gascity` main give the same result in both directions.

**And it cannot go green by accident.** Each case also imports a fixture pack, written into the test's tmp dir, whose single agent sets both keys on purpose. The test asserts that canary is reported in the same invocation, before it checks the pack under test. Without that, a pack with no advisories and a binary that stopped emitting advisories are indistinguishable, and the guard would pass forever after any upstream change to the check.

That guard earned itself during development. My first offender filter matched `"gastown.` with the quote adjacent, which never fires for a rig-scoped agent because the advisory names it `"demo/gastown.refinery"`. The canary was green throughout; the A/B is what caught it. The filter now parses the quoted name and compares the binding.

## CI

Added to the existing `GC_TEST_BIN` step beside `test_gc_role_prompt_integration.py`. This touches `.github/workflows/ci.yml`, one line.

Without that, the file still gets collected by the "Run Python pack tests" step, where `GC_TEST_BIN` is unset and all cases skip. A guard that only runs on a contributor's machine is the failure this is meant to close.

## Verification

```
$ GC_TEST_BIN=<gc> python3 -m pytest tests/test_pack_composition_warnings.py -q
8 passed

$ python3 -m pytest tests contributing/tests gascity/tests discord/tests github/tests \
    slack-full/tests slack-channel/tests pr-pipeline/tests profiler/tests -q
14 failed, 1266 passed, 15 skipped
```

All 14 failures reproduce on `upstream/main` with this branch's changes stashed: 12 in `discord/tests` and `github/tests`, and 2 in `gascity/tests/test_formula_assets.py::...city_claim_command...`. None is touched by this change.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -61,7 +61,7 @@ jobs:
                       pr-pipeline slack-channel slack-full slack-mini; do
             "$GC_BIN" lint "$pack"
           done
-          GC_TEST_BIN="$GC_BIN" python3 -m pytest tests/test_gc_role_prompt_integration.py -q
+          GC_TEST_BIN="$GC_BIN" python3 -m pytest tests/test_gc_role_prompt_integration.py tests/test_pack_composition_warnings.py -q
 
       # Installed, not parsed. `gc lint` above reads each pack on its own and
       # the per-pack suites read their own TOML; neither can say what gc does
--- a/gastown/agents/refinery/agent.toml
+++ b/gastown/agents/refinery/agent.toml
@@ -4,5 +4,4 @@ work_dir = ".gc/worktrees/{{.Rig}}/refinery"
 nudge = "Run 'gc prime' to check merge queue and begin processing."
 pre_start = ["{{.ConfigDir}}/assets/scripts/worktree-setup.sh {{.RigRoot}} {{.WorkDir}} {{.AgentBase}} --sync"]
 idle_timeout = "2h"
-sleep_after_idle = "300s"
 max_active_sessions = 1
--- a/tests/test_pack_composition_warnings.py
+++ b/tests/test_pack_composition_warnings.py
@@ -43,13 +43,39 @@ def advisory_binding(line: str) -> str | None:
     return match.group(1).rsplit("/", 1)[-1].split(".", 1)[0]
 
 
+def owning_pack(agent_toml: Path) -> Path | None:
+    """The nearest ancestor of an agent.toml that is itself a pack."""
+    for parent in agent_toml.parents:
+        if parent == REPO_ROOT.parent:
+            return None
+        if (parent / "pack.toml").is_file():
+            return parent
+    return None
+
+
 def agent_bearing_packs() -> list[str]:
-    """Pack directories that ship at least one agent definition."""
-    names = {
-        path.parents[2].name
-        for path in REPO_ROOT.glob("*/agents/*/agent.toml")
+    """Pack directories that ship at least one agent definition, REPO_ROOT-relative.
+
+    Derived from `pack.toml` presence rather than from a fixed depth. A glob of
+    `*/agents/*/agent.toml` reads as "every pack" and silently drops any pack that
+    is not a direct child of the repo root: `gascity/roles` is a pack with its own
+    `pack.toml` and twelve agents, and it was never parameterized. The nested case
+    is the one a depth-pinned pattern always misses, so key on the marker file.
+    """
+    packs = {
+        pack.relative_to(REPO_ROOT).as_posix()
+        for pack in (
+            owning_pack(path)
+            for path in REPO_ROOT.rglob("agents/*/agent.toml")
+        )
+        if pack is not None
     }
-    return sorted(names)
+    return sorted(packs)
+
+
+def binding_name(pack_path: str) -> str:
+    """The import binding for a pack path; `gascity/roles` cannot be a bare key."""
+    return pack_path.replace("/", "-")
 
 
 @dataclass(frozen=True)
@@ -103,7 +129,8 @@ def write_canary_pack(root: Path) -> Path:
     return pack_dir
 
 
-def write_city(root: Path, pack_name: str) -> Workspace:
+def write_city(root: Path, pack_path: str) -> Workspace:
+    binding = binding_name(pack_path)
     city_dir = root / "city"
     rig_dir = root / "fixture"
     gc_home = root / "gc-home"
@@ -114,7 +141,7 @@ def write_city(root: Path, pack_name: str) -> Workspace:
     home.mkdir()
 
     canary_pack = write_canary_pack(root)
-    pack_source = (REPO_ROOT / pack_name).resolve()
+    pack_source = (REPO_ROOT / pack_path).resolve()
 
     city_dir.joinpath("pack.toml").write_text(
         textwrap.dedent(
@@ -123,7 +150,7 @@ def write_city(root: Path, pack_name: str) -> Workspace:
             name = "composition-warnings"
             schema = 2
 
-            [imports.{pack_name}]
+            [imports.{binding}]
             source = {json.dumps(str(pack_source))}
 
             [imports.{CANARY_BINDING}]
@@ -173,6 +200,15 @@ def write_city(root: Path, pack_name: str) -> Workspace:
 
 
 def config_show(gc_test_bin: Path, workspace: Workspace) -> str:
+    """Run `gc config show`, refusing to return output from a failed load.
+
+    Returning `stdout + stderr` alone makes "the config loaded and named no
+    offender" indistinguishable from "the config did not load at all": a binary
+    that aborts before it reaches advisory emission prints no advisory line, and
+    every offender assertion below is a search for an absent string. The canary
+    catches most of that, but only for failures that still emit some advisories.
+    A nonzero exit is the one signal that separates the two, so read it.
+    """
     result = subprocess.run(
         [str(gc_test_bin), "config", "show"],
         cwd=workspace.rig_dir,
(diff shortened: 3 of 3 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277209112

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277209185

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277251741

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277251810

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277287919

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277287988

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277288057

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277308488

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @sjarmak! This PR removes `sleep_after_idle` from
`gastown/agents/refinery/agent.toml` -- the one gastown agent that set it alongside
`idle_timeout` -- and, more valuably, adds a guard so the pairing cannot quietly come back.
The guard is the part that matters to the project: `tests/test_pack_composition_warnings.py`
stands every agent-bearing pack up in an isolated scratch city and runs the real
`gc config show` against it, so the advisory is caught where a user actually meets it rather
than by re-parsing TOML that a pack's own tests already parse.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- Pack discovery enumerated the working filesystem (`REPO_ROOT.rglob`), which made the suite
  unrunnable from a live checkout: any nested worktree or scratch clone under the repo root was
  collected as a pack, and each dot-prefixed binding rendered invalid TOML. Discovery now derives
  from `git ls-files -z`, and `owning_pack` additionally requires the *attributing* ancestor
  `pack.toml` to be tracked -- which closes a quieter re-attribution variant the first review round
  had not named. On the tracked tree this yields exactly the eight real packs (bmad,
  compound-engineering, gascity/roles, gastown, gstack, oversight-rig, profiler, superpowers).
- The scratch subprocess inherited the caller's whole environment, undoing a fix CI had previously
  caught. This is now closed structurally rather than by copying a filter: the shared harness owns
  the `GC_`/`BEADS_`/`XDG_` scrub, and it was verified both by a direct probe (poisoned vars
  stripped) and by running the full suite from a session that carried those vars unscrubbed.
- About 150 lines of the shared live-city harness had been re-implemented one directory away. The
  suite now imports `REPO_ROOT`, `gc_output`, `gc_test_bin` and `write_city` from `gc_live_city`;
  what stays local is exactly the idle-advisory canary fixture, with a docstring recording why the
  harness's own canary cannot substitute for it.
- The failure message shipped the wrong mechanism as repo-wide advice. It now states the actual
  source-supported condition -- dropping `sleep_after_idle` is behavior-preserving only for an
  on-demand named session whose value matches the 300s default, because the awake set reads
  `sleep_after_idle` and never `idle_timeout`. This matters beyond wording: it is the guidance the
  next maintainer will act on, and a negative control confirmed the corrected message renders
  end-to-end on a real failure.
- The canary could not distinguish a binding-named advisory from a pack-named one, so a future gc
  that changed attribution would have silently muted the offender match instead of failing. The
  canary's declared name and binding name now differ, which makes that change loud.
- A silently-empty parameter set would have read as "nothing to check" forever; an explicit
  empty-set guard now turns pytest's silent empty-parametrize skip into a loud collection failure.
  That one was unrequested hardening on your part and is a genuinely good catch.

The production change itself was independently re-verified rather than taken at face value: refinery
is an `on_demand` named session and gc's on-demand default idle timeout is five minutes, so the
removed explicit 300s and the default are the same value -- the drop is behavior-preserving. Two
reviewers re-read `compute_awake_set.go` and `validate_semantics.go` this round and both executed the
suite. Worth noting, an earlier review round had asserted a "masked, therefore inert" reading of the
interaction; that was wrong, and correcting it is what produced the fixed failure message above.

Review findings addressed:
- All six items from the first review round are closed by commit `34ae234`: tracked-tree pack
  discovery, reuse of the shared live-city harness, the inherited-environment leak, the corrected
  failure-message mechanism, the canary attribution split, and a stale agent count in the docstring.
  That commit touches only `tests/test_pack_composition_warnings.py` (109 insertions, 136 deletions);
  the production change and `tests/gc_live_city.py` are untouched by it.

Non-blocking follow-ups (shared for visibility -- NOT completed in this PR):
- `mc-ptav4` -- Test Evidence Quality, nit: pack discovery shells out to `git ls-files` with
  `check=True` at *collection* time, so a non-git copy of the tree (exported tarball, `cp -R` scratch
  lab) errors while collecting the file, where the previous version collected anywhere. Every
  supported runner has `.git`, and a loud self-naming failure is arguably the intended shape, so this
  is recorded as a decision rather than a defect. Left as-is deliberately: the `git ls-files ...
  check=True` block is the house idiom, byte-identical in three files.
- `mc-pe8c0` -- Architectural Consistency, nit: `.github/workflows/ci.yml` still names that step
  "Lint and exercise shared role prompt composition", but it now also runs the pack-composition
  suite, so a CI failure notification points at the wrong subsystem. Left as-is deliberately: a
  rename would strand a "the shared-role-prompt step" pointer in
  `tests/test_gastown_lint_findings.py`, and that is not worth re-cutting an approved head for.
- `mc-iqa7y` -- standing, pre-existing, and explicitly out of scope for this PR: CI installs
  `cmd/gc@latest` unpinned, so an upstream rewording of the advisory fragment could redden the canary
  on unrelated PRs. Your PR adds one more consumer of that binary but does not cause the condition;
  it is filed so the risk is tracked rather than re-discovered.

### Maintainer Changes
Maintainer fixes were pushed directly to this PR branch. One commit,
`34ae234 test(packs): share the live-city harness and derive packs from the tracked tree` --
`tests/test_pack_composition_warnings.py`, 1 file changed, 109 insertions(+), 136 deletions(-).
Your two commits are preserved with their original authorship, and the branch was rebased onto
current `main` with the patch content unchanged.

### Final Review Status
Ready for the merge queue: final head `34ae234` has passing required GitHub checks, and the
merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can
pick it up.

The focused suite was re-run on the rebased head before the push (8 passed), since moving onto
current `main` produces a different tree than the one review saw.

I also refreshed this PR's description, which is the one thing that had gone stale: the run counts
predated the roster's growth to eight packs ("7 passed" and "1 failed, 6 passed" are now "8 passed"
and "1 failed, 7 passed"), and the development anecdote quoted `"fixture/gastown.refinery"`, the rig
name of the local harness that this round replaced -- the shared harness writes its rig as `demo`.
The quoted gc advisory text is gc's own wording and was left exactly as you had it.

CI: https://github.com/gastownhall/gascity-packs/actions/runs/34198822774

### Review Iterations
2 review passes. The first requested changes across pack discovery, harness reuse, environment
isolation, the failure-message mechanism and canary attribution; the second confirmed every one of
them closed with executed evidence from independent reviewers, and approved with no gating items
remaining.

---

PR #337, opened 2026-08-20
fix(slack-full): accidental-mrkdwn guard — approximately-tildes no longer strike through

aphexcx:
## Problem

Slack mrkdwn pairs tilde characters into strikethrough, and its pairing is looser than any documented rule. An agent posting a finance summary containing `~$58.5k … ~$16.5k` (tilde as "approximately") had half the message rendered struck through. Slack offers **no escape sequence** for formatting characters — a backslash renders literally, and `mrkdwn: false` would also kill intentional `*bold*` and bullets — so callers cannot protect themselves.

## Fix

A new shared module, `scripts/slack_mrkdwn.py`, guards outbound text at the pack-command send path. All five text-sending commands — `reply-current` (both the legacy extmsg path and the company-token path), `publish`, `publish-to-channel`, `upload --initial-comment`, and `delegate` — pass the body through `escape_accidental_mrkdwn()`, which substitutes U+223C TILDE OPERATOR (visually near-identical, never a delimiter) for any tilde that could pair accidentally.

The guard is deliberately conservative about which bytes it touches:

- **Code spans** (``` fences and `inline`) never have their bytes modified — literal text there must survive copy-paste. They are *masked* rather than cut out before the tilde census runs, because Slack pairs tildes **across** an inline span: ``~$5k (see `calc.py`) vs ~$7k`` is one pairable line and both tildes are substituted.
- **Lines with fewer than two tildes are never modified** — strikethrough needs a same-line pair, so a lone `cd ~/repo` keeps its ASCII byte. "Lone" counts per rendered line, so two home-relative paths on one line (`rsync ~/a ~/b`) can pair and both are substituted — put twin paths in a code span or use `--raw`.
- **`~` before an optionally-signed digit/currency** (`~$5k`, `~-$13.5k`, `~9/2`, `~.5`, `~₿0.5`) reads as "approximately" and is never treated as a delimiter.
- **Deliberate tight-wrapped strikethrough** (`~word~`, `~a few words~`) still renders, including when wrapped in mrkdwn delimiters (`*~cancelled~*`, `_~cancelled~_`); `*bold*`, `_italics_`, and bullets are untouched.
- **Idempotent**, length-preserving, and a new `--raw` flag on every affected command sends the body byte-for-byte verbatim.

### Design notes

- Guarding in the Go adapter instead was considered and rejected: a raw-bypass would need a new field plumbed through gc's extmsg outbound API, and the command layer is the shared choke point for every caller shape — including the company-token path, which posts straight to `chat.postMessage` without touching the adapter.
- Underscore/asterisk pairs are deliberately not guarded: accidental italics are legible (unlike strikethrough), snake_case underscores are mid-word and cannot delimit, and homoglyph-substituting identifiers would corrupt copy-pasted commands — a worse papercut than the one being fixed.
- `post-message` is **out of scope**: it renders agent text through the Block Kit path in the Go adapter, so guarding it needs a second implementation of the heuristic plus a `gc-slack-cli` rebuild. The exclusion is documented in `commands/post-message/help.md`, the CHANGELOG, and the agent-facing template fragment, and tracked as `gp-x5bdy`.
- An **unmatched** ` ``` ` is treated as a code block through end-of-message, so it suspends the guard for the remainder. That is the module's one under-substituting branch; the premise it rests on, and the reason it is kept, are recorded in the module docstring and pinned by a test.

## Tests

`python -m pytest slack-full/tests` on this branch: **431 passed.** New coverage (31 new test functions):

- `tests/test_slack_mrkdwn.py` — 17 unit tests on the guard, including the verbatim repro line, the cross-code-span census, deliberate-strikethrough preservation, mrkdwn-delimiter boundaries, CRLF, signed/currency forms, idempotency, closed and unclosed fences, and the mid-line stray-fence shape.
- CLI plumbing tests in the `reply-current` (legacy + company paths, `--raw`), `publish`, `publish-to-channel`, `upload`, and `delegate` suites asserting guarded text reaches the transport and `--raw` bypasses.

Pure script-side change — no adapter rebuild or service restart needed.

---

_Maintainer note: this description was refreshed during adoption to match the final reviewed patch. The original submission's design and prose are the contributor's; the bullets above were updated where the review/fix loop changed behavior (cross-span tilde census, `*`/`_` boundaries, per-line twin-path rule, `post-message` exclusion, unmatched-fence premise) and where the test counts moved._

### DIFF
--- a/slack-full/CHANGELOG.md
+++ b/slack-full/CHANGELOG.md
@@ -26,6 +26,26 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   commands sends the body verbatim. Pure script-side change — no
   adapter restart needed.
 
+  Two limits worth knowing. "Lone tilde" is per rendered line, so two
+  home-relative paths on one line (`rsync ~/a ~/b`) are a pairable pair
+  and both get substituted; wrap twin paths in a code span, or use
+  `--raw`, when the exact bytes matter. And `gc slack post-message` is
+  deliberately out of scope: its milestone summary, field values, and
+  rollup items are rendered as mrkdwn by the Go Block Kit renderer,
+  downstream of this Python guard, so a payload carrying
+  "~$58.5k … ~$16.5k" still strikes through (titles are `plain_text`
+  and unaffected). That surface predates this change and guarding it
+  would need a second implementation of the heuristic in Go plus a
+  `gc-slack-cli` rebuild; tracked as gp-x5bdy.
+
+  Scope is this pack only. The sibling packs `slack-channel` (same
+  `publish` / `publish-to-channel` / `reply-current` verbs) and
+  `slack-mini` (`post-message`) carry no guard, so the same `gc slack`
+  verb behaves differently depending on which pack a city installed.
+  Porting was left out deliberately to keep this change script-side
+  with no adapter rebuild; guard-port vs. documented exclusion is
+  tracked as gp-cbxzk.
+
 ### Fixed
 
 - Inbound files recorded inside Slack itself — voice clips (file subtype
--- a/slack-full/CHANGELOG.md
+++ b/slack-full/CHANGELOG.md
@@ -8,6 +8,24 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Accidental-mrkdwn guard on the send path (gp-o42): `gc slack
+  reply-current` (legacy and company paths), `publish`,
+  `publish-to-channel`, `upload --initial-comment`, and `delegate` now
+  neutralize tildes that would pair into unintended Slack strikethrough
+  ("~$58.5k … ~$16.5k" rendered half a runway summary struck through).
+  Slack mrkdwn has no escape sequence, so accidental delimiter tildes
+  are substituted with the visually identical U+223C TILDE OPERATOR —
+  but only on lines where a pair could actually form: lone tildes keep
+  their ASCII byte (`cd ~/repo` survives copy-paste), code spans are
+  never touched, a tilde before an optionally-signed digit/currency
+  (`~$5k`, `~-$13.5k`, `~9/2`) is never a delimiter, and deliberate
+  tight-wrapped `~word~` strikethrough still renders. `*bold*`,
+  `_italics_`, and bullets are unaffected. New `--raw` flag on all five
+  commands sends the body verbatim. Pure script-side change — no
+  adapter restart needed.
+
 ### Fixed
 
 - Inbound files recorded inside Slack itself — voice clips (file subtype
--- a/slack-full/commands/delegate/help.md
+++ b/slack-full/commands/delegate/help.md
@@ -49,3 +49,10 @@ Examples
   gc slack delegate --turn-ref gct-0123456789abcdef0123 --cancel --to riley
 
 Routes to: scripts/slack_company_outbound.py delegate
+
+Formatting guard: tildes that would accidentally pair into Slack
+strikethrough (e.g. "~$58.5k … ~$16.5k" — tilde as "approximately") are
+neutralized by default with a visually identical substitute (U+223C).
+Deliberate tight-wrapped `~word~` strikethrough, code spans, lone
+tildes, and every other formatting character pass through untouched.
+Pass --raw to send the body byte-for-byte verbatim.
--- a/slack-full/commands/delegate/help.md
+++ b/slack-full/commands/delegate/help.md
@@ -48,11 +48,13 @@ Examples
   gc slack delegate --turn-ref gct-0123456789abcdef0123 --to riley --body-file /tmp/ask.txt
   gc slack delegate --turn-ref gct-0123456789abcdef0123 --cancel --to riley
 
-Routes to: scripts/slack_company_outbound.py delegate
-
 Formatting guard: tildes that would accidentally pair into Slack
 strikethrough (e.g. "~$58.5k … ~$16.5k" — tilde as "approximately") are
 neutralized by default with a visually identical substitute (U+223C).
 Deliberate tight-wrapped `~word~` strikethrough, code spans, lone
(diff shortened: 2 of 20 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277251379

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277251448

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277287771

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277287840

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277308020

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @aphexcx! This PR closes a genuinely nasty papercut: Slack pairs bare `~` characters into strikethrough with looser rules than anything documented, and it offers no escape sequence — so an agent writing `~$58.5k … ~$16.5k` had half a finance summary silently struck through and no way to defend itself. Your `scripts/slack_mrkdwn.py` guard fixes that at the one place every send path already funnels through, keeps the fix script-side (no adapter rebuild, no service restart), and pairs it with a `--raw` escape hatch so byte-exact sends are still possible. That combination — real bug, narrow blast radius, reversible — is exactly the shape of change that is easy to say yes to.

### Original PR Review
Decision: approve (resolved from an initial `request_changes`).

Specific gaps fixed:
- **The tilde census was computed per code-span segment, not per rendered line.** Slack pairs tildes *across* an inline span, so ``budget ~$5k (see `calc.py`) vs ~$7k`` slipped through unguarded — the original bug, reproduced through a shape the first version did not cover. The final patch masks code-span bytes in place instead of splitting on them, so the census sees the whole rendered line while the span's bytes still survive verbatim. Re-executed at the final head across a 17-probe matrix, including multi-line fences and two-span lines, with the design's invariants checked directly: `len(guard(x)) == len(x)`, `guard(guard(x)) == guard(x)`, and no mask byte reaching the output even for input that already contains one.
- **`upload --initial-comment` and `delegate` were guarded but untested.** Both call sites were deletable with the suite still green. The final patch adds guarded-default and `--raw`-bypass tests for each; mutation-checking them in a scratch copy (replacing both guard calls with identity) reddens exactly the four new tests and nothing else.
- **`post-message` renders agent text as mrkdwn and is not guarded.** It goes through the Block Kit path in the Go adapter, so guarding it means a second implementation of the heuristic plus a `gc-slack-cli` rebuild — worth keeping out of a script-side PR. Rather than leave it as a silent hole, the exclusion is now written where an agent would actually look (`commands/post-message/help.md`, the CHANGELOG entry, and the `slack-v0` template fragment) and tracked as `gp-x5bdy`.

Review findings addressed (all in maintainer commit `1fd1f38`):
- **Contract fidelity — an unmatched ` ``` ` suspends the guard for the rest of the message.** The trailing fence alternative is unanchored under `DOTALL`, so a stray triple-backtick mid-sentence passes the remainder through with ASCII tildes intact. Every other branch in the module fails toward substituting (worst case: a cosmetic U+223C); this one fails toward the original bug. It is kept — corrupting bytes Slack renders as code is the more concrete harm, and a stray unmatched fence in prose is rare — but it is no longer implicit: the premise, its evidence, and the narrowing that follows if it is falsified are recorded in the module docstring, and the shape is pinned by `test_stray_midline_fence_suspends_the_guard` with a no-fence control.
- **Boundary sets and idiom nits.** `*`/`_` were added to the opener/closer sets so `*~cancelled~*` and `_~cancelled~_` keep their deliberate strikethrough, `\r` was added to the closer-followers for CRLF bodies, `₿` joined the currency class, and the module docstring's opener enumeration — which had drifted narrower than the code — now matches it.
- **`delegate/help.md` trailer convention.** The new formatting-guard paragraph had landed below the `Routes to:` line, making it the only one of 15 help files not ending on that trailer; it now sits above it, matching `post-message/help.md`.

- `gp-x5bdy`: **`post-message`'s Block Kit surface still strikes through on the original payload shape** (minor, carry-over from iteration 1, mitigated). Documented in three places and tracked, but documented is not fixed — the decision to port the guard into the Go adapter or keep the exclusion permanently is deferred to that bead.
- `gp-cbxzk`: **the sibling packs `slack-channel` and `slack-mini` carry the same defect with no guard** (minor, scope). `slack-channel` exposes the same `publish` / `publish-to-channel` / `reply-current` verbs and `slack-mini` the same `post-message`, so an identically-named `gc slack` verb now behaves differently depending on which pack a city installed. Deliberately *not* fixed here — widening this PR would cost it the script-side/no-rebuild property that makes it safe — but the pack-family scope is now explicit in the CHANGELOG and the guard-port-vs-documented-exclusion decision is tracked.

### Maintainer Changes
`1fd1f38 fix(slack-full): close the review findings on the accidental-mrkdwn guard` — 16 files changed, +367/−29 (guard module, five help files, the template fragment, CHANGELOG, and three test modules). Your commit was replayed unchanged onto current `main` — the branch also needed a `CHANGELOG.md` `[Unreleased]` add/add resolution against `#363`/`#364`, resolved as a union with your Added block above the upstream Fixed block, verified as a zero-deletion line union with every other file replaying at an identical patch-id.

### Final Review Status
Ready for the merge queue: final head `1fd1f38` has passing required GitHub checks (Check, Analyze (go), Analyze (python), CodeQL), the full `slack-full` suite is green at that head (431 passed) after the rebase onto current `main`, and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/gascity-packs/actions/runs/34196349040

### Review Iterations
2 review passes. Iteration 1 (three lanes: design-implementation, implementation, design-test-risk) returned `request_changes` on a ten-item checklist; iteration 2 re-reviewed the fix commit and returned `approve` with no gating items, having independently re-executed each fix rather than accepting the reviewers' assertions.

---

PR #339, opened 2026-08-20
gastown: propulsion covers the INBOX as a peer work queue (not-done-until-archived)

csauer02-personal-user:
Refs #342.

## What

Extends the propulsion fragment (`gastown/template-fragments/propulsion.template.md`) so the
inbox is a **peer work queue** to the hook, with an explicit archive discipline:

> CLEAR YOUR HOOK. READ YOUR INBOX. ACT ON YOUR INBOX. ACT ON YOUR HOOK.
> IT IS NOT DONE UNTIL IT IS ARCHIVED.

- Inbox items sort into three kinds — **obligation** (archive when resolved), **standing
  reference** (never archive; you will need to read it again), **stale/noise** (archive once
  verified) — with the state rule overriding the kind: **unresolved never archives**.
- **Mixed items** (a standing order that also asks for an action) owe both halves: discharge
  the action, keep the reference readable.
- **Archived is the receipt, never the remedy** — archiving something unresolved converts a
  visible obligation into an invisible one nothing downstream will surface again.
- Adds the third surface neither queue can show: **the state your role owns** (a held lease,
  a published ref, an integration root) — checked at the end of each unit of work.
- Mayor startup: inbox processing becomes a peer step rather than the hook-empty fallback.

## Why

A fleet running this fragment repeatedly stalled on unread coordination mail while hooks ran
clean — dispatch, escalations, and review verdicts sat unprocessed because the doctrine named
only one of the two queues. Separately, seats "cleared" inboxes by archiving unresolved
obligations, which converts visible work into invisible work. Both failure modes were observed
live; this text is what stopped them.

## LOC breakdown

| category | insertions | deletions |
|---|---|---|
| prod code | 0 | 0 |
| tests | 0 | 0 |
| docs / prompt templates | 88 | 10 |
| generated | 0 | 0 |

One file. The pre-image blob is byte-identical to current `main` (`b5faf6b0`), so this applies
cleanly with no divergence carried.

---

**Revision `8a17718`** (addressing intake review): the three-kind taxonomy, state rule, and mixed-items paragraphs are removed in favor of #342's framing — one archive rule, stated once ("archive only when the obligation is resolved, or represented by durable tracked work"). The retention paragraph is corrected against gascity source (`mail.retention_ttl`: empty/0 disables the purge; wisp-tier only) and now forbids reasoning from either retention assumption. A contract test (`tests/test_propulsion_inbox_contract.py`, 181 lines, #335-style with mutation proof 4/4) pins the load-bearing sentences. Current LOC: prompt templates +66/−10, tests +181, no prod code.

### DIFF
--- a/gastown/template-fragments/propulsion.template.md
+++ b/gastown/template-fragments/propulsion.template.md
@@ -7,66 +7,48 @@ The entire system's throughput depends on ONE thing: when an agent finds work
 on their hook, they EXECUTE. No confirmation. No questions. No waiting.
 
 **And the hook is not your only queue. Work that arrives in your INBOX obeys
-that same rule: you have TWO work queues and they are PEERS.**
+that same rule: you have TWO work queues and they are PEERS.** Inbox processing
+is never conditional on the hook being empty.
 
-CLEAR YOUR HOOK. READ YOUR INBOX. ACT ON YOUR INBOX. ACT ON YOUR HOOK.
-IT IS NOT DONE UNTIL IT IS ARCHIVED.
+CLEAR YOUR HOOK. READ YOUR INBOX. ACT ON BOTH.
+IT IS NOT DONE UNTIL IT IS ARCHIVED, AND IT IS NOT ARCHIVABLE UNTIL IT IS
+RESOLVED.
 
 Mail is not correspondence you get to when the hook is empty. It carries
 dispatch, escalations, review verdicts and human instruction, and an unworked
 inbox stalls the engine exactly the way an unworked hook does. Neither queue is
 the fallback for the other.
 
-**Not everything in an inbox is an obligation, and "not done until it is
-archived" is false for one of the three kinds. Sort each item before you touch
-it:**
-- **AN OBLIGATION** — do this, answer this, act on this. Archive it WHEN
-  RESOLVED and not before. This is the kind the line above is about.
-- **A STANDING REFERENCE** — a doctrine, a policy, an order that binds you from
-  now on. It has no end state, so there is nothing to discharge — but THE INBOX
-  IS NOT DURABLE STORAGE (read mail is lifecycle-managed and will not persist),
-  so "leave it in the inbox" loses it exactly as surely as archiving it. MAKE IT
-  DURABLE THE MOMENT YOU CLASSIFY IT: file it as a bead, or land it in your
-  role's standing instructions. Once it lives somewhere durable, the mail is a
-  delivery receipt and may be archived. A standing order that exists only in a
-  mailbox is one the town is already losing.
-- **STALE OR NOISE** — nothing to discharge and nothing to keep. Archive it,
-  once you have verified that is what it is.
+**Reading and archiving are different acts, and keeping them apart is the whole
+discipline.** Reading clears the unread count. Archiving records that the
+obligation is discharged. One rule governs the second act:
 
-**THE STATE RULE OVERRIDES THE KIND: UNRESOLVED NEVER ARCHIVES.** Kind tells you
-what an item IS; state tells you whether you are allowed to file it away. If you
-have not actually handled it — including an item you called stale but never
-verified — it does not get archived, whatever the lines above say. Resolve it
-first, or make the obligation durable somewhere it will be seen.
+> ARCHIVE ONLY WHEN THE OBLIGATION IS RESOLVED, OR WHEN IT IS REPRESENTED BY
+> DURABLE TRACKED WORK — a bead, or a line in your role's standing instructions.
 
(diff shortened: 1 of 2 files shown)


csauer02-personal-user:
Thanks for the detailed intake — all three items are addressed in `8a17718`, and the change now sits under #342's framing rather than beside it.

**On the taxonomy.** #342 says the fix "does not require a new mail taxonomy", and re-reading the branch with that in hand, the taxonomy was the *cause* of item (2): the three kinds forced "IT IS NOT DONE UNTIL IT IS ARCHIVED" to be walked back three separate times — once per kind, once for the state rule, once for mixed items — which is exactly the "absolute slogans beside later qualifications" you flagged. So the three kinds, the state-rule paragraph, and the mixed-items paragraph are deleted rather than reworded, and the invariant is stated directly, once, in `propulsion-base`:

> ARCHIVE ONLY WHEN THE OBLIGATION IS RESOLVED, OR WHEN IT IS REPRESENTED BY DURABLE TRACKED WORK — a bead, or a line in your role's standing instructions.

The completion slogan keeps both directions on one line so it can't be read as license to archive early ("IT IS NOT DONE UNTIL IT IS ARCHIVED, AND IT IS NOT ARCHIVABLE UNTIL IT IS RESOLVED"), and reading is separated from archiving explicitly: reading clears the unread count, archiving records discharge. The Mayor and crew startup steps use the acceptance criterion's own words — "resolved or represented by durable tracked work" — so the two cannot drift apart. Prompt expansion is down from 97 lines to 66.

**On retention (item 1).** You're right and the old text was wrong. Checked against gascity: `MailConfig.RetentionTTLDuration` treats empty or `"0"` as disabled; the purge arm in `wisp_gc.go` runs only when the TTL is positive; and `PurgeReadMessageWisps` lists with `TierMode: beads.TierWisps`, so main-tier messages are never candidates. The fragment now states that behavior and, deliberately, forbids reasoning from *either* direction — a seat should not argue "the mail will be swept" or "the mail will still be there", because both depend on the city's config. The durability point now rests on "the mailbox is a delivery channel; the bead graph is the record", which holds under every setting.

**On the guard (item 3).** `tests/test_propulsion_inbox_contract.py`, in the #335 lane — root `tests/`, so CI's existing "Run Python pack tests" step picks it up with no workflow change, and it is not `GC_TEST_BIN`-gated so it runs on every PR. It pins the load-bearing sentences, prohibits the known-bad shapes (hook-empty gating, the three retention claims above, "Read + archive is fine"), walks each role's numbered startup steps so a *reworded* re-gating is caught structurally and not just by string, and requires every role fragment to compose `propulsion-base`. Following #335's convention of showing the test can fail:

| mutation | result |
|---|---|
| reinstate `If it returns no work, **process inbox to zero unread**` | 2 failed (bad-phrasing + startup-ordering) |
| replace the archive rule with "Archive when you are done reading." | 2 failed (load-bearing + stated-once) |
| restore "read mail is lifecycle-managed and will not persist" | 2 failed (both retention prohibitions) |
| drop the archive precondition from the crew startup step | 1 failed (crew precondition) |
| unmutated | 23 passed |

There is also an anti-vacuity arm in the spirit of `write_canary_pack`: if the fragment is moved or renamed, `test_fragment_source_is_where_we_think_it_is` fails rather than every other check quietly passing on an empty read.

One thing I left in beyond #342's scope, and happy to drop it if you'd rather keep this change strictly to the issue: a short paragraph on the surface neither queue can show — the state a role owns (a held lease, a published ref, an integration root behind the tip). It's one paragraph and it's the part of the original PR that #342 doesn't cover; it can move to its own change if you prefer #339 to land as exactly the #342 fix.

Local run: `tests/test_propulsion_inbox_contract.py` 23 passed; full root `tests` 139 passed / 29 skipped; `tests/test_gc_role_prompt_integration.py` and `tests/test_gastown_lint_findings.py` 8 passed each against a real `gc`. LOC: 181 test lines, 66/-10 prompt-template lines, no prod code.

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277251010

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277287556

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277287629

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277287698

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277307522

PR #344, opened 2026-08-21
fix(slack-full): stamp a reference marker on keyed publishes so delivery can be read back

sjarmak:
## What this fixes

When the adapter posts to Slack, nothing afterwards can confirm the message
landed. `publishDedupCache` is in-process, so it holds a receipt for
`publishDedupTTL` and loses everything on restart, and a sweep that wants to
verify delivery has no way to find a specific message in channel history.

This stamps a short reference marker on outbound publishes that carry an
idempotency key, so a later reader can locate that exact message in Slack
itself:

```
_ref:50e90a583c36_
```

The marker is the first 12 hex characters of the SHA-256 of the idempotency
key, appended after a blank line. Messages with no idempotency key are
unchanged.

## Why not message metadata

Slack's message metadata would be the right home for this, but it requires app
scopes the adapter does not request. Text is what we can write with the scopes
an installed slack-full app already has.

## The second commit

Appending the marker makes an outbound message longer than the caller sized it,
so `registerAdapter` now advertises `40000 - 20 = 39980` as
`MaxMessageLength`. Callers that respect the advertised capability then produce
a post that still fits after stamping.

Two consequences worth stating rather than leaving to be discovered:

- Unkeyed messages get no marker but are still advertised the reduced ceiling.
  That costs 20 characters out of 40,000 and keeps one number for callers to
  reason about, which seemed better than a conditional limit.
- A caller that ignores the advertised capability can still hand us text that
  does not leave room for the marker. Slack **truncates** an over-length
  `chat.postMessage` from the end rather than rejecting it, so a blind append
  there would deliver the message with the marker silently cut off — the one
  failure mode this change exists to prevent, since a reader would then find no
  marker on a message that was in fact delivered. The reservation exists so
  well-behaved callers never reach that band; the third commit handles the
  callers that do.

## The third commit (maintainer adoption fixup)

Applied during maintainer review, on top of the contributor commits:

- The marker is no longer stamped when the text is empty. A keyed publish with
  empty text previously took the stamping path and returned `Delivered: true`
  for a message Slack never accepted, which also poisoned the dedup cache.
  Empty text now flows the base `no_text` → `permanent` path.
- The marker is no longer stamped when it would not fit
  (`len(rewrittenText)+referenceMarkerOverhead > slackMaxMessageLength`). The
  caller's text is posted unstamped and a line is logged; caller content is
  never trimmed to make room for the marker.
- Slack's truncation warning (`ResponseMetadata`) is parsed and logged, so a
  delivered-but-truncated message is visible rather than silent.
- `referenceSuffix`'s doc comment states the reader-side reconciler has not
  landed yet, instead of describing the agreement in the present tense.

Marker absence therefore reads as *unknown*, never as *undelivered*.

## Cross-system contract

`referenceSuffix` has to agree with the reader on the other side. The doc
comment pins the agreement with a test vector
(`dr-3msk6.3-test-vector` maps to `50e90a583c36`) so a future change to the
digest or the truncation width fails loudly rather than silently producing
markers nobody can find.

## Testing

`go build ./...`, `go vet ./...`, `go test -count=1 ./...` and
`go test -race ./...` all pass in `slack-full/adapter`.

The new tests were checked against mutations rather than only run green. Each
mutation below was applied to the production code, the suite was run, and the
file was restored and re-verified by checksum:

| mutation | expected red | result |
| --- | --- | --- |
| marker separator `"\n\n"` to `"\n"` | marker test | `TestHandlePublishAppendsReferenceMarkerWhenIdempotencyKeySet` failed |
| advertised max back to `slackMaxMessageLength` | capability test | `TestRegisterAdapterAdvertisesMarkerSafeMaxMessageLength` failed, got 40000 want 39980 |
| dedup `Get` returns `false` | dedup test | `TestHandlePublishDedupesOnIdempotencyKey` failed, Slack called 3x not 1x |
| append guard forced false | marker test | `TestHandlePublishAppendsReferenceMarkerWhenIdempotencyKeySet` failed |
| append guard forced true | omission test | `TestHandlePublishOmitsReferenceMarkerWhenNoIdempotencyKey` failed |

The last two cover the guard in both directions, so "marker present when keyed"
and "marker absent when unkeyed" are each independently asserted.

The maintainer fixup's two new guards were mutation-checked the same way during
review: dropping `&& rewrittenText != ""` reddens
`TestHandlePublishOmitsReferenceMarkerForEmptyText`, and forcing the fit check
true reddens the two ceiling subtests while the at-capability case stays green.

## Scope

Three files: `slack-full/adapter/main.go`, `slack-full/adapter/main_test.go`,
and `slack-full/CHANGELOG.md`. No change to routing, threading, credentials, or
the dedup cache's own behaviour, which was already present upstream.

### DIFF
--- a/slack-full/CHANGELOG.md
+++ b/slack-full/CHANGELOG.md
@@ -100,6 +100,31 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Outbound publishes carrying an idempotency key are stamped with a
+  low-visibility reference marker (`_ref:<12 hex>_`) so a later reader can
+  find the exact message in Slack. The in-process dedup cache only lives
+  for its TTL and does not survive a restart, so it cannot provide durable
+  delivery evidence; the marker can. The readback contract — what a marker's
+  absence does and does not mean, and where a reader has to look — is stated
+  normatively in the `referenceMarker` doc comment in `adapter/main.go`, and
+  this entry deliberately does not restate it, so there is one place to edit.
+  Two properties worth knowing without opening the source: absence always
+  means UNKNOWN, never undelivered (all history predating this change is
+  unmarked); and a marker on a threaded reply is not reachable from
+  `conversations.history`, which does not return thread replies.
+
+  Note on Slack's ceiling, since the obvious assumption is wrong: Slack does
+  not reject text over 40,000 — it truncates from the end and still answers
+  `ok:true`. The marker is the tail of the text, so an oversized post would
+  deliver a mangled partial marker under a receipt that looks perfect. The
+  advertised `MaxMessageLength` therefore reserves the marker's 20 bytes, and
+  `handlePublish` enforces the same budget itself rather than trusting an
+  advertisement nothing in gc reads today. Over the budget the caller's text
+  is posted **unstamped** rather than trimmed: the message is the caller's,
+  the bookkeeping is ours, and an honest absence is already something the
+  contract requires readers to tolerate. When Slack reports a truncation
+  anyway, the publish receipt carries `metadata["truncated"]="true"`
+  (`dr-3msk6.3`).
 - `gc slack retry-peer-fanout` — operational recovery for peer-fanout.
   Walks recent `extmsg.peer_fanout_failed` events (added in this change
   too), filters by `--since` / `--conversation` / `--max`, deduplicates
--- a/slack-full/adapter/main.go
+++ b/slack-full/adapter/main.go
@@ -791,6 +791,19 @@ type publishReceipt struct {
 	MessageID    string          `json:"message_id,omitempty"`
 	Delivered    bool            `json:"delivered"`
 	FailureKind  string          `json:"failure_kind,omitempty"`
+	// Metadata carries out-of-band facts about a delivery that the four
+	// named fields above cannot express — currently only "truncated". gc
+	// does not decode this body into its public PublishReceipt directly: it
+	// unmarshals into a fixed intermediate shim, wirePublishReceipt
+	// (gascity internal/extmsg/http_adapter.go:140-157), whose six fields
+	// are copied across one by one in toPublishReceipt. `metadata` is one
+	// of those six and reaches PublishReceipt.Metadata; any key outside
+	// that set is silently discarded at gc's boundary, so widening this
+	// struct with a new top-level field would produce a value no gc
+	// consumer can ever observe. Measured against gc's real code, not
+	// inferred. The request side already uses this idiom
+	// (publishRequest.Metadata).
+	Metadata map[string]string `json:"metadata,omitempty"`
 }
 
 type externalActor struct {
@@ -857,6 +870,15 @@ type slackPostMessageResp struct {
 	TS      string `json:"ts,omitempty"`
 	Channel string `json:"channel,omitempty"`
 	Error   string `json:"error,omitempty"`
+	// Slack reports non-fatal problems here instead of failing the call. The
+	// one that matters to us is "message_truncated": Slack truncates text past
+	// slackMaxMessageLength rather than rejecting it, and returns ok:true, so
+	// the receipt looks perfect. handlePublish stamps the reference marker at
+	// the very end of the text, so this warning is the only wire signal that a
+	// delivered keyed message is no longer findable by its marker.
+	ResponseMetadata struct {
+		Warnings []string `json:"warnings,omitempty"`
+	} `json:"response_metadata,omitempty"`
 }
 
 // Slack files-upload-v2 API types.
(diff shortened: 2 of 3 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277250864

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277250941

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277287414

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277287483

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277307773

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @sjarmak! This PR gives outbound Slack publishes a
durable identity — a short `_ref:<12 hex>_` marker derived from the idempotency
key — so a reader can find a specific message in channel history after the fact.
That matters because `publishDedupCache` is in-process and TTL-bound: today, once
the adapter restarts, there is no way to answer "did this message actually land?"
Your change turns delivery from something we assert into something we can verify,
and it does it with the scopes an installed slack-full app already has rather
than asking for message-metadata scopes we don't request.

The second commit is the part I'd single out: reserving the marker's 20 bytes in
the advertised `MaxMessageLength` (`40000 - 20`) is the kind of consequence that
usually gets discovered in production, and you both handled it and wrote down the
trade-off. The `dr-3msk6.3-test-vector` pin on `referenceSuffix` is the same
instinct — it makes a cross-system agreement fail loudly instead of silently
producing markers nobody can find.

branch.

### Original PR Review

Specific gaps fixed:
- **A keyed publish with empty text reported success for a message Slack never
  accepted.** The marker was appended unconditionally on the keyed path, so the
  empty-text case took the stamping branch and returned `Delivered: true`. Worse,
  that receipt was then cached, so the false success survived retries. The guard
  is now `req.IdempotencyKey != "" && rewrittenText != ""`, and empty text flows
  the base `no_text` → `permanent` path instead. Verified by mutation: removing
  `&& rewrittenText != ""` reddens `TestHandlePublishOmitsReferenceMarkerForEmptyText`
  and nothing else, so the test isolates exactly this guard.
- **The platform premise behind the reservation was inverted.** The PR body and a
  `main.go` comment said Slack would *reject* an over-length message. Slack
  truncates it from the end — which means a blind append would deliver the message
  with the marker cut off, the one failure mode this change exists to prevent.
  The marker is now skipped when
  `len(rewrittenText)+referenceMarkerOverhead > slackMaxMessageLength`, posting
  your text unstamped rather than trimming it, and the prose was corrected. Also
  verified as a boundary and not a blanket: forcing the fit check true reddens the
  "one byte over the advertised capability" and "at Slack's raw ceiling" subtests
  while "at the advertised capability" stays green.
- **`referenceSuffix` described agreement with a reconciler that exists at no ref.**
  The doc comment now states the reader side has not landed and names the landing
  obligation, so the contract reads as forward-looking rather than as a description
  of something already in place.

Worth saying explicitly: the review considered a stricter remedy for the ceiling
case — refuse to publish above the limit — and deliberately rejected it, because
it would convert messages that deliver today into hard failures. Skip-and-log was
the right call and it's what shipped. Caller content is never trimmed to make room
for the marker, and marker absence now reads as *unknown*, never as *undelivered*.

Review findings addressed:
- All three required changes above were fixed in `ff12a68`, along with four
  non-gating items raised on the first pass: Slack's truncation warning
  (`ResponseMetadata`) is now parsed and logged, the readback contract is
  documented, and the publish log prints `posted=`.

- `mc-quk21` — *Test Evidence Quality, minor.* Both new guards measure the
  post-alias-rewrite text, but all nine `handlePublish(` call sites in the suite
  pass `nil` for `userAliases`, so `rewrittenText == req.Text` throughout and that
  axis is unpinned. A variant that measures `len(req.Text)` instead stays green
  while posting 40,009 bytes when an `@handle` rewrite expands the text — measured
  end to end. The code is correct today; the property that makes it correct isn't
  yet protected from a refactor. A ready-to-paste subtest is recorded on the bead.
- `mc-ajoy8` — *Contract & Interface Fidelity, minor.* A delivered-but-truncated
  message is `ok:true` with a clean receipt, so the truncation signal ends in a log
  line and the reconciler this change serves can't observe marker loss
  programmatically. Note for whoever picks this up: the obvious fix (a top-level
  `Truncated bool` on the receipt) is inert — gc decodes `/publish` into a fixed
  shim and drops unknown keys. The `metadata` map is the route that survives.
- `mc-2luti` — *Debuggability, nit.* The `text=` label carries `len(rewrittenText)`
  in the skip log and `len(req.Text)` in the publish log, so a `grep` on `text=`
  mixes semantics. Renaming one to `rewritten=` fixes it.
- `mc-tyi15` — *Architectural Consistency, minor.* `slack-channel` and `slack-mini`
  still advertise a raw 40000, which is correct while they stamp nothing — but
  there's no tripwire at those registration sites for whoever ports the marker.
  Best landed with the port itself, not here, where the cross-reference would be
  speculative.
- `mc-t1xsv` — *Blast Radius, nit.* Unkeyed publishes also pay the 20-byte
  reservation. Harmless today since nothing in gc reads `MaxMessageLength`; worth a
  sentence in the registration comment if a gc-side sizer ever appears.
- `mc-2csyt` — *Recorded residual.* Whitespace-only keyed text (`" "`, `"\n"`) is
  still stamped, and it's reachable — `slack_chat_publish.py` doesn't strip
  `--body-file` content. Whether that reproduces the empty-text problem depends on
  how Slack treats whitespace-only text, which we couldn't verify; the bead records
  the `strings.TrimSpace` tightening for once that's settled.

None of these blocks the merge.

### Maintainer Changes
`ff12a68` — `fix(slack-full): don't stamp a reference marker where it lies or
cannot survive` (3 files changed, 527 insertions(+), 10 deletions(-)): the
empty-text and fit-check guards, truncation-warning parsing, the corrected
`referenceSuffix` doc, a CHANGELOG entry, and the tests covering each. Your two
commits were preserved with authorship intact and rebased onto current `main`; the
adopted patch was checked by patch-id across the rebase so nothing shifted. The PR
body was also corrected where it stated Slack rejects over-length messages.

### Final Review Status
Ready for the merge queue: final head `ff12a68` has passing required GitHub checks
(Check, Analyze (go), Analyze (python), CodeQL), and the merge-ready metadata is in
place. The adapter suite is green under `-race`, and `gofmt`/`go vet` are clean. I
am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/gascity-packs/actions/runs/34191599014

### Review Iterations
2 review passes: the first closed `request_changes` on the three gaps above, the
second re-derived each fix independently — including the mutation checks quoted
here — and closed `approve`.

---

PR #347, opened 2026-08-22
fix(gascity): create and persist the shared drain worktree for same-session items

julianknutsen:
Fixes the same-session shared-drain failure reported in gastownhall/gascity#5429.

## Problem

Every same-session shared drain fans items to `do-work-item` (directly, or via `extends` in bmad / compound-engineering / gstack / superpowers). Its `implement-item` step reads the authoritative worktree from `work_dir` metadata on the source anchor — but nothing on the same-session path ever **writes** it. The only writer in the repo (`do-work/prepare-worktree.md`) is reachable only under `drain_policy=separate`; the drain step itself is a `gc.kind=drain` control bead no agent executes, and pre-drain operator steps are forbidden from creating worktrees. The shared-worktree contract shipped read-side-only, so all 13 shared-drain sites across the maintained packs fail closed at the first item (or, for packs whose readers never mentioned `work_dir`, silently worked in the launcher checkout). This is a day-one gap, not a regression — full triage in the issue thread.

## Fix

Ports the pattern proven (but unmerged) in `b44744e` for gstack into the gascity pack, at the `do-work-item` leaf:

- **New `prepare-shared-worktree` operator step** in `do-work-item`, gating `implement-item`. Its exec check (`gascity/assets/scripts/checks/prepare-shared-worktree.sh`) does a drain-keyed create-or-reuse of one deterministic shared worktree at `$LAUNCHER_ROOT/worktrees/shared-<drain-control-id>` and persists `work_dir` on each item's `gc.drain_member_id` source anchor with read-back verification. Idempotent per item: the first item creates, later items reuse — preserving same-session semantics (item N sees items 1..N-1's commits).
- **Base ref = freshly fetched remote default branch** (never launcher HEAD, never hardcoded `main`), matching the sibling `do-work` separate-lane convention — flipping `drain_policy` no longer changes what base your items build on. (`b44744e`'s launcher-HEAD base was load-bearing only inside its own unmerged evidence-gate system; verified end-to-end for all four derived packs that origin-default is correct.)
- **`needs = ["prepare-shared-worktree"]` in the four derived packs' `implement-item` overrides** — SDK `extends` step overrides are wholesale replacement (`internal/formula/parser.go` `mergeSteps`), so the inherited edge is otherwise lost and the steps would race. The inherited step itself needs no vendoring: `../assets/` prompts resolve through the layer stack with fallback to the declaring pack, and the `.gc/scripts/checks/` path resolves at runtime (the same mechanism gstack already uses for `build-artifact-valid.sh`).
- **Reader alignment**: `do-work-item/implement-item.md` now names the `work_dir` key and fails closed explicitly (previously implicit); bmad / compound-engineering / gstack item prompts gain a minimal reader clause (their prior behavior was silent launcher-checkout work). Also fixes a latent key bug: prompts referenced `gc.drain_item_index`, but the SDK stamps `gc.drain_index` (`internal/beadmeta/keys.go`).
- **Reconcile-timing convergence**: `gc.work_dir` reaches the item root only via a reconcile tick after the first step is claimed, so the check fails closed and converges within its `max_attempts = 3` budget; for pooled/ephemeral operator sessions (where the reconciler refuses the stamp without ownership evidence) the operator prompt records its verified rig root on the workflow root first, keeping convergence deployment-independent.

## Regression net (three layers)

1. **Contract pins** (`gascity/tests/test_formula_assets.py`): writer/reader fragment pins for `do-work-item` mirroring `do-work`'s, executable-bit assertion, check-path wiring, needs edge.
2. **Cross-pack class guard**: every `[steps.drain] context="shared"` site in every maintained pack must resolve (through the `extends` chain, and additionally under `[gascity, pack]` layer shadowing) to an item formula whose resolved steps include the writer gating the implement step — with negative-control self-tests (helper rejects a writer-less formula and a dropped edge). Verified against the pre-fix tree: fails on exactly the 13 sites.
3. **Execution tests** (`tests/test_prepare_shared_worktree.py`): the real script runs against on-disk git fixtures (default branch deliberately named `trunk`; launcher HEAD deliberately ahead of origin so a HEAD-based worktree is detectably wrong; origin advanced post-clone so the fetch itself is load-bearing; an `init+fetch` launcher with no `origin/HEAD` ref exercising the refresh path) with a fake `gc` shim emitting the production one-element-list JSON shape. Covers create, stamp+read-back, idempotent reuse, out-of-band-deletion self-heal (`worktree prune`), and six fail-closed cases asserting exit code and diagnostic. All hardening was mutation-probed: each guard, the stamp, the unwrap, the fetch, and the prune were individually broken in a scratch copy and each break is caught.

Review process: five-lens review council + adversarial verification per finding; 10 confirmed findings applied in the follow-up commits (notably: a `set -euo pipefail` interaction that dead-coded the `origin/HEAD` refresh on `init+fetch` checkouts, discarded git stderr in the sandboxed gate env, and a fake-shim JSON shape that left the list-unwrap unprobed).

## Notes for reviewers

- `worktrees/` inside the launcher checkout follows `do-work`'s existing convention; shared drains add one more untracked dir there.
- `git worktree add` prints `Preparing worktree (detached HEAD …)` to stderr on the create path — it lands in the first attempt log and is harmless.
- A future landing of `b44744e` merges cleanly over this: same step id means gstack's own prepare variant wholesale-overrides the inherited one.
- Suite: `1325 passed, 29 skipped, 9530 subtests`; shellcheck zero findings.


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277215932

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277250718

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277250791

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277287268

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277287337

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277307348

PR #354, opened 2026-08-25
fix(gastown): witness patrol — real liveness signals, a salvage gate, and visible wisps

### LINKED ISSUE #351: mol-witness-patrol: Case C/D salvage can push a revert of merged work; two blind-query defects in the same file
Three defects in `gastown/formulas/mol-witness-patrol.toml`, all verified live at HEAD `aab8030d`. The first can destroy already-merged work; the other two are the same "query can't see what it's asked to guarantee" pattern.

We hit #1 in production, caught it before it pushed, and have been running a local override of this file for ~24h. Happy to open a PR if useful — filing as an issue first since the fix for #1 is a judgement call about defaults.

---

## 1. Case C/D salvage can publish a REVERT of already-merged work

**Lines 286–313.** `recover-orphaned-beads` Cases C and D salvage any dirty worktree unconditionally:

```bash
git add -A
git commit -m "witness: salvage work from orphaned agent (<bead>)"
git push origin HEAD
```

justified by *"All work is useful work — never discard."* (line 306)

That rule assumes a dirty worktree holds **new** work. It can equally hold **stale** work — an older base checked out, or a partially-applied rebase/checkout left staged. Then the diff is not an addition, it's a reversal, and the recipe pushes it.

**The live instance.** A worktree whose HEAD was already merged to `origin/main`, with 3 files staged:

```
staged diff vs HEAD:  +86 / -545        <- deletion-heavy
```

Every staged blob was byte-identical to that file at an **older** base, and differed from both HEAD and `origin/main`. Left alone, the witness would have committed and pushed a 545-line revert of merged work, with a commit message describing it as a rescue.

**Why the obvious checks don't catch it.** Our witness had already assessed this exact worktree as "nothing at risk" using `git status` (clean) and `git merge-base --is-ancestor` (merged). Both read safe. Diffing the *working tree* against `origin/main` reads `-3234`. That's the whole trap: the signals a reviewer reaches for first are the ones that miss it.

Note also that `merge-base --is-ancestor` fails open under squash-merge, so ancestry can't be the discriminator.

**Fix we're running.** Two gates before any Case C/D push. Either one alone stops both instances we've seen:

*Gate 1 — PRIMARY: is the worktree's owning bead already closed?* Authoritative — a closed bead has nothing to rescue whatever the worktree looks like. Resolve worktree path → owning bead. Note `bd list` hides closed beads by default, which is the exact fail-open the gate exists to prevent:

```bash
# --all is mandatory here, not optional.
BEADS=$(bd list --all --include-infra --limit=0 --json 2>/dev/null)
OWNERS=$(echo "$BEADS" | jq -r --arg wd "$WORKTREE" \
  '.[] | select((.metadata.work_dir // "") == $wd) | .id + " " + .status')
```

Resolve by `metadata.work_dir`, and count the salvage target itself too — agent-home work_dirs are shared, so the owning bead is not always the bead being salvaged.

*Gate 2 — BACKSTOP: would publishing this delete lines that exist on main?* A rescue is additive. Diff the **working tree** against `origin/main`, not HEAD:

```bash
NUM=$(git -C "$WORKTREE" diff origin/main --numstat 2>/dev/null)
ADD=$(echo "$NUM" | awk '$1 ~ /^[0-9]+$/ {a+=$1} END {print a+0}')
DEL=$(echo "$NUM" | awk '$2 ~ /^[0-9]+$/ {d+=$2} END {print d+0}')
NET_DELETION=0; [ "$DEL" -gt "$ADD" ] && NET_DELETION=1
```

`git diff origin/main` with no second commit compares main to the working tree, so it catches HEAD's commits and uncommitted edits in one number. Diffing against HEAD instead misses it — a worktree can be perfectly clean and still hold a full revert.

| Gate 1 SPENT | NET_DELETION | Action |
|---|---|---|
| 1 | — | Spent worktree. Capture non-destructively, push **nothing**, go to cleanup. |
| 0 | 1 | Net-deletion rescue. Capture, escalate for human eyes, push **nothing**. |
| 0 | 0 | Ordinary salvage. Case C/D as written — stays automatic. |

Whenever a gate withholds content we still preserve the bytes via `git stash create`, so "never discard" is honoured without handing the refinery something mergeable.

**Replay results:** the trap worktree → SPENT, both gates firing independently; a second known-bad worktree → SPENT; an ordinary crashed polecat → SALVAGE, unchanged and still automatic. The common rescue path is not affected.

---

## 2. Wisp reconcile is blind to the wisps it reconciles

**Line 534**, `next-iteration`:

```bash
OPEN_WISPS=$(gc bd list --assignee="$GC_AGENT" --status=open --type=molecule --limit=0 --json | jq -r '.[].id')
```

`gc bd list` excludes ephemeral beads by default, and wisps are ephemeral. So "keep one, burn the surplus" never fires — each restart pours a fresh wisp and leaks the previously queued one. Unbounded.

Confirmed by three independent reproductions (witness, deacon, mayor). Same bead, same moment:

```
$ gc bd list --status open --limit=20
No issues found.

$ gc bd query --json 'ephemeral=true AND status=open'
<rig>-wisp-a25  open  assignee=<rig>/gastown.witness  mol-witness-patrol
```

Scale in one city: `bd list` flagless returns 53 rows; `bd list --all --include-infra` returns 921. The flagless form hides ~94% of the ledger.

**This file already knows the right idiom** — lines 162 and 196 use `--include-infra --include-gates --all` for session beads. It just isn't applied here.

Two notes on the fix:
- The status filter needs widening too. A freshly-poured successor wisp is `status=open`, so a fallback filtering `status=in_progress` alone misses it. Use `ephemeral=true AND status IN (open,in_progress)`.
- There's no `--include-ephemeral` on `gc bd list` (checked `--help`), and `--wisp-type` doesn't expose them either. `gc bd query 'ephemeral=true'` was the only working path when we looked.

Related: the deacon patrol's equivalent step uses `--type=wisp`, which is not a valid type — so its burn is unreachable and it leaks one wisp per cycle. Worth checking that recipe in the same pass.

---

## 3. `check-refinery` Step 2 is unexecutable as written

**Lines 415–419:**

> **Step 2: Assess refinery wisp freshness:**
> Check when the refinery last completed a patrol cycle by looking at recent burned wisps. If the refinery has work assigned but its last wisp is stale, the refinery may be stuck.

Step 2 supplies **no command at all** — Steps 3 and 4 both have bash blocks, Step 2 has only prose. The agent is asked to judge wisp freshness with no wisp-visible query in scope, and the queries that are in scope are flagless and cannot see wisps.

Measured on one population: flagless → 0 rows, `--include-infra` → 2 rows.

This produced a real false report. A confident *"refinery has no wisps"* assessment hid a leaked duplicate refinery wisp. Note the failure shape: the step didn't error, it reported **health**. A query whose universe excludes the subject returns "all clear" rather than "cannot answer", and the agent can't tell those apart.

Suggested: give Step 2 an explicit wisp-visible query, and have it distinguish "no wisps found" from "query cannot see wisps".

---

## The common thread

#2 and #3 are the same bug in different places, and #1 is its destructive cousin: **a query whose universe is narrower than the guarantee it's used to make, where the narrowness reads as good news.** `bd list` hiding closed beads is what makes the salvage gate fail open; `bd list` hiding ephemeral beads is what makes reconcile and freshness-check fail open.

If a broader fix is in scope, the defaults are worth revisiting — every one of these is a call site that reasonably assumed the default query showed everything relevant.

Versions: pack pinned at `33d3a430`, defects re-verified against HEAD `aab8030d`.

### LINKED ISSUE #353: mol-witness-patrol check-polecat-health: both prescribed liveness signals are unreliable — polecats have no wisps, bead UpdatedAt goes stale on healthy agents
`gastown/formulas/mol-witness-patrol.toml`, step `check-polecat-health`.

The step names two progress signals. In a live city one of them does not
exist, and the other is misleading. Both verified on rig `symphony`
(patrol 2026-08-23T22:22Z, re-verified 2026-08-25T17:5xZ).

## Signal 1: "the polecat's molecule wisp freshness" — does not exist

> Check the work bead's `UpdatedAt` timestamp and the polecat's molecule
> wisp freshness.
> ...
> - **Very stale, no wisp progress** → likely stuck

Polecats carry no wisps. Full ephemeral census, 2026-08-23T22:22Z:

```
  sy-wisp-pap  symphony/gastown.refinery  mol-refinery-patrol
  sy-wisp-c93  symphony/gastown.refinery  mol-refinery-patrol
  sy-wisp-t4s  symphony/gastown.refinery  mol-refinery-patrol
  sy-wisp-fq0  symphony/gastown.refinery  mol-refinery-patrol
  sy-wisp-mke  symphony/gastown.witness   mol-witness-patrol
  sy-wisp-clx  (unassigned)               sling-context: ...
```

Only the refinery and the witness hold wisps. Neither active polecat had
one. Re-checked 2026-08-25: same result, still zero polecat wisps.

So "no wisp progress" is vacuously true for every polecat at every
moment — and the step lists it as a *stuck* indicator.

## Signal 2: work bead `UpdatedAt` — stale on healthy agents

`UpdatedAt` tracks bead mutations, not work. A polecat updates its bead
at claim and again at handoff and can legitimately go a long time
between. Measured at 22:22Z:

```
sy-xjpz  bead UpdatedAt 21:27Z  -> 54 min "stale"
sy-hfbk  bead UpdatedAt 21:55Z  -> 26 min "stale"
```

Both polecats were working hard at that exact moment:

```
sy-xjpz worktree source mtimes: 22:22:07Z, 22:21:47Z, 22:19:37Z
  session peek: fixing a live defect in the model compiler path,
  4 shells, 57m45s elapsed, actively editing
sy-hfbk worktree source mtimes: 22:21:43Z, 22:12:27Z
  session peek: adding web tests, lint passing, 31m elapsed, editing
```

The two prescribed signals combined read "54 minutes stale, no wisp
progress" for a polecat that had modified a source file 25 seconds
earlier.

## Consequence

An agent following the recipe literally has no positive evidence of life
available to it, and both named signals point at "stuck". The step's
remedy is to file a warrant, which routes to the dog pool and runs
`mol-shutdown-dance` against the agent. That is a real cost paid by a
productive polecat: the dance gives 3 chances to respond, but a polecat
deep in a long build or test run may not answer promptly.

The step is saved today only by its own hedge ("use judgment", "no
hardcoded thresholds"). That hedge is doing all the work, with no
reliable signal behind it.

## Signals that do work

None of these is mentioned in the step.

**1. `gc session list --json` → `last_active`.** First-class, structured,
per-agent, second-precision. Verified it tracks real session output and
is not a uniform heartbeat — sampled 2026-08-25T17:53:07Z:

```
symphony/gastown.furiosa  2026-08-25T17:53:08Z   (mid-turn)
symphony/gastown.nux      2026-08-25T17:53:08Z   (mid-turn)
symphony/gastown.refinery 2026-08-25T17:49:13Z
symphony/gastown.witness  2026-08-25T17:27:29Z   (idle 26m)
gastown.deacon            2026-08-25T17:25:18Z   (idle 28m)
```

Working agents advance every few seconds; idle ones go stale. Note this
proves *life*, not *progress* — a tight loop keeps it fresh too.

**2. Worktree source-file mtimes.** Direct evidence of edits, precise to
seconds. Exclude `obj/`, `bin/`, `node_modules/` — build output moves
without agent progress.

**3. `gc session peek <agent> --lines 20`.** Shows the live turn, elapsed
time, and whether the agent is mid-tool-call. The only signal that
separates "working hard on something slow" from "spinning".

## Two secondary defects in the same step

**Step 1 is not executable as written.** It says "Filter for beads
assigned to polecat-pattern agents in YOUR rig", but the bead assignee is
the session *name* (`gastown__polecat-ga-nw1n`) which carries no rig. The
rig only appears in `gc session list --json`. The join key is
`session_name`, also present on the bead as `metadata.gc.session_name`.

**The worktree path in bead metadata is not what the step would need.**
Live polecat beads carry `metadata.gc.work_dir` pointing at the *shared
pool home* — two polecats can occupy the same one — and
`metadata.work_dir` (the per-bead worktree the contract says branch-setup
records) is frequently never written. Observed on `sy-7pw5`: worktree
`.../gastown.polecat/worktrees/sy-7pw5` exists on branch
`polecat/sy-7pw5`, but the bead has no `work_dir` key. The reliable
derivation is `<session work_dir>/worktrees/<bead-id>`.

## Proposed fix

Patch below. It replaces Steps 1–4 of `check-polecat-health`:

- deletes the wisp-freshness signal outright, with a note saying why
- makes `last_active`, worktree mtimes and `session peek` the primary
  signals, in that order, and says what each does and does not prove
- demotes bead `UpdatedAt` to an explicit weak hint
- makes Step 1 executable by naming the `session_name` join key
- adds a fail-closed gate on Step 3: no warrant without a `session peek`
  in the same cycle showing no forward motion
- routes "in_progress bead with no live session" to
  `recover-orphaned-beads`, where it belongs, instead of here
- asks Step 4 to record the three signals so the next cycle can tell
  "quiet for one cycle" from "quiet for six"

All new shell is backslash-free on purpose. `description` is a TOML
multi-line *basic* string, so escapes are processed: the existing
`gc bd create ... \` continuations in this step are silently collapsed
onto one line in the rendered output, and any `\(`, `\.` or `\|` would
be an invalid escape. The patch uses `[...] | @tsv` for jq, repeats
`-type f` per `-o` branch instead of `\( ... \)`, and uses
`xargs stat -c '%y %n'` instead of `find -printf '...\n'`.

Verified: patched file parses (`tomllib`), 5 steps, ids unchanged.

## Downstream note

This city cannot serve the fix to itself through `gc formula show`. That
command resolves from the pack cache, which is content-addressed and
actively restored — a local edit there is reverted within about a minute
(observed). `<city>/formulas/` is on the search path but loses precedence
to the pack cache. Only `bd formula` (`.beads/formulas/`) can be
overridden locally, so the two read paths now disagree until this lands.
That is a `gastownhall/gascity` matter, filed separately downstream; it
is mentioned here only because it is why this needs to land upstream to
actually take effect.

Related: #351 (same file, same genre — a witness safety mechanism whose
documented procedure can damage healthy work).

Filed from rig `symphony`, bead `sy-p997`.

## Replay against a live city

Ran the patched step's own commands verbatim on rig `symphony`,
2026-08-25T18:05:17Z, against the two in_progress polecat beads.

```
                 bead UpdatedAt   last_active    newest worktree edit
sy-p997 (me)     17:49:23Z        18:05:19Z      (no per-bead worktree)
sy-7pw5 (nux)    18:05:19Z        18:05:19Z      17:52:41Z
```

`sy-p997` is the discriminating case, and it is this bead's own session.
At 18:05 its bead had not been touched for 16 minutes, it had no wisp
(no polecat does), and it had no per-bead worktree to show mtimes for —
this is a formula-only fix, so no branch was ever cut.

- Old step: "very stale, no wisp progress" → likely stuck → warrant →
  `mol-shutdown-dance` against an agent that was mid-implementation of
  this very patch.
- Patched step: `last_active` 0s ago → alive; absent worktree explicitly
  flagged weak evidence; Step 3's gate blocks the warrant without a
  `session peek` showing no forward motion → skip. Correct.

`sy-7pw5` is the ordinary case and is unchanged by the patch: fresh on
every signal, skipped by both versions.

The missing-worktree path was the one rough edge the replay found — a
bare `git -C` on an absent directory printed `fatal: cannot change to
...`, which reads like a failure rather than "nothing to see". The patch
guards both probes with `test -d` and prints an explicit weak-evidence
note instead.

## Patch

```diff
--- a/gastown/formulas/mol-witness-patrol.toml
+++ b/gastown/formulas/mol-witness-patrol.toml
@@ -445,20 +445,78 @@
 be alive and stuck — infinite loop, blocked on something, or just not
 progressing. The controller can't detect this; it's a work-layer judgment.
 
-**Step 1: Find active polecat work beads:**
+**Step 1: Find active polecat work beads and map them to live sessions:**
 ```bash
 gc bd list --status=in_progress --json --limit=0
+gc session list --json | jq -r '.sessions[] | select(.template | test("polecat")) | [.rig, .agent_name, .id, .session_name, .state, .last_active, .work_dir] | @tsv'
 ```
-Filter for beads assigned to polecat-pattern agents in YOUR rig.
-
-**Step 2: For each, assess progress:**
-
-Check the work bead's `UpdatedAt` timestamp and the polecat's molecule
-wisp freshness. Consider:
+Join the two on the bead's assignee. That assignee is the session NAME
+(for example `gastown__polecat-ga-nw1n`), not the agent alias, and it
+carries no rig — so you cannot filter by rig from the bead alone. The
+session row supplies `rig`, `agent_name` and `work_dir`; the bead's
+`metadata.gc.session_name` and `metadata.gc.session_id` are the same
+key. Keep only rows whose `rig` is YOURS.
+
+An in_progress bead whose session_name matches no live session is NOT a
+stuck polecat — it is an orphan, and `recover-orphaned-beads` owns it.
+
+**Step 2: For each, assess liveness and progress — in this order:**
+
+Do NOT look for "molecule wisp freshness" on a polecat. Polecats carry
+no wisps at all; only the witness and the refinery hold them. "No wisp
+progress" is therefore vacuously true of every polecat at every moment
+and is evidence of nothing (sy-p997, verified against a full ephemeral
+census).
+
+Signal 1 — session liveness. PRIMARY, cheap, structured.
+`last_active` from the Step 1 session query is real session output
+activity, not a heartbeat: idle agents go stale while working ones
+advance every few seconds. Compare it against now. Fresh means the
+agent is emitting, so it is alive. It does NOT mean the agent is
+progressing — a tight loop keeps `last_active` fresh too. That is what
+Signals 2 and 3 are for.
+
+Signal 2 — worktree edits. PRIMARY evidence of actual progress.
+The per-bead worktree is `<session work_dir>/worktrees/<bead-id>`, on
+branch `polecat/<bead-id>`. Derive it that way rather than trusting bead
+metadata: `metadata.gc.work_dir` records the shared pool home, which two
+polecats can occupy at once, and `metadata.work_dir` is frequently never
+written at all.
+```bash
+WT="<session work_dir>/worktrees/<bead-id>"
+test -d "$WT" || echo "NOTE: no per-bead worktree yet — weak evidence, never grounds for a warrant on its own"
+find "$WT/src" "$WT/web" "$WT/tests" -type f -name '*.cs' -o -type f -name '*.ts' -o -type f -name '*.tsx' 2>/dev/null | grep -Ev '/(obj|bin|node_modules)/' | xargs -r stat -c '%y %n' | sort -r | head -5
+test -d "$WT" && git -C "$WT" status --porcelain | head -5
+```
+Adjust the extensions to your rig's languages. Exclude obj/, bin/ and
+node_modules/ — build output moves without any agent progress. A recent
+source mtime is direct, second-precision evidence of work. Its absence
+is only weak evidence: an agent can legitimately spend a long stretch
+reading, building, or running a test suite without writing a file.
 
-- **Recently updated** → making progress, skip
-- **Stale but agent is in a long tool call** → might be fine, use judgment
-- **Very stale, no wisp progress** → likely stuck
+Signal 3 — the live turn. DECISIVE, and required before any warrant.
+```bash
+gc session peek <agent-name> --lines 20
+```
+This shows what the agent is doing right now, how long the current turn
+has been running, and whether it is mid-tool-call. It is the only signal
+that separates "working hard on something slow" from "spinning".
+
+Weak hint — the work bead's `UpdatedAt`. A tiebreaker, never grounds on
+its own. It tracks bead MUTATIONS, not work: a polecat touches its bead
+at claim and again at handoff and can legitimately go an hour between.
+During the patrol that filed sy-p997, two polecats read 54 and 26
+minutes "stale" while both had edited source seconds earlier.
+
+Reaching a verdict:
+
+- Recent worktree edits, or a peek showing forward motion → progressing, skip
+- `last_active` fresh, no edits, peek shows one coherent long-running
+  action such as a build, a test suite or a large read → fine, note it and move on
+- `last_active` fresh, no edits, peek shows the same action repeating
+  with no forward motion across two checks → likely stuck
+- `last_active` stale AND no recent edits AND peek shows a dead or
+  wedged prompt → likely stuck
 
 There are no hardcoded thresholds. Consider the nature of the work, the
 time elapsed, and whether the agent shows any signs of activity. This is
@@ -466,22 +524,32 @@
 
 **Step 3: For stuck polecats, file a warrant:**
 
+Fail-closed gate: file a warrant ONLY if you ran `gc session peek` on
+that agent during THIS cycle and it showed no forward motion. A warrant
+routes to the dog pool and runs `mol-shutdown-dance` against what may
+well be a healthy agent, and that costs a real interrogation. A stale
+`UpdatedAt`, a missing wisp, or a quiet worktree are each, alone, never
+sufficient.
+
 Do NOT kill the agent directly. File a warrant bead and let the dog pool
 handle the shutdown dance (multi-stage interrogation with due process):
 
 ```bash
-gc bd create --type=task \
-  --title="Stuck: <agent>" \
-  --metadata '{"target":"<session-name>","reason":"No progress on <bead> for <duration>","requester":"witness","gc.routed_to":"{{binding_prefix}}dog"}' \
-  --label=warrant
+gc bd create --type=task --title="Stuck: <agent>" --metadata '{"target":"<session-name>","reason":"<what peek showed, and what it still showed on recheck>","last_active":"<timestamp>","last_edit":"<timestamp or none>","requester":"witness","gc.routed_to":"{{binding_prefix}}dog"}' --label=warrant
 ```
 
+Put the evidence in the metadata. The dog runs the dance against a live
+agent and needs to know what you actually observed.
+
 The dog pool picks up the warrant and runs `mol-shutdown-dance`, which
 gives the polecat 3 chances to prove it's alive before killing it.
 
 **Step 4: Log assessment:**
 
-Note which polecats were checked and their status for observability.
+Note which polecats were checked and their status for observability. For
+each, record the three signals you read — `last_active`, the newest
+worktree edit, and what peek showed — so the next cycle can tell "quiet
+for one cycle" from "quiet for six".
 Don't escalate to mayor for individual stuck polecats — the warrant
 system handles it. Only escalate if you see a pattern (many stuck
 polecats, systemic issue).
```

tmwells:
Fixes three defects in `mol-witness-patrol`, all found by a live witness on a running city, plus the matching wisp queries in the witness prompt. Same file, same pack, so they land together.

Closes #353. Closes #351. Addresses #252 and #324.

## 1. `check-polecat-health`: both prescribed liveness signals were unreliable (#353)

The step named two progress signals. One does not exist; the other is misleading.

**"Molecule wisp freshness" does not exist for polecats.** A full ephemeral census found wisps only on the witness and the refinery — neither active polecat had one. So "no wisp progress" was *vacuously true of every polecat at every moment*, and the step listed it as a **stuck** indicator.

**Bead `UpdatedAt` is stale on healthy agents.** It tracks bead mutations, not work: a polecat touches its bead at claim and again at handoff. Measured during the patrol that filed this:

| bead | UpdatedAt | actually |
|---|---|---|
| `sy-xjpz` | 54 min stale | editing the model compiler, source written 25s earlier |
| `sy-hfbk` | 26 min stale | adding web tests, lint passing, actively editing |

Combined, the two signals read *"54 minutes stale, no wisp progress"* for a polecat that had modified a source file 25 seconds earlier. The step's remedy is a warrant, which routes to the dog pool and runs `mol-shutdown-dance` against the agent. Only the step's own hedge ("use judgment") stood between a productive polecat and an interrogation.

**Fix.** The wisp signal is deleted outright, with a note saying why. New signal order, each labelled with what it does and does *not* prove:

1. **`gc session list` → `last_active`** — PRIMARY. Proves *life*, not progress.
2. **Worktree source mtimes** — PRIMARY evidence of actual progress. Absence is explicitly weak.
3. **`gc session peek`** — DECISIVE, and now required before any warrant. The only signal separating "slow" from "spinning".

`UpdatedAt` is demoted to a weak hint. Step 3 gains a **fail-closed gate**: no warrant without a peek in the same cycle showing no forward motion.

`last_active` was checked for the opposite failure — a heartbeat that ticks regardless would be a false-life signal. It is not one: sampled at one instant, working agents read 0s while an idle witness read 26m and the deacon 28m. It tracks real session output.

Two secondary defects fixed in the same step:

- **Step 1 was not executable.** It said to filter by rig, but the bead assignee is the session *name* (`gastown__polecat-ga-nw1n`), which carries no rig. It now joins against `gc session list` on that key.
- **The worktree lookup was wrong.** Live beads carry `metadata.gc.work_dir` = the *shared pool home* (two polecats occupy one), and `metadata.work_dir` is frequently never written. Derivation is from the session row instead.
- "in_progress bead, no live session" is routed to `recover-orphaned-beads`, where it belongs, rather than being treated as a stuck polecat here.

**Replay against the patched step.** The discriminating case was the session writing the fix: 16 minutes of bead staleness, no wisp, no worktree. The old step reads that as "very stale, no wisp progress" → warrant → shutdown dance against an agent mid-fix. The patched step reads `last_active` 0s → alive, flags the absent worktree as weak, and the Step 3 gate blocks the warrant. An ordinary working polecat is skipped by both versions.

## 2. `recover-orphaned-beads`: Case C/D salvage could publish a revert (#351)

"All work is useful work — never discard" was implemented as "publish whatever is dirty". A worktree can hold NEW work or SPENT work — an older base checked out, a half-applied rebase, or a commit whose bead the refinery already squash-merged. Publishing spent content is not a rescue; it is a revert authored by the safety mechanism.

- **Guard 0** — fail closed when `work_dir` does not resolve. `cd ""` *succeeds* in bash and leaves you where you were, so an unguarded `cd "$WORKTREE"` silently no-ops and every following `git add -A` / `commit` / `push` runs against the witness's own checkout. Every git call is now addressed with `git -C`.
- **Gate 1 (primary)** — resolve the worktree back to its owning bead; a closed bead has nothing to rescue. Uses `--all`, because the flagless form **hides closed beads** — the exact fail-open this gate exists to prevent.
- **Gate 2 (backstop)** — net deletion measured `git diff origin/main` against the **working tree**, not HEAD. A worktree can be perfectly clean and still hold a commit that reverts main; against HEAD that reads +0/−0 and the gate waves it through.

Ancestry and blob-vs-ancestor matching are **advisory only** — both fail open (ancestry under squash-merge; blob matching as soon as one path matches no commit in main), so neither may decide. Withheld content is preserved non-destructively at `refs/witness/salvage/<bead>` via `git stash create`, never on a `polecat/*` branch the refinery may be about to merge.

Replayed against both known traps: each classified SPENT and withheld; the ordinary crashed-polecat rescue still classifies as salvage and stays automatic. With Gate 1 forced open, Gate 2 alone still holds both traps.

Also fixes the metadata key: the contract field is `work_dir`, and `.worktree` reads null on every live bead.

## 3. Wisp reconciliation is blind without `--include-infra` (#252, #324)

Molecule roots are infra beads, so `gc bd list --type=molecule` returns nothing without `--include-infra`. The reconcile-to-exactly-one guard therefore never fired and one wisp leaked per restart. The flag is added to the `next-iteration` query and to the four queries in `agents/witness/prompt.template.md`, with the reason recorded inline so it does not get dropped again.

## Rebased onto current `main`

The fixes were authored against the pinned pack (`33d3a43`) and **re-applied onto `main`**, which has since moved. Deliberately preserved rather than reverted:

- the warrant **dedup** block and `--labels=warrant` (#330) — the new fail-closed gate is layered *in front of* it, not instead of it
- the cross-actor `gc bd close <bead> --force` (beads#3734)
- the "not a valid **gc bd** type" wording
- beads commands routed through `gc`, never bare `bd`

New shell is deliberately **backslash-free**: `description` is a TOML multi-line basic string, so line-ending backslashes are processed and silently collapse the rendered command onto one line. (This is pre-existing and harmless — the collapsed form is still valid shell — so existing continuations are left alone.)

## Verification

- `gc lint gastown` — findings **byte-identical** before and after (27 → 27), so the pinned set in `tests/gastown_lint_upstream_defects.txt` is unaffected
- `tests/test_no_bare_bd_commands.py` — passes (both test functions, run directly)
- all four `gastown/tests/test_*.sh` — pass
- TOML parses; step ids, ordering, `needs` and all non-description keys unchanged
- `FAIL-SAFE: empty liveness map`, asserted by `test_gascity_pack_inference_gate.py`, still present

`pytest` was unavailable in the authoring environment, so the two suites that require it were not run locally; CI covers them.

### DIFF
--- a/gastown/agents/witness/prompt.template.md
+++ b/gastown/agents/witness/prompt.template.md
@@ -164,10 +164,12 @@ nothing).
 # Step 1: Reconcile your patrol wisps to exactly one (town ledger, via gc bd).
 # Collect every open/in_progress patrol wisp assigned to you, keep one, and
 # burn the surplus so restarts never accumulate duplicates. Wisp roots are
-# molecules — filter --type=molecule, never --type=wisp.
+# molecules — filter --type=molecule, never --type=wisp. Molecule roots are
+# infra beads, so --include-infra is REQUIRED: without it gc bd list returns
+# nothing and every restart pours a duplicate wisp.
 WISP_IDS=$(
-  gc bd list --assignee="$GC_AGENT" --status=in_progress --type=molecule --limit=0 --json | jq -r '.[].id'
-  gc bd list --assignee="$GC_AGENT" --status=open --type=molecule --limit=0 --json | jq -r '.[].id'
+  gc bd list --assignee="$GC_AGENT" --status=in_progress --type=molecule --limit=0 --include-infra --json | jq -r '.[].id'
+  gc bd list --assignee="$GC_AGENT" --status=open --type=molecule --limit=0 --include-infra --json | jq -r '.[].id'
 )
 WISP=$(printf '%s\n' $WISP_IDS | sed -n '1p')           # keep one (prefers in_progress)
 for extra in $(printf '%s\n' $WISP_IDS | sed '1d'); do  # burn any surplus
@@ -203,14 +205,15 @@ If `next-iteration` already ran, do not pour again; run `gc hook`.
 ```bash
 CURRENT_WISP=${GC_BEAD_ID:-}
 if [ -z "$CURRENT_WISP" ]; then
-  CURRENT_WISP=$(gc bd list --assignee="$GC_AGENT" --status=in_progress --type=molecule --limit=1 --json | jq -r '.[0].id // empty')
+  CURRENT_WISP=$(gc bd list --assignee="$GC_AGENT" --status=in_progress --type=molecule --limit=1 --include-infra --json | jq -r '.[0].id // empty')
 fi
 # Reconcile queued (open) patrol wisps to exactly one. A prior cycle may have
 # poured a next wisp without burning, or a restart may have raced — keep the
 # first and burn the surplus so wisps never accumulate. Wisp roots are
 # molecules (never --type=wisp, which is not a valid gc bd type and matches
-# nothing).
-OPEN_WISPS=$(gc bd list --assignee="$GC_AGENT" --status=open --type=molecule --limit=0 --json | jq -r '.[].id')
+# nothing). Molecule roots are infra beads — --include-infra is REQUIRED, or
+# the query returns nothing and the reconcile silently no-ops.
+OPEN_WISPS=$(gc bd list --assignee="$GC_AGENT" --status=open --type=molecule --limit=0 --include-infra --json | jq -r '.[].id')
 ASSIGNED_WISP=$(printf '%s\n' $OPEN_WISPS | sed -n '1p')
 for extra in $(printf '%s\n' $OPEN_WISPS | sed '1d'); do
   gc bd mol burn "$extra" --force
--- a/gastown/formulas/mol-witness-patrol.toml
+++ b/gastown/formulas/mol-witness-patrol.toml
@@ -257,11 +257,42 @@ lifecycle directly.
 The canonical work chain: `worktree → (push) → branch → (merge) → target`.
 Our job is to ensure work reaches the branch so it's schedulable.
 
+**"All work is useful work — never discard" does not mean "publish whatever is
+dirty."** A worktree can hold NEW work (salvage it) or SPENT work — an older
+base checked out, a half-applied rebase left staged, or a commit whose bead the
+refinery already squash-merged. Publishing spent content is not a rescue; it is
+a revert authored by the safety mechanism, and it loses work louder than
+discarding would. Classify BEFORE you commit anything.
+
 Read the bead metadata to find the worktree and branch:
 ```bash
 META=$(gc bd show <bead> --json | jq '.[0].metadata')
-WORKTREE=$(echo "$META" | jq -r '.worktree // empty')
+# The metadata contract field is work_dir, set by the polecat's branch-setup
+# step. `.worktree` is a legacy misspelling and reads null on every live bead —
+# keep it only as a fallback.
+WORKTREE=$(echo "$META" | jq -r '.work_dir // .worktree // empty')
 BRANCH=$(echo "$META" | jq -r '.branch // empty')
+```
+
+**Guard 0 — never let a salvage git command run outside a resolved worktree.**
+`cd ""` SUCCEEDS in bash and leaves you exactly where you already were. If
+WORKTREE does not resolve, an unguarded `cd "$WORKTREE"` silently no-ops and
+every following `git add -A` / `git commit` / `git push` runs against YOUR OWN
+checkout — and this prompt puts you in the canonical rig repo for repo work. The
+result is the wrong repo's dirt published under a rescue message. Fail closed,
+and address the worktree explicitly with `git -C` so a resolution failure can
+never inherit your cwd:
+
+```bash
+if [ -z "$WORKTREE" ] || [ ! -d "$WORKTREE" ]; then
+  echo "SALVAGE REFUSED (<bead>): no resolvable work_dir — treat as Case E."
+  # Go to Case E. Run NO git command.
+fi
+TOP=$(git -C "$WORKTREE" rev-parse --show-toplevel 2>/dev/null)
+if [ -z "$TOP" ] || [ "$(readlink -f "$TOP")" != "$(readlink -f "$WORKTREE")" ]; then
+  echo "SALVAGE REFUSED (<bead>): work_dir is not a worktree root — Case E."
+  # Go to Case E. Run NO git command.
+fi
 ```
 
 Check the bead and worktree state, then act:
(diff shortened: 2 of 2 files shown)


The1nk:
Adopted Guard 0 downstream and it holds — but its stated rationale is inverted for bash, and that's the half a future reader will trust, so it seemed worth correcting before it propagates.

The PR says:

> `cd ""` *succeeds* in bash and leaves you exactly where you already were.

Measured just now, same box, same shell, `cd /tmp; W=""; cd "$W"; echo $?; pwd`:

| shell | `cd ""` | pwd |
| --- | --- | --- |
| bash 5.3.9(1) | **rc=1** | unchanged |
| bash --posix | rc=1 | unchanged |
| sh (dash 0.5.12) | rc=0 | unchanged |
| dash 0.5.12 | rc=0 | unchanged |

So it's the other way round: bash *fails* the `cd`, dash/sh succeed. POSIX leaves an empty operand unspecified, so both are conformant — which is exactly why it's worth stating per-shell rather than as "bash does X".

**The guard is right and the code is already correct — only the sentence needs changing.** The half that creates the hazard is not the exit status, it's `pwd` being unchanged, and that holds in *every* shell above. And because this is a prompt step — prose an agent executes, not a script under `set -e` — nothing aborts on a non-zero `cd`, so the following `git add -A` / `git commit` / `git push` run against the witness's own checkout whether `cd` returned 0 or 1.

That is precisely why testing `[ -z "$WORKTREE" ]`, as the block already does, is the correct test and `cd`'s exit status is not.

Suggested rewording:

> If `WORKTREE` does not resolve, `cd "$WORKTREE"` leaves you exactly where you already were: `cd ""` fails in bash (rc 1) and succeeds in dash, but in **both** the working directory is unchanged. This is a step an agent executes, not a script under `set -e`, so nothing aborts on the failure — every following `git add -A` / `git commit` / `git push` runs against YOUR OWN checkout.

Flagging it because someone trusting the current sentence might reasonably "simplify" the guard to key on `cd`'s exit status. That would fail closed in bash and fail **open** in dash — the one direction this guard exists to prevent.

Everything else in section 2 replayed cleanly for us; the emptiness test and the `git -C` addressing are what we grafted.

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277250343

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277250416

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277286476

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277286545

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277286614

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277307110

PR #361, opened 2026-08-27
fix(gastown): reconcile ephemeral patrol wisps

kaushikNaarayan:
Fixes patrol reconciliation queries that used `gc bd list`, whose projection excludes ephemeral wisp roots. Replaces the copied witness, refinery, and deacon lookups with `gc bd query` filtered by `ephemeral=true`, status, and assignee. Adds a regression test covering every affected prompt/formula.\n\nTests: `go test ./...`; `make registry-format-validate`.


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277250189

julianknutsen:
Let's go ahead and use the bd command flags that allow ephemeral to show up or fix the beads code if there is a bug. We want to move towards deprecating bd query since it is hard ot maintain over time.

PR #363, opened 2026-08-27
fix(slack): always derive inbound attachment mime_type — Slack-native recordings carry none

aphexcx:
## Symptom

A Slack **voice clip** (or video clip) posted into a mapped channel never reaches the bound gc session — and neither does its text caption. The adapter logs `inbound POST failed: 422 ...` and drops the message on the floor. In our fork, whose inbound pipeline retries instead of dropping, the same root cause was worse: one founder voice memo wedged a production channel's inbound delivery for ~34 hours until a human intervened. Either way, the trigger is any file recorded *inside* Slack.

## Root cause

Files recorded natively in Slack — voice clips (file subtype `slack_audio`, name `audio_message.m4a`) and video clips (`slack_video`) — arrive with an **empty `mimetype` AND empty `filetype`**; `files.info` confirms Slack simply never sets one for these. `downloadSlackFiles` passed that empty value through verbatim into `externalAttachment.MIMEType`, which is tagged `omitempty`, so the inbound payload omitted the `mime_type` key entirely. gc's `extmsg.ExternalAttachment` declares `mime_type` **required**, so gc deterministically answers 422 and the whole `externalInboundMessage` — attachments and text alike — is rejected.

## Fix

- New `attachmentMIMEType(f slackFile)` derives a media type for every inbound file, precedence: Slack's own `mimetype` → the file name's extension (a pinned table of the types Slack clients actually produce, then the host's `mime.TypeByExtension` with parameters stripped) → the Slack `filetype` code treated as an extension → the Slack-native subtype (`slack_audio`/`slack_video` → `audio/mp4`/`video/mp4`) → `application/octet-stream`. It never returns `""`.
- `externalAttachment.MIMEType` drops `omitempty`: the key is always serialized, so a payload can never again be rejected for a missing required key (an empty string would be at worst a degraded value, never a 422).
- `slackFile` gains the `Subtype` field — the only hint Slack provides on its native recordings.

## How verified

- Table test over the derivation precedence, including the fall-through boundaries (unknown name extension → title, unknown filetype → subtype) and canonicalization edges (trailing whitespace in the name, already-dotted filetype codes, case-insensitivity).
- A voice-clip download test reproducing the exact incident shape (`mimetype:"", filetype:"", subtype:"slack_audio"`) and asserting the wire record carries `audio/mp4`.
- A serialization guard asserting `mime_type` is present in the JSON even when empty.
- `gofmt`/`go vet` clean; the full adapter suite passes (run on linux/amd64, Go 1.25).
- The same fix (same derivation table and precedence) has been running in production in our fork since 2026-08-27 after three rounds of adversarial review there; this PR ports the transport-agnostic root-cause half, leaving out the fork-specific retry/dead-letter machinery that doesn't apply to this codebase's drop-on-error inbound path.

### DIFF
--- a/slack-full/CHANGELOG.md
+++ b/slack-full/CHANGELOG.md
@@ -21,6 +21,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `filetype` code → Slack-native subtype → `application/octet-stream`)
   and always serializes `mime_type` (no longer `omitempty`), so a
   payload can never again be rejected for a missing key.
+  Scope: this covers the legacy `/extmsg/inbound` delivery path. Files
+  shared in a company/multi-bot room take the company hydration path,
+  which never posts to `/extmsg/inbound` and still renders Slack's raw
+  `filetype`, so a Slack-native recording there is unchanged by this fix
+  and still lists an empty filetype.
 
 ### Changed
 
--- a/slack-full/CHANGELOG.md
+++ b/slack-full/CHANGELOG.md
@@ -8,6 +8,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- Inbound files recorded inside Slack itself — voice clips (file subtype
+  `slack_audio`) and video clips (`slack_video`) — arrive with an empty
+  `mimetype` AND `filetype`, which the adapter passed through verbatim.
+  The inbound payload then omitted `mime_type`, a REQUIRED property of
+  gc's `extmsg.ExternalAttachment`, so gc rejected the whole message
+  with 422 and the recording plus its text caption silently never
+  reached the bound session. The adapter now derives a media type for
+  every inbound file (Slack's `mimetype` → file-name extension → Slack
+  `filetype` code → Slack-native subtype → `application/octet-stream`)
+  and always serializes `mime_type` (no longer `omitempty`), so a
+  payload can never again be rejected for a missing key.
+
 ### Changed
 
 - Renamed the pack directory from `slack-pack/` to `slack-full/` and the
--- a/slack-full/adapter/attachment_mime.go
+++ b/slack-full/adapter/attachment_mime.go
@@ -1,0 +1,97 @@
+package main
+
+import (
+	"mime"
+	"path"
+	"strings"
+)
+
+// --- inbound attachment media types ------------------------------------
+//
+// gc's extmsg.ExternalAttachment declares mime_type a REQUIRED property,
+// and Slack does not always populate one: files recorded inside Slack
+// itself — voice clips (file subtype "slack_audio", name
+// "audio_message.m4a") and video clips ("slack_video") — arrive with
+// mimetype "" AND filetype "", even from files.info. Passing Slack's
+// value through verbatim produced a payload without the key, a
+// deterministic 422 from gc, and the whole inbound message — the file
+// AND its text caption — silently never reached the bound session.
+// attachmentMIMEType therefore always derives a value; the adapter
+// never emits an attachment without one.
+
+// fallbackAttachmentMIMEType is the last-resort attachment type when
+// neither Slack nor the file name says anything usable.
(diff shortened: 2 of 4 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277286113

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277286251

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277286182

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @aphexcx! This PR makes the Slack adapter always derive an inbound attachment `mime_type` instead of trusting Slack to supply one. Slack-native recordings — voice clips, screen recordings — carry no `mimetype` and often no `filetype`, so the adapter omitted a key that gc's `ExternalAttachment` schema requires, and gc rejected the entire message with a 422. The recording *and* the human's caption disappeared with nothing but a log line behind them, which is exactly the kind of failure that is miserable to diagnose from the outside. Closing it is a real reliability win for every Slack room where people share voice clips.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- The premise was verified rather than taken on trust: `mime_type` really is in `ExternalAttachment.required`, and removing the fix reproduces the 422 — so this is the right layer to fix it, and nothing downstream was absorbing the failure.
- `attachmentMIMEType` (`attachment_mime.go:52-97`) was confirmed total: Slack `mimetype` → pinned extension table → host MIME database → `application/octet-stream` floor. It cannot return `""`, so the missing-key class cannot reopen through a gap in the derivation. The pinned-table-ahead-of-host-DB ordering was justified by a live probe where the host resolver disagrees on `.heic` and `.opus`.
- Dropping `omitempty` was checked against the schema instead of assumed: `mime_type` carries no `minLength` and no enum, so an empty string is at worst a degraded value and never a 422. Your serialization guard test pins that behavior so a future "cleanup" cannot quietly restore the bug.
- The rest of the endpoint was swept for sibling instances of the same class: every required property of `ExternalInboundMessage`, `ConversationRef`, and `ExternalActor` is present in the adapter tags, and `slackKindFromChannelType` is total over the `ConversationKind` enum.
- The new `slackFile.Subtype` field was cleared of persistence and replay hazards: receipts digest the stored raw body bytes, so a decode-only field cannot shift any digest, and there is no unkeyed `slackFile{...}` literal that a field reorder could break.
- Build and test evidence was reproduced independently by both reviewers: `go build ./...` and `go vet ./...` clean, `go test ./...` green in `slack-full/adapter`, plus your three new tests exercised individually.

Review findings addressed:
- Change Impact / Blast Radius (minor): the CHANGELOG entry read as though the fix covered every Slack file path. `f01c243` scopes it to the legacy `/extmsg/inbound` path, so an operator who sees the same empty-filetype symptom in a company/multi-bot room is not told it is already fixed. The underlying company-path code gap is untouched here and is filed as a follow-up below.

- mc-ko8np: Change Impact / Blast Radius, minor — the company/multi-bot file path never gained the new derivation, so a Slack-native recording shared in a company room still renders with an empty filetype (`company_hydration.go:208-235`). Pre-existing, not introduced by this PR, and correctly out of scope for this merge.
- mc-ofpiz: Error Handling & Resilience, minor — a non-2xx from `postInbound` still discards the human's message with only a log line (`main.go:2587-2589`), so the next required-key drift would be just as silent. That block is byte-identical to `main` and was deliberately left alone here.

### Maintainer Changes
`f01c243 fix(slack): scope the mime_type CHANGELOG entry to the legacy inbound path` — 1 file changed, 5 insertions(+).

### Final Review Status
Ready for the merge queue: final head `f01c243` has passing required GitHub checks, and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/gascity-packs/actions/runs/33895641652

### Review Iterations
1 review pass performed — two independent reviewers plus synthesis, no fix iterations required. The only maintainer-side change is the CHANGELOG scoping commit above.

---

PR #364, opened 2026-08-27
fix(slack): pack pin bumps can no longer strand the adapter — run.sh service entrypoint with build-on-missing

aphexcx:
## Symptom

After a pack pin bump (`gc import install`), `gc service list` shows the `slack` service dead and it never comes back on its own. Nothing is wrong with the pack contents — the service command's target simply no longer exists. We were bitten by this twice in production before fixing it at the root.

## Root cause

The `[[service]]` proxy_process command points straight at `./adapter/gc-slack-adapter`, which is a **gitignored build artifact**. `gc import install` re-materializes the pack cache **git-only**, so every pin bump wipes the binary and leaves the supervisor exec'ing a path that isn't there, until a human notices and rebuilds by hand.

## Fix

The service command becomes `./adapter/run.sh`, which is checked in and therefore survives every materialization:

- **Fast path (idempotent):** an existing executable `adapter/gc-slack-adapter` is exec'd directly — no build, no behavior change for a healthy deployment or a dev checkout.
- **Self-heal:** when the binary is missing, run.sh says so loudly on stderr, finds a Go toolchain (PATH first, then Homebrew/system fallbacks for the minimal PATH of supervisor environments like launchd/systemd), and rebuilds from the sources sitting next to it. The build goes to a PID-suffixed temp file and is published with an atomic rename, so concurrent supervisor restarts can't exec a half-written binary.
- **Env sourcing moves into the entrypoint:** run.sh sources `~/.config/gc-slack-adapter/env` (override: `GC_SLACK_ADAPTER_ENV`) as before, but a missing env file is now a loud warning rather than a hard exit — supervised deployments may inject env another way, and the adapter already fail-fasts at startup with a precise `missing required env vars:` error.

`pack.toml`, `.gitignore`, README and CONTRIBUTING are updated to describe the new scheme.

## How verified

- New hermetic pytest suite (`slack-full/tests/test_adapter_run_sh.py`, runs in this repo's existing CI pytest step): a stub `go` toolchain on PATH records invocations and fakes the build, covering build-on-missing (exactly one build, binary published, no temp files left), exec-without-build when the binary exists, idempotency across restarts, env-file sourcing reaching the adapter process, the missing-env-file warning, HOME-unset startup (supervisors may not set HOME), a stale mid-build temp file being ignored, and a guard that `pack.toml` keeps pointing the service at a checked-in, non-gitignored command (so this outage class can't be silently reintroduced).
- Full `slack-full/tests` suite passes (391 tests); `shellcheck` clean; registry validation passes.
- The same entrypoint design has been running in production in our fork since 2026-08-21 and has survived multiple pin bumps since, including rebuild-from-cold on a supervisor with a minimal PATH.

### DIFF
--- a/slack-full/.gitignore
+++ b/slack-full/.gitignore
@@ -2,9 +2,14 @@ __pycache__/
 *.pyc
 .gc/
 
-# Built adapter binary — produced by `go build` in adapter/ and read
-# in place by the [[service]] proxy_process command.
+# Built adapter binary — produced by `go build` in adapter/ (or by
+# adapter/run.sh's build-on-missing self-heal). The [[service]]
+# proxy_process command is adapter/run.sh (checked in), which rebuilds
+# this binary when a git-only pack re-materialization has wiped it, so
+# the service can't strand. A crash mid-build (SIGKILL, host loss) can
+# leave a PID-suffixed temp behind — ignore those too.
 adapter/gc-slack-adapter
+adapter/gc-slack-adapter.build.*
 
 # Built operator CLI binary — produced by `go build` in cli/ and exec'd
 # by the commands/<cmd>.sh wrappers via $GC_PACK_DIR/cli/gc-slack-cli.
--- a/slack-full/CHANGELOG.md
+++ b/slack-full/CHANGELOG.md
@@ -16,16 +16,48 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   pack pin bump wiped the binary and left the slack service dead until
   someone rebuilt it by hand. The service command is now
   `adapter/run.sh` (checked in, survives every materialization): it
-  sources the secrets env file (warns but continues when the file is
-  absent — the adapter fail-fasts on missing required env itself),
-  execs an existing binary untouched, and on a missing binary finds a
-  Go toolchain (PATH plus Homebrew/system fallbacks for minimal
-  supervisor environments), rebuilds from the colocated sources with
-  loud stderr logging, publishes via PID-suffixed temp file + atomic
-  `mv` (safe under concurrent supervisor restarts), and execs.
+  sources the secrets env file, execs an existing binary untouched,
+  and on a missing binary finds a Go toolchain (PATH plus
+  Homebrew/system fallbacks for minimal supervisor environments),
+  rebuilds from the colocated sources with loud stderr logging,
+  publishes via PID-suffixed temp file + atomic `mv` (safe under
+  concurrent supervisor restarts), and execs.
+
+  Env-file handling distinguishes the two cases. An absent *default*
+  (`~/.config/gc-slack-adapter/env`) warns and continues, since
+  supervised deployments legitimately inject the env another way and
+  the adapter fail-fasts on missing required keys itself. An absent
+  path you set *explicitly* in `GC_SLACK_ADAPTER_ENV` is fatal: the
+  adapter validates that credentials are present, never where they
+  came from, so continuing would boot it against whatever Slack token
+  happens to be in the ambient environment.
+
+  Because the rebuild now sits on the service startup path, it is
+  hardened for the minimal supervisor environments it targets: the
+  toolchain search also covers `/usr/bin/go` and `/bin/go` (where a
+  distro Go lands, and which a restricted PATH hides); `GOCACHE` and
+  `GOPATH` are defaulted under `TMPDIR` when HOME, `XDG_CACHE_HOME`
+  and `GOCACHE` are all unset, which `go build` otherwise rejects
+  outright; and the toolchain is checked against `go.mod`'s `go`
+  directive up front, so a too-old Go under `GOTOOLCHAIN=local` fails
+  with a named remedy instead of a raw compiler error (the default
+  `GOTOOLCHAIN=auto`, which can fetch the required toolchain itself,
+  warns and proceeds).
 
 ### Changed
 
+- **Behavior change on upgrade:** the slack service now *exits 1 at
+  startup* when `GC_SLACK_ADAPTER_ENV` is set to a path that does not
+  exist, where it previously logged one warning and started on the
+  ambient environment. Pointing the variable at a nonexistent path is
+  this pack's own established idiom for "no env file", so a unit file or
+  wrapper using it that way for the **service** will stop starting after
+  this pin bump. Audit before upgrading with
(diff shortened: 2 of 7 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277285897

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277285966

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277286035

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @aphexcx! This PR gives `slack-full` a tracked `run.sh` service entrypoint that rebuilds the adapter binary when it is missing, so a pack pin bump that re-materializes the cache git-only can no longer strand the Slack service. That closes a recurring, silent outage class: previously the service simply stopped coming back after an upgrade and a human had to notice the dead process and rebuild by hand.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- An explicitly-configured `GC_SLACK_ADAPTER_ENV` pointing at a file that does not exist used to start the service anyway, on whatever credentials happened to be in the ambient environment. `run.sh` now hard-fails on that case *before* the exec fast path, so a stale unit file cannot boot the adapter on a decommissioned workspace's token.
- The self-heal's toolchain handling had three real mechanisms to fail on a bare supervisor host: the search skipped `/usr/bin/go` and `/bin/go`, nothing asserted the Go version against `go.mod` before building, and `go build` needed a cache the environment may not provide. All three are handled now — the search covers the standard locations, the version gate hard-fails under `GOTOOLCHAIN=local` and warns-and-proceeds under `auto` with a likely-cause hint on a genuine build failure, and `GOCACHE`/`GOPATH` fall back under `TMPDIR` when nothing else is available.
- Both operational failure paths were untested. The suite went from 8 to 13 tests, and the Go stub gained honest failure knobs (`GO_STUB_VERSION`, `GO_STUB_BUILD_FAILS`) rather than a mock that can only succeed; `test_starts_without_home_set` was re-scoped to assert the build was actually handed a cache.

Review findings addressed:
- **Error handling** — `GOPATH` was defaulted only inside the build-cache guard, but its absence derives from `HOME` alone, so an operator who partially remediated by exporting `GOCACHE` still lost a toolchain fetch to `module cache not found`. `adaff4c` splits it into its own guard and adds a parametrized test covering `{nothing set, XDG_CACHE_HOME set, GOCACHE set}`.
- **Behavioral correctness** — `command -v go` returns the bare name `go` when `go` is a shell function, which the relative-path normalization turned into a nonexistent `<cwd>/go` while suppressing the absolute-path fallback, so a real installed Go went unfound behind a confusing build error. `adaff4c` switches the lookup to `type -P` and adds a test that defines a `go` function in the sourced env file.
- **Debuggability** — with neither `HOME` nor `XDG_CONFIG_HOME` set (the supervisor environment this script targets), the remedy text advertised the root-owned `/.config/gc-slack-adapter/env`, a path the operator cannot create. `adaff4c` leaves the default empty and says there is no usable default path instead, with a test asserting the degenerate path is never printed.
- **Contract fidelity** — the new fatal startup condition was filed under `Fixed`, where an operator auditing a version bump would not look. `adaff4c` adds a `### Changed` entry naming the behavior change and the one-line pre-upgrade audit (`grep -r GC_SLACK_ADAPTER_ENV` across unit files and wrappers), and records in `run.sh` that the strict rule is service-only.

Non-blocking follow-ups (shared for visibility — NOT completed in this PR). Each is filed so it does not get lost:
- `mc-c1tl3` — *Change impact, minor:* the pack's own `doctor/check-binaries.sh` still hard-fails on the missing adapter binary and prescribes the manual rebuild this PR designs away; narrowing it must update `EXPECTED_DOCTOR_DELTA` in the same change. Not addressed here.
- `mc-0o607` — *Contract fidelity, minor:* `GC_SLACK_ADAPTER_ENV` now means two things in one pack — fatal in `run.sh`, a silent no-op in `scripts/slack_intake_common.py` for the `slack_chat_*` commands. This PR documents the split as deliberate and service-only; tightening the command path is a separate decision. Not addressed here.
- `mc-3io0p` — *Change impact, minor:* `slack-mini` and `slack-channel` still point `[[service]]` at gitignored binaries — the same defect class this PR fixes for `slack-full`. Not addressed here.
- `mc-sz5da` — *Behavioral correctness, minor:* the fast path keys on the binary being *missing*, never *stale*, so a binary built from older sources is exec'd unconditionally. Not addressed here.
- `mc-jf3sd` — *Resource lifecycle, minor:* orphaned `gc-slack-adapter.build.*` temps are never swept. The blind `rm -f` was correctly rejected in this PR because it would delete a concurrent build's temp; `adapter/tmp_sweep.go` is the in-tree pattern that solves it safely. Not addressed here.
- `mc-2jptu` — *Operability, nit:* `GOTOOLCHAIN` values other than `local` and `auto` (`path`, an explicit `<name>`) are treated as "Go resolves it itself", but `path` will not download — a comment-accuracy clarification. Not addressed here.
- `mc-c2pdi` — *Carried open question:* the first start after a pin bump now performs a build, and possibly a toolchain download, on the service startup path. Whether a slow first build can be killed into a restart loop depends on the supervisor's start/health deadline, which lives outside this repo, so the premise was left open rather than guessed. Worth deciding deliberately. Not addressed here.

### Maintainer Changes
`adaff4c` — fix(slack-full): harden run.sh env-file and self-heal paths (4 files changed, 400 insertions(+), 41 deletions(-)), covering the env-file hard-fail, the toolchain search and version gate, the `GOCACHE`/`GOPATH` self-heal defaults, the `### Changed` CHANGELOG entry, and the accompanying tests.

### Final Review Status
Ready for the merge queue: final head `adaff4c` has passing required GitHub checks, and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/gascity-packs/actions/runs/33895812085

### Review Iterations
2 review passes. Iteration 1 requested changes on two majors — the ambient-credential boot and the self-heal's toolchain/cache mechanisms — and iteration 2 verified both fixed against the tree and approved. The remaining minors from iteration 2 were applied to the maintainer fixup before finalization, and the adapter test suite was re-run green at each step.

---

PR #365, opened 2026-08-28
witness/refinery: never publish a bead the PLAIN rail halted

edwarddavison:
## The defect

The PLAIN rail has a designed stopping point. A polecat on a no-push rail finishes,
commits to its branch, writes `branch_ready` / `halt_reason`, clears its own assignee,
and hands the bead to the mayor to publish. The branch is deliberately unpublished and
the session is deliberately gone.

Two patrol formulas cannot tell that apart from a crash, because they never look at the
markers the polecat left. Both read the same evidence — dead session, unmerged branch —
and both conclude "repair this".

**First face — the witness (2026-08-27 16:51 AEST, detmold/gastown.witness, co-wisp-lv80).**
`recover-orphaned-beads` found `det-1aa0` and `det-efer` open/unassigned with
`branch_ready=true, halt_reason=auto_push_false`, classified them as orphans, and applied
Case C salvage: **pushed both branches to origin and removed the worktrees.** It preserved
"no refinery handoff" and overrode "no push". `det-efer` was a new BBAgentRead solution
package with unattested privilege names. The witness flagged its own judgment call.

**Second face — the refinery.** Once `polecat/det-efer` was on origin,
`mol-refinery-patrol.find-work` matched it on `has metadata.branch` alone and attempted a
merge; it hit a rebase conflict, was rejected, re-slung to the pool, and a polecat rebased
and correctly re-halted it. `det-1aa0` had no such luck: **merged to `main` at `ff751f4`
at 17:32 08-27 and closed by automation** — an operator-held bead published with no
approval anywhere in the chain. `det-efer` escaped only because it conflicted, and it is
conflict-free now (detmold refinery, co-wisp-k866).

The rule was already written down. It was in memory as *empty assignee + empty
`gc.routed_to` + `branch_ready=true` = FINISHED-AND-HALTED, not undispatched*. No formula
step checked it. **That is the actual lesson: a documented rule that no step executes is
not a control.** Hence the fixture — the rule ships as a predicate that is run, not prose
that is read.

## The rule: two questions, not one

Conflating these is the bug. Each key answers only the question it can answer.

| | asks | true when | who writes it, and when |
|---|---|---|---|
| **`finished`** | did the worker stop **on purpose**? | `halt_reason` present (any value), or `branch_ready` truthy | the polecat, at the END of its run |
| **`publish_blocked`** | may automation push or merge? | `finished`, **or** `auto_push` present and not truthy | `auto_push`: the CALLER, BEFORE dispatch |

- **A finished bead is not an orphan.** Report it; touch nothing. No push, no worktree
  removal, no pool reset, no close.
- **`auto_push=false` alone is not "finished".** It is the `det-d3rn` shape: a standing
  hold set at creation, before any work exists. A worker that dies on such a bead really
  is dead and the bead must become schedulable again — but nothing may leave the box.
  Reading `auto_push` as "finished" would freeze live rigs that run the key as a standing
  convention; reading `branch_ready` as merely "unpublishable" is what let the det-efer
  salvage push and then merge.
- **A merge is a publish.** `metadata.branch` records where the work *is*. It is not
  consent to land it.

We also run a liveness classifier over the same key set, and it treats `auto_push` the
OPPOSITE way — deliberately, not inconsistently. That engine asks *"is this worker
dead?"*, where a caller's pre-dispatch intent is no evidence at all; this gate asks *"may
automation publish?"*, where pre-dispatch intent is exactly the right authority. Same
keys, opposite correct treatment, because the questions differ. The comment above the
`def halt_gate:` block says so at the point of divergence.

## Two traps a naive fix walks into

1. **Blanket-skipping `auto_push=false` freezes recovery.** It is a standing convention on
   whole rigs. Skip only on `finished`; on `publish_blocked` alone, recover *without*
   publishing — and **keep the worktree**, because the push is what makes a worktree
   disposable and with the push withheld the worktree is the only copy of the work.
   Deleting it there is a new data-loss path in a fix meant to close one.
2. **Filtering a `--limit=1` refinery query starves the queue.** `gc bd list` has no
   metadata-key negation — `--has-metadata-key` and `--metadata-field` are both
   positive-only (verified against the live CLI), so the gate must be a client-side
   filter. Applied to a one-row result, a single held bead at the head masks every
   mergeable bead behind it: the refinery reports `IDLE` while real work waits, and
   because the hold is deliberate it waits forever. Fetch a page, filter, take the first
   survivor. Both arms are locked in by the fixture.

The gate **fails open**: an unreadable bead classifies as neither finished nor blocked and
behaves exactly as it does today. Failing closed would let one malformed bead freeze
orphan recovery for a whole rig.

## Verification

```
bash gastown/tests/test_halt_rail_gate.sh
```

Ten cases, all green, and the file is picked up by the existing
`for test_script in gastown/tests/test_*.sh` step in `ci.yml` — no workflow change.

**It runs the shipped text.** Each case pulls the bash recipe out of the formula's own
step description, through the TOML parser, and `eval`s it verbatim against a stubbed
`gc`. Never a retyped copy: retyping is how a tested predicate drifts from the prose an
agent actually reads at pour time, and that drift *is* the defect being fixed here. Going
through the parser also means every assertion is made against the post-escape bytes, so a
consumed backslash shows up as a failure rather than as a silent behaviour change.

**Both arms.** Halted beads are skipped *and* named; ordinary orphans still salvage;
`branch_ready=false` still means *not halted*; a pre-emptive `auto_push` hold is
recoverable but publish-blocked; an unmarked queue is untouched; a halted bead at the head
does not starve the queue; an empty queue still reads `IDLE`; an unreadable bead still
recovers. The fixtures are bead shapes observed in a live city, not invented ones.

**Drift guard.** The `def halt_gate:` block is embedded in both steps and the test
compares the two whole blocks on every run — never `grep -F`, which treats a multi-line
fixed pattern as one pattern per line ORed together and so passes whenever any single line
matches, which is always.

Mutation-tested — all killed:

| mutation | caught by |
|---|---|
| one-space edit inside either embed | drift guard |
| `branch_ready` matched by `has()` alone | `branch_ready=false` flips to halted |
| `halt_reason` matched by equality instead of presence | the empty and free-form values misclassify |
| `auto_push` dropped from `publish_blocked` | the pre-emptive hold becomes publishable |
| metadata-as-JSON-string normalisation removed | the rig-variance fixture misclassifies |
| gate fails closed instead of open | unreadable bead reads as halted |
| refinery takes the queue head without filtering; filter removed | an all-halted queue yields a held bead |
| refinery skip-report silenced | skipped beads go unnamed |
| refinery `--limit` reverted to 1 | patched-query assertion |
| witness halt-skip report line removed | patched-step assertion |

Two of those mutations found killing defects **in the test itself**, both by running the
mutations rather than by reading: the `grep -F` drift guard above, and a query assertion
that looked for `limit=25` anywhere in the step — which the paragraph *explaining* the
limit satisfies on its own, so reverting the real query to `--limit=1` passed. Both assert
against the executable lines now.

## What is in this PR

- `gastown/formulas/mol-witness-patrol.toml` — three hunks on `recover-orphaned-beads`:
  a Step 1b halt gate that runs before any salvage, a publish-authorisation rule stated
  once above the case list (with Case H recorded so the next reader does not reconstruct
  the salvage path for a halted bead), and two rows in the Step 4 notification table.
- `gastown/formulas/mol-refinery-patrol.toml` — one hunk on `find-work`: the client-side
  filter, `--limit=25`, and the skip report.
- `gastown/tests/test_halt_rail_gate.sh` — the fixture, wired into the existing
  `for test_script in gastown/tests/test_*.sh` CI loop. It extracts the recipes from the
  formulas' own step descriptions after TOML parsing, so it tests the shipped bytes.

Both patched steps are backslash-free on purpose, and the test asserts it: step text lives
in a TOML basic multi-line string where a legal escape is silently consumed (`printf
'%s\n'` arrives holding a real newline) and an illegal one — jq's backslash-parenthesis
interpolation, the form a formula author reaches for by reflex — stops the file parsing
and takes the whole pack down at pour time. Both shapes were live in the first draft.
Hence `.id + " (" + .gate.marker + ")"` rather than the interpolated form, and
`echo "$SKIPPED"` rather than `printf`.

One note for reviewers running a pinned pack: merging this does not land it. A pinned
snapshot keeps the bug until the pin moves, and any city-local override that shadows
these two formulas needs the same hunks applied to it — otherwise the merge and the pin
bump both land and the running agents still carry the defect.

## Separate finding, deliberately not patched here

`recover-orphaned-beads` Step 2 reads `metadata.worktree`, but every formula writes
`work_dir` (live: 3 in-progress beads carry `work_dir`, 0 carry `worktree`), and Step 1
and Case E of the same step text both say `work_dir`. As written, `WORKTREE` is always
empty and Cases B/C/D can never resolve one. The detmold witness nevertheless found
worktrees and pushed — so agents are routing around the broken recipe by judgment, which
means the audited text is not the executed behaviour. Different defect, adjacent lines, kept out of
this patch to keep it reviewable — flagged here because a reviewer will notice it, and
tracked separately on our side.

### DIFF
--- a/gastown/formulas/mol-refinery-patrol.toml
+++ b/gastown/formulas/mol-refinery-patrol.toml
@@ -169,11 +169,70 @@ needs = ["check-inbox"]
 description = """
 **Config: event_timeout = {{event_timeout}}**
 
-Search for work beads assigned to you with branch metadata:
+Search for work beads assigned to you with branch metadata, then drop any the
+PLAIN rail has marked no-publish:
+
 ```bash
-WORK=$(gc bd list ${GC_RIG:+--rig="$GC_RIG"} --assignee=$GC_AGENT --status=open \
-  --exclude-type=epic --has-metadata-key=branch --limit=1 --json | jq -r '.[0].id // empty')
+HALT_GATE='def halt_gate:
+  def truthy: if type == "boolean" then .
+              elif type == "string" then (ascii_downcase == "true")
+              else false end;
+  (if type == "array" then (.[0] // {}) elif type == "object" then . else {} end) as $b
+  | ($b.metadata) as $raw
+  | (if ($raw | type) == "string" then ($raw | fromjson? // {})
+     elif ($raw | type) == "object" then $raw
+     else {} end) as $m
+  | (if ($m | has("halt_reason")) then
+         ((($m.halt_reason // "") | tostring | .[0:60]) as $v
+          | if $v == "" then "halt_reason" else "halt_reason=" + $v end)
+     elif ($m | has("branch_ready")) and ($m.branch_ready | truthy) then "branch_ready"
+     else "" end) as $fin
+  | (($m | has("auto_push")) and (($m.auto_push | truthy) | not)) as $nopush
+  | {finished: ($fin != ""),
+     publish_blocked: (($fin != "") or $nopush),
+     marker: (if $fin != "" then $fin
+              elif $nopush then "auto_push=" + ($m.auto_push | tostring)
+              else "" end)};
+
+map({id: .id, gate: halt_gate})'
+
+CANDIDATES=$(gc bd list ${GC_RIG:+--rig="$GC_RIG"} --assignee=$GC_AGENT --status=open \
+  --exclude-type=epic --has-metadata-key=branch --limit=25 --json)
+
+GATED=$(printf '%s' "$CANDIDATES" | jq -c "$HALT_GATE")
+WORK=$(printf '%s' "$GATED" | jq -r 'map(select(.gate.publish_blocked | not)) | .[0].id // empty')
+SKIPPED=$(printf '%s' "$GATED" | jq -r 'map(select(.gate.publish_blocked)) | .[] | .id + " (" + .gate.marker + ")"')
 ```
+
+**A merge is a publish.** A bead carrying `halt_reason`, a truthy `branch_ready`, or
+a falsy `auto_push` is on the no-push rail: an operator or the mayor publishes it, on
+their schedule, or not at all. The refinery must not merge it whatever its branch
+state — not when the branch is clean, and not after a rejected attempt has been
+re-slung and re-halted. `metadata.branch` records where the work *is*; it is not
+consent to land it.
+
+**`--limit=25`, not `--limit=1`, is load-bearing.** The gate is a client-side filter
+because `gc bd list` has no metadata-key negation — `--has-metadata-key` and
+`--metadata-field` are both positive-only. Filtering a `--limit=1` result would make
+one halted bead at the head of the queue mask every mergeable bead behind it: the
+refinery would report `IDLE` while real work waited, and a permanently-held bead
+would starve the queue permanently. Fetch a page, filter, take the first survivor.
+Raise the limit if a rig ever holds more than ~25 halted beads at once; that is
+itself worth telling the mayor about.
+
+Report what you skipped — a silent skip is indistinguishable from an empty queue,
+and the operator holding those beads is the one person who needs to know the
+refinery saw them and left them alone:
+
+```bash
+if [ -n "$SKIPPED" ]; then
+  echo "HALT-SKIP (no-publish rail, not merged):"
+  echo "$SKIPPED"
+fi
+```
+
+Carry the skip list into `patrol-summary` for this cycle. Do not mail per bead: a
+held bead is a steady state, not an incident.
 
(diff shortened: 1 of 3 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277285824

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277329299

PR #366, opened 2026-08-28
fix(slack-full): atomic inbound reminders — head-protected budget, thread-reply anchor, deletion notice

citadel-mayor:
## What

Three delivery-mangling incidents in one week (papercuts `pc_a59c241284b8`, `pc_b334cff7f9c6`, and the 8/27 22:57Z escalation on gp-0qw) reached bound sessions as a bare reply-hint tail — no channel, no author, no message body. When a delivery is mangled downstream, the TAIL survives, so anything the adapter lets grow unbounded is paid for by the head: the part that identifies the message. gp-9gc pinned the shape: the once-per-channel reply how-to inflated deliveries and the cut kept the boilerplate over the body.

This PR implements the adapter-side atomicity contract from bead gp-0qw (folding in gp-9gc):

1. **Head-protected reminder budget** (`reminder_budget.go`, default 3500 chars of `Text`, `SLACK_REMINDER_TEXT_BUDGET` overrides, `0` disables). Shedding order when a delivery overflows:
   - the reply how-to detaches first — it re-arms (`replyHelp.unmark`) and rides a later, smaller delivery — gp-9gc's exact fix direction;
   - withheld peer context restores to its buffer and rides the next delivery;
   - the thread-context preamble (the one unbounded decoration) is replaced by an explicit omission notice;
   - last, the body is trimmed **tail-first behind a protected 200-rune head**, with an explicit marker naming the ts + channel so the full text is always recoverable — never a silent cut. The thread anchor and files block are never sacrificed;
   - budget `0` (every directly-constructed test config) reproduces the legacy layout **byte-for-byte** (pinned by test).
   - The contract applies on the urgent path **and on single-entry coalesced deliveries** (the message unit's parts ride the buffer and the crash spool); multi-message batches stay verbatim per the coalescer's digest contract ("nothing dropped or summarized") with only the how-to budget-gated — an overflowing batch is also the delivery shape the transport pastes atomically.
2. **Thread-reply anchor**: thread replies lead with `[thread reply — reply with --thread-ts <ts>; parent @author: "<first line>"]`. The anchor is captured once at intake, so buffered copies and spool replays carry the same parent line the immediate path renders (bare-ts fallback for pre-existing spool lines). The alias injection names the thread and now threads its reply how-to under the **root** ts (`--thread-ts` with a child reply's ts can misthread). The 8/26 incident hid a founder approval in an unplaceable thread reply for ~50 minutes — the anchor makes thread-ness explicit even when everything else about the delivery degrades.
3. **Deletion notice**: `message_deleted` events (previously dropped unread) rewrite still-buffered copies of the deleted ts to `[message deleted by sender before delivery]` (attachments dropped with the text), and a per-`(channel, ts)` deletion tombstone (15-min TTL) catches the copies a rewrite cannot see: a deletion event racing ahead of its message's enqueue, and a failed in-flight delivery restoring the original text afterwards. A dangling ts that resolves to nothing is exactly the pc_b334cff7f9c6 chase.
4. **Delivery honesty**: audited the twin-race dedup — channel claims / deliveredIDs already commit only after `postInbound` accepted a complete payload; that boundary is now documented at the commit site, and the success log gained `posted=NNch` (the actual POSTed payload length) so the next fragment incident is attributable adapter-vs-transport at a glance.

## Root cause of the fragment incidents (gc-side, follow-up needed)

The 8/27 22:57Z fragment was **not** the adapter's twin dedup: the winning twin POSTed the full 168-char body (success log) and gc accepted it. The mangling is on gc's last hop:

- `POST /extmsg/inbound` returns 200 **before** members are notified — the fan-out runs in `runBackground` (gascity `internal/api/huma_handlers_extmsg.go`), and per-member nudge errors are swallowed (`handler_extmsg.go` `notifyResolved`);
- reminders ≤4096 bytes are typed into the session TUI as a **raw `send-keys` keystroke burst**; only >4096 gets atomic bracketed paste (`internal/runtime/tmux/tmux.go` `sendLiteralText`). A busy TUI can lose the head of a multiline burst mid-word and submit only the tail — matching all three incident shapes.

The durable transport fix (always bracketed-paste for nudges + delivery receipts the adapter can gate its dedup on) belongs in the gascity repo; this PR bounds the adapter's exposure and makes every degraded delivery explicit and recoverable.

## Codex review

`codex exec --sandbox read-only`, five rounds. Rounds 1–4 (5×P1; 4×P1 + 2×P2; 4×P1; 1×P1) were all applied with fail-before/pass-after tests: single-entry buffered deliveries share the immediate path's budget contract; the preamble — and its omission notice, on files-only units too — sheds before the bound breaks; deletion tombstones cover the enqueue race, post-close spills and restores, failed-in-flight restores, mid-isolation deletions, dead-lettering, and the shutdown-drain spool producers; deletions are additionally **persisted as spool records** and applied on replay regardless of line order, so neither the check-to-durable-write window nor a restart resurrects deleted text; the intake-captured parent anchor survives buffering and every spool producer; the spool stores the message-unit parts without the redundant folded text (rebuilt on read, so lines shrink instead of doubling against the replay cap); the alias reply how-to threads under the root; reaction notifications never get a reply anchor; tombstone TTL reclaims across channels. The one accepted residual: the operator-only dead-letter file (never auto-delivered) can capture text deleted in the microsecond between `charge()`'s tombstone check and its write. Round 5 verdict is recorded in the bead comment.

## Testing

- `go test ./...` and `go test -race ./...` in `slack-full/adapter`: pass.
- `python -m pytest slack-full/tests -q`: 480 passed.
- New unit coverage: budget math (byte/rune safety, trim marker, protected head floor, legacy-bytes pinning), anchor formatting (neutralization, clipping).
- New integration coverage at the `processSlackEvent` seam: deletion notice for buffered copies (+ no-op for unknown ts, nil-coalescer safety), help-block withheld-then-re-arms on both paths, body trim with head preserved, thread anchor on urgent/DM and buffered single replies.

Bead: gp-0qw (P1, folds gp-9gc). Related: gp-151, gp-92h (not addressed here).


julianknutsen:
Please rebase against origin/main

PR #367, opened 2026-08-28
fix(slack-full): same-ts twin claims conclude on a delivery receipt, not on gc's 2xx (gp-32q)

citadel-mayor:
## The bug

A Slack bot-mention arrives as a **twin pair** — a `message` event and an `app_mention` event with the same
channel+ts but distinct event ids, so the event-id dedup cannot collapse them. `slack-full` serializes the
channel-audience delivery per `(channel, ts)` with a two-state claim (gp-ios): the first twin owns the delivery,
the other parks and, once the owner concludes, **skips its copy**.

Until this PR the owner concluded that claim on *"gc's `/extmsg/inbound` returned 2xx"*. That is not evidence the
session received anything: gc's inbound handler ran the member notification in the **background** and answered
`200` before the session was touched.

Twice in 24 hours a founder message was posted whole, gc answered 200, and the bound session received only the
reply-boilerplate tail:

```
8/27 22:57:10 CT  inbound: chan=C0AMHQ74ZLK ts=1787871429.928809 parked behind in-flight same-ts twin delivery
                  inbound: chan=C0AMHQ74ZLK ts=1787871429.928809 channel copy already delivered by same-ts twin — skipping channel post
8/28 04:08:01Z    identical pair for C0ASAPRETDK ts=1787890080.963799 (27ch text + 1 file, posted=867ch)
```

The one copy that could have recovered the message is exactly the copy that got suppressed. Adapter build was
`b51fc350` — PR #23's delivery-honesty audit is live and did not close this, because #23 tightened what the
adapter *posts*, and the loss is on the hop *after* the post. The mangling itself is gp-2rq (gascity core).

**A third reproduction landed at 05:39Z on 8/28, while this was being written, and changed the requirement.** A
223-byte founder message reached the mayor session as 134 bytes: **89 bytes eaten off the HEAD, tail byte-exact,
cut mid-word.** It read as a complete sentence starting mid-thought. Had the cut landed on a word boundary it
would have been acted on as a corrupted founder instruction and never noticed. So the property worth gating on is
not "it arrived" but **"all of it arrived"** — the receipt has to attest COMPLETENESS. That incident is a named
test vector in `delivery_receipt_test.go`.

## The fix

The claim concludes on a **delivery receipt**: gc vouching that the complete payload reached the session. Applied
to all three claim-holding legs — the urgent channel copy, the coalesced batch, and the address-by-handle session
injection. The wire contract was negotiated with the gp-2rq author and confirmed with three counters before
implementation.

| receipt | verdict | action |
|---|---|---|
| `delivered`, and no count — summary **or per member** — short of what gc built | **vouched** | conclude the claim; a twin may skip |
| `no_route` | **no route** | conclude — a twin's re-post resolves the same empty route |
| `failed` / `partial`, a member gc names as undelivered, or a `delivered` whose bytes fall short | **unconfirmed** | re-post ONCE in place (claim still held, so no twin can duplicate the attempt), then release the claim so the parked twin re-posts |
| `pending` | **held** | conclude the claim, re-post **nothing** |
| no `delivery` block at all, unknown status, unusable counts | **unsupported** | keep the exact pre-gp-32q commit-on-2xx behavior |

**Why `pending` holds rather than retries** (binding mayor ruling, 2026-08-28): gc waits for a session to reach an
idle boundary before pasting — `NudgeIdleTimeout`, 30s live on this city — which is longer than any HTTP budget
the adapter can hold. So `pending` is the *normal* receipt for a **busy** session, and busy sessions are exactly
the population this bug affects: the truncation only manifests on a busy TUI. Re-posting there would fire at
precisely the messages being fixed and deliver both copies, since gc never cancelled the first send — it just
could not report on it yet. `failed`/`partial` mean gc *knows* something did not land, which is what makes a
re-post clean. A hold concludes the claim (which also stops the parked twin from re-posting on its own — the
second duplicate path) and logs a `HELD` line carrying the receipt id.

**Why completeness is checked per member**: the summary counts are SUMS across the notify fan-out — one inbound
notifies every transcript member with a *different* per-member reminder — so a member that got half its payload
can hide inside a total that adds up. There is a test that constructs exactly that.

**Fail open, deliberately.** Anything the adapter does not fully understand degrades to `receipt=unsupported`,
which is byte-identical to the pre-gp-32q path: no receipt block, a status outside the confirmed enum, a count
that is not a whole non-negative number, a mixed-unit pair, two keys normalizing to one name. Only statements gc
actually makes may withhold a claim's conclusion — reading schema drift as "not delivered" would re-post every
delivery in the workspace, an outage strictly worse than the mangling this gate catches.

**Shutdown drain.** No re-post while draining. The failure path now **spools before it releases anything** and
concludes on whether the spool actually took the copy: a confirmed spill commits the claim (a takeover would post
a second copy of a message the restart already replays, and its POST can outlive the spool seal); a refused spill
releases it, because then the twin is the last recovery path. The event id follows the same evidence — except
when an addressed injection is still owed, since a spooled line replays as a channel copy and does not
reconstruct an injection. A twin that skips on a drain conclusion also takes no busy mark: the hourglass is
cleared by a reply to a delivery that is no longer happening, and the registry that could remove it does not
survive the restart.

## Deployment: safe to pin BEFORE the gc fix

Every gc shipped so far — including the binary live right now — emits no receipt. That is detected as
`receipt=unsupported` and keeps behavior **byte-identical to today**, with a test pinning it. The gate arms itself
the moment a receipt-emitting gc (gp-2rq) is running. No flag day, no ordering constraint between the two fixes.

## Observability

```
inbound: chan=C1 user=U_ALICE ts=… posted=867ch receipt=rcpt_0539 receipt_status=delivered receipt_verdict=unconfirmed receipt_bytes=134/223 receipt_digest=d6d50cab6a19385b receipt_member_undelivered=gc__mayor/delivered receipt_member_bytes=134/223
inbound: UNDELIVERED chan=C1 ts=… gc accepted the payload but did not vouch that it reached the session after 1 re-post(s) — …
inbound: HELD chan=C1 ts=… gc accepted the payload and has not finished delivering it — not re-posting (a busy session's normal path) — …
```

The digest is the one gc prints on its own `nudge-receipt` line, so a single delivery can be followed across both
logs during an incident. Receipt strings come from another process and ultimately from a terminal, so everything
interpolated into a log line is truncated, control- and whitespace-stripped, and capped in member count.

`SLACK_DELIVERY_RECEIPT_GATE=off` disarms the gate wholesale — the escape hatch if a gc build ever reports
receipts wrongly and the adapter starts re-posting deliveries that did land.

## Also fixed on the way (found by review, pre-existing)

- **The thread-context watermark was advanced before the forward and never put back on failure**, so every
  recovery copy — twin takeover, Slack redelivery, spool replay — shipped without the preamble the failed
  attempt had already "spent". Now rolled back conditionally, on both the channel and injection legs.
- **`errDeliveryUnvouched` reached the coalescer's isolation loop as a transient fault**, where restore-uncharged
  would have re-notified the channel every window forever. It is chargeable there too: bounded ladder, then the
  dead-letter file.
- **The claim was released before the failure cleanup settled**, so a woken twin could mark busy before the old
  owner's cancel ran and have its fresh mark deleted, stranding an hourglass.

## Tests

`delivery_receipt_test.go` — 34 tests. Headline: the bead's repro, two near-simultaneous deliveries of one ts
racing the urgent path while gc accepts every POST and vouches for none, asserting owner post + in-place re-post
+ **parked-twin takeover**, each carrying the complete payload; and the 05:39Z reproduction, asserting that an
arrival-only receipt verdicts as vouched (the shape that must not be enough) while the same receipt carrying gc's
byte counts convicts.

Every behavioral fix has a negative control — the guard was deleted and the test observed to fail:

```
TestDeliveryReceipt_DrainingTwinDefersToSpool        gc saw 2 POSTs, want 1 / spool carries the message 2 time(s), want exactly 1
TestDeliveryReceipt_UnvouchedSingleInIsolationIsCharged   dead-lettered [], want both singles after 3 charged attempts
TestDeliveryReceipt_DrainDeferredTwinLeavesNoBusyMark     a busy reaction was added on a drain-deferred delivery nothing can clear
```

Suites: `go test ./...` ok 50.175s, `go test -race ./...` ok 53.374s, `pytest slack-full/tests` 480 passed.

## Review

Five codex rounds; 22 findings, 5 of them P1. All accepted except three, each declined with reasoning recorded
in the code:

- **`pending` must not be re-posted through the aggregate shortfall** (final-round P1): a pending member's bytes
  are not in the delivered sum yet, so the summary necessarily falls short while it waits — reading that as
  truncation would re-post exactly the busy-session case the ruling protects. Held now outranks the aggregate
  check, while a member gc itself reports short still outranks held.
- **Declined — the coalescer's in-flight-claim window.** A buffered copy whose `(channel, ts)` claim is held by
  an in-flight urgent twin is kept and posted rather than parked. That is deliberate pre-existing design
  (pc_c920ff5fe90c): dropping it would leave no redelivery if the urgent owner then failed, since Slack already
  got its 200. Untouched by this PR, and it fails open to a duplicate, never to loss.
- **Declined — a channel-skipping twin's alias injection carries no thread preamble.** Pre-existing on `main`
  (the `skipChannelPost` guard on the thread-context block); this PR adds one more route to it. Closing it means
  resolving the alias claim before that block, reordering two claim lifecycles — its own change, not a rider
  here. Documented in place; the failure mode is a thinner injection, never a missing or truncated one.
- **Declined, but the comment corrected — the drain check is a check, not a barrier.** A re-post whose check
  wins the race with the drain flip by a scheduler hop can reach the spool after the seal and log `LOSS`. The
  earlier comment claimed this was fully bounded; it is not, for the urgent and injection legs, and it now says
  so. Closing it needs a synchronization primitive spanning the drain flip, three call sites and the seal.

Bead: gp-32q. Papercut: pc_2e2378b9918e. Core half: gp-2rq.


julianknutsen:
Please rebase against origin/main

PR #372, opened 2026-08-29
fix(slack-full): gc slack status is no longer load-sensitive — timeout+retry, daemon-busy verdict, JSON contract (gp-aut)

citadel-mayor:
## Summary

`gc slack status` was a false-alarm machine under daemon load (gp-aut; papercuts pc_915b4a8ac8b1 / pc_076c045ee7e3, citadel 2026-08-29 00:20–00:57Z with 5 worker sessions spawning): it either exited 1 with a bare Python `TimeoutError` traceback from `http/client.py readline`, or printed nothing under `--json`, while real reads/sends completed in <1s. The mayor had to tear down two monitors built on it.

**Root cause.** `urllib` only wraps *connect-phase* timeouts in `URLError`. A timeout while *reading the response* (`h.getresponse()` → `_read_status()` → `readline`, or `resp.read()`) escapes as a raw `TimeoutError`, so `slack_intake_common._request`'s `HTTPError`/`URLError` handlers never saw it and nothing upstream caught it. Reproduced against a deliberately stalling local HTTP server before touching code:

```
UNHANDLED escape: TimeoutError: TimeoutError('timed out')
readline in tb: True | http/client in tb: True
```

## Changes

1. **`slack_intake_common.py`** — new `GCAPITimeout(GCAPIError)` carrying `.timeout`; `_request` maps connect-phase (`URLError` whose `.reason` is a `TimeoutError`), read-phase (bare `TimeoutError`) **and error-body** (`HTTPError.read()` inside the except clause) timeouts to it. Existing `except GCAPIError` handlers in every other script keep catching it — no behaviour change for publish/bind/upload beyond "no more raw tracebacks". Two pre-existing degrade sites (`/sessions` alias lookup, transcript lookup) now treat a timeout like any other transport failure instead of crashing; neither produces a verdict, so that is the intended policy.
2. **`slack_chat_status.py`** — `--timeout` flag (default `$GC_SLACK_STATUS_TIMEOUT`, else **60s**, up from the 30s `_request` default; `nan`/`inf`/≤0 rejected on the flag, fall back on the env var). It is a per-request socket **inactivity** timeout — an attempt fails once the daemon has been silent that long; a response that keeps trickling bytes is not cut off. Each GET is retried **once** on `GCAPITimeout`; a second stall aborts the probe and exits **2** with a distinct `gc slack status: daemon busy (timeout after Ns) …` line on stderr — under `--json` it also prints `{"schema_version":"1","ok":false,"error":{"code":"daemon_busy","message":…,"exit_code":2}}` on stdout. The per-fetch `except GCAPIError: return []` degrade paths explicitly re-raise `GCAPITimeout` so a busy daemon can never render as "(none registered) / inbound: 0". A silent daemon costs at most 2 × timeout (the first fetch's two attempts); no retry cascade. Success `--json` output gains the `schema_version`/`ok: true` envelope (additive — all existing keys unchanged).
3. **`slack_chat_retry_peer_fanout.py`** — same guard: a daemon timeout while listing failed/retried fan-outs exits 2 with a daemon-busy line instead of a clean `candidates: 0` (which the new mapping would otherwise have produced — Codex round-1 finding).
4. **`commands/status/schemas/{result,failure}.schema.json`** — gc treats a command-local `schemas/result.schema.json` as the JSON-support declaration (`cmd/gc/json_schema.go` `readCommandSchema`), which removes the `does not declare JSON support; allowing --json pass-through` warning and makes gc buffer stdout and emit a contract failure object when the command fails with empty stdout.
5. **Tests** (`test_slack_chat_status.py` +16, `test_slack_chat_retry_peer_fanout.py` +1) — a real `ThreadingHTTPServer` that stalls its first N connections (counted and decided in the accept path, so exact-count assertions are immune to handler-thread lag) drives the actual socket timeout paths end to end, in both the status-line phase (the citadel shape) and the body phase for 200 and 5xx: `GCAPITimeout` mapping, daemon-busy line + exit 2 (no traceback), JSON failure object, retry-once-then-succeed (4 daemon hits for 3 fetches), no retry storm (exactly 2 hits when persistently busy), env/flag precedence and non-finite handling, not-swallowed-by-fallbacks, envelope, and `jsonschema` validation of both real outputs against the shipped schemas.
6. `help.md` + `CHANGELOG.md`.

## Verification

- `uv run --with pytest --with PyYAML --with jsonschema python -m pytest slack-full/tests -q` → **502 passed** (32/32 in the status file, 7/7 peer-fanout; every new test was watched failing first — including the round-1 Codex cases, which reproduced as raw `TimeoutError: timed out` / exit 0 before the fix — then green).
- 5× soak of the socket-level tests after the accept-path rework: 9 passed each run (~6s), no flakes.
- Branch is rebased onto `origin/integration/gp-cng-fda15ac7` (`137c10f`, tree byte-identical to `main`/`fda15ac`) at the gp-cng worker's request; `git diff origin/main...HEAD` is exactly this change.
- **Real `gc` binary, end to end**, from a throwaway city whose `pack.toml` imports this worktree's `slack-full` by local path, against a stalling local daemon on `GC_API_BASE_URL`:

```
$ gc slack status --json-schema
  json_supported = True | schemas: ['failure', 'result']
$ gc slack status --json --timeout 0.5
  exit=2
  stdout={"error":{"code":"daemon_busy","exit_code":2,"message":"daemon busy (timeout after 0.5s)"},"ok":false,"schema_version":"1"}
  stderr=gc slack status: daemon busy (timeout after 0.5s); the gc daemon is serving other work — retry shortly or raise --timeout / GC_SLACK_STATUS_TIMEOUT
  daemon hits = 2          (one retry, then abort — no pass-through warning)
$ gc slack status --timeout 0.5
  exit=2, daemon hits = 2
```

(Before the change the same probe reported `json_supported: false` and the live city still does — it runs the pinned SHA.)

## Operator notes

- Exit **2** = "gc daemon busy", not a Slack outage. Monitors should treat it as "retry later"; `--timeout` / `GC_SLACK_STATUS_TIMEOUT` raise the per-request budget.
- After the pin bump: the usual slack-full ritual (rebuild/copy the adapter shim + `.real` into the new cache, `gc service restart slack`) — this PR touches no adapter code, so copying is safe.

## Codex review

`codex exec --sandbox read-only` on the diff, two rounds.

**Round 1** (on the initial commit) — 4 findings + 1 nit, all addressed in the second commit:
- **High** — `HTTPError.read()` inside `_request`'s `except` clause could still leak a raw `TimeoutError` when gc sent a 4xx/5xx status+headers and then stalled on the *error body*. Fixed (wrapped; `TimeoutError` → `GCAPITimeout`, other `OSError` → `GCAPIError`) and covered by new body-phase socket tests for 200 and 5xx.
- **Medium** — `slack_chat_retry_peer_fanout._events` swallowed the new `GCAPITimeout` as `[]`, turning a previously-visible read timeout into a false `candidates: 0` / exit 0. Fixed: re-raise + daemon-busy exit 2 before any POST; tested for both listings.
- **Medium** — `--timeout` is an inactivity timeout, not a wall-clock bound; docs overclaimed. Fixed in docstring/help/CHANGELOG/this body.
- **Low** — `nan`/`inf` passed the `> 0` check. Fixed with `math.isfinite`; parametrized tests.
- Nit — request counting in the handler thread could flake exact-count assertions under thread starvation. Fixed: counted and decided in `verify_request` (accept path).

**Round 2** (on the fixed tree) — *"No HIGH, MEDIUM, or LOW findings. Nothing blocks this PR."* Re-verified all five points (including that `verify_request` runs before `ThreadingMixIn` spawns the handler and that the flushed headers land the client timeout in the body read), and agreed with leaving the two best-effort `GCAPIError` degrade sites in `slack_intake_common` as-is. Two nits (a docstring overstating "accepted the connection"; the peer-fanout test only stalling the first listing) are addressed in the final commit.


julianknutsen:
Please rebase this against origin/main and verify the changed files. This has a lot of mistaken deletes as well as infra changes to our runners that should not merge.

PR #373, opened 2026-08-29
fix(slack-full): follow up every HELD delivery with gc — recover or dead-letter a lost background send, never a silent hold (gp-3yg)

aphexcx:
## Summary

A delivery the receipt gate HELD (gc answered `pending` for a busy session, gp-32q) is now **followed up** with gc afterwards, and a definite loss is recovered or dead-lettered — never a silent `HELD` terminal state. Bead **gp-3yg**, P0 per Afik (2026-08-29 01:39Z: any Slack problem is P0; fallbacks are degraded — fix and retire them).

### Why

The pending=HOLD ruling is correct and stays: gc waits for a busy session to reach an idle boundary before pasting (NudgeIdleTimeout 30s), longer than any HTTP budget, so `pending` is a busy session's normal receipt and a re-post there duplicates the very messages being protected. But a hold also **concluded the claim**, and that was the end of the story: if gc's background send never landed, the adapter held no second copy, no alarm, no retry. Live evidence (2026-08-28):

1. **Silent loss** — Taylor's C0B0Y964Q1Z message (ts 1787901226) was never delivered to any session, produced no dead letter, and the watermark advanced past it. Found 19 hours later when Afik bumped the thread.
2. **False-negative dead letter** — C0BEZ3CQK5X ts 1787905440 was dead-lettered at 08:24:40Z with "no delivery receipt" (gc receipt `failed 0/1215`, member error *"submit Enter delivered to tmux but not confirmed (busy state never observed)"*) while the session had **answered it at 08:24:38Z**. The adapter re-posted it six times first.

### What changed (adapter)

- New `receipt_follow_up.go`. After a HELD verdict on any of the three claim-holding legs (urgent channel copy, coalesced batch, address-by-handle injection) the adapter polls `GET /v0/city/{city}/extmsg/inbound/receipts/{receipt_id}` (5s cadence, 10 min horizon):
  - `concluded` + vouched/no_route → `RESOLVED` (the hold was right).
  - `concluded` with every member either delivered whole or pasted whole with only the submit unconfirmed (gascity **gp-2io** classifies that member `pending` WITH its real byte counts — the busy-probe shape of the 08:24Z incident) → `RESOLVED` on that per-member evidence when the summary is `pending` or `partial`: a mixed room folds delivered+pending to `partial`, which the verdict would read as "a retry is clean", and it is not — the payload is in every pane. The same classifier runs on the synchronous receipt of a recovery re-post, so a recovery copy that lands whole into a mixed room is not falsely dead-lettered. Never re-posted, never dead-lettered (a "verify by hand" record for every fast turn the probe missed would be the incident's false-alarm noise, reborn). A concluded-but-pending receipt **without** landing evidence still waits out the deadline and is recorded as unknown-state.
  - `concluded` + failed/partial/short, or `unknown` (gc restarted; nobody is still trying) → `LOST`: **one** re-post of the identical envelope (dedup-keyed — gc dedups the transcript by provider message id) → `RECOVERED` **only when vouched**. A recovery copy that is itself held is followed once more; one that reaches nobody, is accepted without a receipt, or is not vouched for leaves the loss **standing**: dead-letter file + `LOSS` log. A dead letter therefore still means "the agent did not get this".
  - Still `pending` at the horizon or at shutdown (one final poll after the drain flips) → **not** re-posted (a slow send is not a lost one) but recorded in the dead-letter file with an explicit "state unknown, verify by hand" reason and a `STUCK` line.
  - Non-200 / unparseable / mismatched receipt id → treated as **transient until the deadline** (gc answers 404 for the route while a city's server is being replaced and during a restart); a window with no usable answer at all → `UNVERIFIED`, fail open to the plain hold; a hold gc once answered `pending` for and then went silent is recorded as `STUCK` instead (the endpoint exists; the silence may be a restart that lost the fan-out). A gc that predates the endpoint answers 404 for good, so this pack is **safe to pin ahead of the gc rebuild** and arms itself once one is running. `unknown` is a 200 state on purpose so it cannot be confused with a missing route.
- Shutdown: `receiptFollowUps.drain` runs after the coalescer drain and before the spool seal: admission closes, sleeping loops wake, in-flight polls are cancelled, a hold noted afterwards is recorded synchronously (LOSS line emitted before the write) and joined by the drain through a mutex+cond in-flight counter (no `WaitGroup` Add/Wait race); bounded at 25s (one recovery re-post already in flight at the 20s client deadline plus its dead-letter write). Every poll answer must echo the requested `receipt_id` (top-level and, when concluded, in the delivery block) or it is treated as unavailable.
- A lost address-by-handle injection is dead-lettered with `alias_session_id`, `alias_handle` and the exact rendered `alias_body` (the alias leg now renders once and re-sends those bytes on every attempt), so a hand recovery re-injects the right bytes into the right session.
- Docs: CHANGELOG, README env table, SETUP troubleshooting (`receipt follow-up:` markers, alias records).

### Companion gc changes (required for the endpoint; adapter is safe without them)

- aphexcx/gascity **PR #3** (branch `fix/gp-3yg-receipt-status`, base `integration`): records every inbound fan-out's conclusion by receipt id (`extmsg.InboundReceiptStore`) and adds the GET endpoint. The submit-unconfirmed reclassification this PR originally carried was **carved out** — sibling P0 bead **gp-2io** owns it with mayor-ruled semantics (member `pending` WITH real byte counts, typed delivery evidence), and this adapter's landed-unconfirmed arm is designed against exactly that shape.
- Against a gc that predates gp-2io (still concludes the busy-probe shape as `failed 0/N`), the follow-up re-posts once (dedup-keyed) — the old incident behavior, gone once the gc roll lands. Against a gc that predates the endpoint entirely, the follow-up fails open (`UNVERIFIED`), so the pin order is free.

### Interpretation of the acceptance text

"definite loss → dedup-keyed re-post + dead letter + loud log": implemented as re-post first, dead letter **only when the loss stands** (recovery copy unvouched / not attempted / exhausted). A dead letter for a message the agent then received would be exactly the false positive incident 2 produced. Easy to flip to "always record" if the mayor prefers.

### Verification

- `cd slack-full/adapter && go vet ./...` — clean; `go test -count=1 ./...` — ok (49.1s); `go test -race -count=1 ./...` — ok (54.2s)
- `uv run --with pytest --with PyYAML --with jsonschema python -m pytest slack-full/tests -q` — 480 passed
- New tests: 38 component tests (pending→landed-late, pending→lost→recovered, lost→unvouched→dead letter, held re-post ladder bound, unknown, unavailable fail-open, transport-error retry, stuck past deadline, drain final poll, loss during drain, duplicate note, nil no-op, wire parsing, agreed path, **concluded landed-unconfirmed resolves quietly — original, recovery, and partial-summary mixed room; concluded-pending without landing evidence still records unknown-state; and the `landedUnconfirmed` table as the spec**) + 6 leg tests (urgent lost/late/standing-loss dead letter, coalesced lost, alias lost byte-identical across a channel-name cache change, alias standing loss record).

### Codex gate

Seven rounds of `codex exec --sandbox read-only` on the adapter diff, all findings applied except one declined with recorded reasoning:

- **r1**: P1 every poll answer must echo the asked receipt id (a stale/mis-routed answer is not an answer); P1 shutdown join of post-close records; P1 an unvouched recovery leaves the loss standing (dead letter); P2 tombstones for late duplicate notes; P2 gc strings sanitized/bounded in logs; P2 alias body rendered once and re-sent byte-identically.
- **r2**: P1 recovery-vs-original give-up table; P1 LOSS line emitted before the write on post-close records; P2 blank-id holds recorded, not dropped; P2 alias dead letter carries session/handle/body.
- **r3**: P1 `WaitGroup` Add/Wait race → mutex+cond in-flight counter.
- **r4**: P1 blank-id record joined by drain; P1 `sawUsable` — gc silence after a usable answer records STUCK instead of failing open; P3 post-close records tombstoned; **one P1 declined with recorded reasoning** (a note arriving after the drain's zero-test is joined by nobody — the straggler contract the adapter already accepts for spool writes past the seal, gp-9e7; the LOSS-grade line is emitted before the write so the alarm survives even when the record does not).
- **r5**: killed mid-run (worker session died), no verdict.
- **r6**: P2 `landedUnconfirmed` gated to `pending`/`partial` summaries + byte-unit evidence; P2 the same classifier on the synchronous recovery response; P2 LOSS-vs-STUCK doc semantics; P3 drain-timeout diagnostics count anonymous records.
- **r7**: confirmation — all four r6 fixes verified, no new findings; the sole remainder is the declined r4 P1.


julianknutsen:
Please rebase this against origin/main

PR #376, opened 2026-08-30
fix(gascity): pick the implementation review verdict by owner, not by bead id

superlzyguy:
`implementation-review-approved.sh` is the Ralph loop-exit gate consumed by 10 formulas across 5 packs. It could strand a finished build. Two independent defects were confirmed, either sufficient on its own.

## The defects

**1. The loop verdict was picked by position, not by owner.** `gmol()` ended in
`unique_by(.id) | sort_by(.updated_at // "")`, and the downstream `| last` consumed that row
order. That re-sort was **live**, and `| last` really did mean "most recently updated":
`updated_at` is `omitempty` on the reader's bead struct, absent only on a bead never updated
since it was created — which no verdict carrier can be, because it got its verdict key from
`gc bd update`. (Measured 56/56 rows carrying `updated_at` on this change's own review root.)

The defect is that "most recently updated" is not "the lane that owned the review". Selection
was positional and ranged over every verdict-bearing bead in the molecule, so the newest row
won even when it came from a bead that did not own the review; and on equal timestamps jq's
stable sort left bead-id order to break the tie.

**2. The file disagreed with itself about "approved".** Its own `approved()` helper
accepted `approve|approved|pass|done`, while the top-level dispatch accepted only `done`
and `approved|pass`. A lane spelling its verdict `approve` was rejected by the gate that
its own helper would have accepted.

## The repair

- Positional selection is replaced with **owner-preferring, permutation-invariant**
  selection: the verdict is chosen by the lane that owns the review, so exchanging bead
  ids in the fixture produces an identical exit code and stdout.
- Recency is read **by value where it is used**, not staged in row order: `newest()` narrows
  to the maximum `updated_at`, but only when *every* survivor is dated; a partially dated set
  is deliberately not ranked. `decide()` then reduces any remaining disagreement to the
  non-approving value, so the gate **fails closed** rather than approving on a tie.
- The approval vocabulary is defined **once, case-insensitively**, for both consumers.
- The lane-fallback path now **announces** which verdict it selected instead of deciding
  in silence, and shares `decide()` — so one bare-key approval no longer outvotes two
  `iterate`s at equal recency.

### One deliberate deviation from the filed suggestion

The source bead suggested filtering on `gc.step_id == "review.apply-review-findings"`.
That was **not** implemented: the literal string selects nothing for 7 of the 10 consuming
formulas and would have broken most consumers. The landed rule reads a contract every lane
prompt states verbatim instead.

## Evidence

| Check | Result |
| --- | --- |
| `pytest -k implementation_review -q`, test change only (red baseline) | 7 failed, 6 passed, 5 subtests |
| same, after the script repair | 10 passed, 89 deselected, 8 subtests |
| ids exchanged in the permutation fixture | identical exit code and stdout for both orderings |
| `python3 -m pytest gascity/tests -q` at the final head | 199 passed, 9461 subtests |
| `pytest -k implementation_review -q` at the final head | 14 passed, 185 deselected, 20 subtests |
| full CI matrix (`.github/workflows/ci.yml:44`, 11 suites), measured at `c690bad` | 1328 passed, 29 skipped, 9484 subtests |
| `grep -n 'sort_by(.updated_at' <repaired script>` | no match (exit 1) |

Two files change together: `gascity/assets/scripts/checks/implementation-review-approved.sh`
and `gascity/tests/test_formula_assets.py` (+809 / −60). Rollback is `git revert` of the
four commits in this branch (`ca5170d`, `8347a90`, `a463594`, `796d20c`) — reverting only the
first three would leave the selection rewrite in the tree without the fixture and comment work
it is coupled to.

## Notes for the merger

- **A parallel branch edits the same test file.** Unmerged
  `gc/ac-s362al-shared-namespace-owner-check` also touches
  `gascity/tests/test_formula_assets.py`. Near-identical sibling test blocks auto-merge with
  no conflict marker and nothing in Python catches it — whoever merges second should re-run
  `python3 -m pytest gascity/tests -q`. That branch's base also carries a pre-`56016d5` copy
  of the target script, so a merge taking that side would silently revert the
  federating-reader fix.
- **Out of scope, tracked separately — and the sibling sorts are live, not inert.** The same
  `sort_by` lives in `design-review-approved.sh:35` and `gap-analysis-approved.sh:33` (the
  latter accepts `done` only). Review measured both: deleting `gap-analysis-approved.sh`'s
  sort turns its guard red in **both** directions — against a molecule whose fresh verdict is
  `iterate` and whose id-last verdict is a stale `done`, the shipped script correctly returns
  `ITERATE(1)` and the sort-deleted script returns `APPROVE(0)`, a real fail-open.
  `design-review-approved.sh:35`'s copy is inert *at that line only*, rescued by the
  selection-site `sort_by(.attempt, .updated_at)` at `:77`; delete both and its guard goes red
  too. **The follow-up must port owner/value-based selection while keeping a recency
  tie-break — it must never merely delete the sorts.** Both siblings are verified untouched
  here and are filed as `mc-57k9j`.
- **No pack release step was performed.** Whether a version bump or registry stamp is also
  needed is a release decision outside this code change. Installed rigs keep running the
  broken gate until their next pack-cache refresh.

https://claude.ai/code/session_01BsWb1VXLXKTEPAaJKEF1dS

### DIFF
--- a/gascity/assets/scripts/checks/implementation-review-approved.sh
+++ b/gascity/assets/scripts/checks/implementation-review-approved.sh
@@ -103,7 +103,8 @@ MATCHES="$(gmol "$PARENT_ROOT")"
 # not to write the bare `code_review.verdict`, so a candidate carrying any
 # `code_review.*_verdict` key is a lane and is dropped -- unless dropping the
 # lanes would leave nothing, in which case every candidate is kept. Narrowing
-# may never starve the gate.
+# may never starve the gate. That fallback decides from lane beads alone, so it
+# announces itself on stderr whenever it reduces more than one candidate.
 #
 # The survivors are then reduced to one value by value, never by position: an
 # approval if any survivor carries one, otherwise the lexicographically
@@ -139,6 +140,8 @@ VERDICT_SELECTION="$(printf '%s\n' "$MATCHES" | jq -r \
   | (
       if ($owners | length) > 1 then
         "review check: \($owners | length) owner-shaped beads carry code_review.verdict at attempt \($attempt) (values: \($values | join(", "))); selected \"\($verdict)\""
+      elif ($owners | length) == 0 and ($surviving | length) > 1 then
+        "review check: no owner-shaped bead at attempt \($attempt); reduced \($surviving | length) lane candidates (values: \($values | join(", "))); selected \"\($verdict)\""
       else
         ""
       end
--- a/gascity/assets/scripts/checks/implementation-review-approved.sh
+++ b/gascity/assets/scripts/checks/implementation-review-approved.sh
@@ -26,11 +26,16 @@ gmol() {   # root_id -> molecule-member JSON array
         echo "gmol: gc ready failed for status: $(tr '\n' ' ' <"$tmp/failed")" >&2
         rc=1
     fi
-    # unique_by sorts by id, so the union comes back in bead-id order. The
-    # verdict extractors below take `| last`, which must mean "most recently
-    # updated" -- without this re-sort the gate picks a verdict by id hash and
-    # can sit on a stale `iterate` forever while a newer `done` is ignored.
-    jq -s 'map(select(type=="array")) | add // [] | unique_by(.id) | sort_by(.updated_at // "")' "$tmp"/*.json || rc=1
+    # unique_by sorts by id, so the union comes back in bead-id order -- and
+    # that is the only order there is. `gc ready --json` emits no `updated_at`,
+    # so the re-sort this line used to carry compared every row equal, left the
+    # id order untouched, and still claimed to mean "most recently updated".
+    #
+    # The verdict selection below does not depend on this order: it partitions
+    # the candidates by ownership and then reduces them to one value by value,
+    # never by position. The lane-status aggregation further down does still
+    # take the id-last value per key; see the note there.
+    jq -s 'map(select(type=="array")) | add // [] | unique_by(.id)' "$tmp"/*.json || rc=1
     rm -rf "$tmp"
     return "$rc"
 }
@@ -56,6 +61,26 @@ metadata_value() {
   ' 2>/dev/null
 }
 
+# The one approval vocabulary. Both consumers read this single definition and
+# both match case-insensitively: the jq lane-status helper, which receives it
+# via --argjson, and the bash dispatch at the bottom of the file. A spelling
+# added here reaches every consumer at once.
+APPROVAL_VERDICTS=(approve approved pass done)
+APPROVAL_VERDICTS_JSON="$(printf '%s\n' "${APPROVAL_VERDICTS[@]}" \
+  | jq -Rsc 'split("\n") | map(select(. != ""))')"
+
+is_approved() {
+  local candidate known
+  candidate="$(printf '%s' "${1-}" | tr '[:upper:]' '[:lower:]')"
+  [ -n "$candidate" ] || return 1
+  for known in "${APPROVAL_VERDICTS[@]}"; do
+    if [ "$candidate" = "$known" ]; then
+      return 0
+    fi
+  done
+  return 1
+}
+
 ROOT_JSON="$(gc bd show "$ROOT_ID" --json 2>/dev/null || true)"
 PARENT_ROOT="$(metadata_value "$ROOT_JSON" "gc.root_bead_id")"
 if [ -z "$PARENT_ROOT" ]; then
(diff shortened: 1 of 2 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277285675

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277285750

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277299508

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277306943

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277307851

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @superlzyguy! This PR repairs the Ralph loop-exit gate that ten
formulas across five packs depend on. `implementation-review-approved.sh` was choosing its
verdict by row position rather than by the lane that actually owned the review, and the file
disagreed with itself about the word "approved" — its own `approved()` helper accepted
`approve`, while the top-level dispatch did not, so a lane spelling its verdict that way was
silently rejected. Either defect on its own can strand a finished build in a retry loop, which
makes this a load-bearing fix well out of proportion to its diff size. The permutation-invariance
framing you brought to it — exchange the bead ids and the gate must produce identical output —
is what made the fix testable rather than anecdotal, and it is now pinned by the suite.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- **Verdict selection could approve on a stale verdict.** The first version preferred the
  owning lane but did not order what remained, so a stale `approved` could beat a fresher
  `iterate` — a case where the previous code had failed closed. Selection is now two
  non-positional, value-based steps: `newest()` narrows to the maximum `updated_at`, but only
  when *every* survivor is dated, and `decide()` then reduces any remaining disagreement to the
  non-approving value. Real disagreement at equal recency now yields `iterate`; a separating
  timestamp lets the newer value win. Both behaviours were driven against the gate's own jq
  program.
- **The `updated_at` premise behind the change was inverted.** The PR described `gc ready --json`
  as emitting no `updated_at`, which would have made the removed `sort_by` a no-op. The field is
  `omitempty` — absent only on a bead never updated since creation, which no verdict carrier can
  be — so the old sort was live and `| last` genuinely meant "most recently updated". Measured
  56/56 rows carrying the field, four independent times across two review passes. The script
  comment, the test docstrings, the fixtures that had lost the field, and the PR body now all
  state the true contract; the real defect is the positional, owner-blind selection, which the
  patch does fix.
- **The filed follow-up would have reintroduced the bug next door.** The original note called the
  sibling `sort_by` copies in `design-review-approved.sh` and `gap-analysis-approved.sh` "inert".
  Deleting `gap-analysis-approved.sh:33`'s sort is a measured fail-open — a stale `done` beats a
  fresh `iterate` — and `design-review-approved.sh`'s is inert only at that one line, rescued by
  the selection-site sort further down. You correctly left both scripts untouched; the fixup goes
  further and *pins* their behaviour with guards, so the misdirected edit now turns a test red
  instead of quietly opening a gate. This was the strongest outcome of the review.
- **Two "behaviour-preserving" comments described a change that did occur.** The report-path and
  lane-status sites shifted from recency-last to id-last. Both now state the actual change, its
  exact divergence condition, and that the residual is deliberately unrepaired.
- **The no-owner fallback was fail-open.** A single bare-key approval could outvote any number of
  `iterate`s at equal recency. `decide()` is now shared by the owner path and the fallback.
- **A mixed-case vocabulary entry was unmatchable.** Both consumers now downcase entry and
  candidate, pinned by a mutation test that copies the gate rather than editing it and carries a
  negative control.

Review findings addressed:
- All six of the above were resolved by maintainer fixup commit `796d20c`
  (`fix(gascity): decide the review verdict by recency, then fail closed`), which layers the
  recency-then-fail-closed selection onto your owner/lane partition, restores the fixture field,
  corrects the in-repo narrative, and adds the sibling guards.
- Test evidence was the strongest axis of the change: direction-complete assertions, id-order
  exchanges, mutation tests with negative controls and self-retargeting guards, and sibling
  guards that pin behaviour rather than text. Every mutation claim in the commit messages that
  review probed reproduced. The full suite runs 199 passed / 9461 subtests on the final head.

- `mc-57k9j` — Contract & Interface Fidelity / minor: port the owner/value-based selection into
  `gap-analysis-approved.sh` and `design-review-approved.sh`, keeping a recency tie-break. The
  bead spells out explicitly that the sorts must not simply be deleted.
- `mc-iebnc` — Contract & Interface Fidelity / minor: five sibling approval gates keep
  independent case-sensitive vocabularies missing values that lane prompts do emit; reconcile or
  document the narrowing.
- `mc-c1pa1` — Architectural Consistency / minor: verdict selection for retry-loop gates exists
  in three implementations with no cross-references, so a fix in one propagates to none. Also
  records that `updated_at` is compared lexicographically, which misorders when one
  fractional-second string is a strict prefix of another — real mechanism, effectively
  unreachable at the widths the reader emits, and not a regression.
- `mc-eefvo` — Debuggability & Operability / nit: the `$basis` audit label reports "fail-closed
  among N values" whenever more than one distinct value is present, but the gate only truly fails
  closed when a non-approving value exists, so an all-approval tie is mislabelled. Audit note
  only; no control flow reads it.
- `mc-74ckm` — Release Safety / nit: `${known,,}` is the first bash-4-only expansion in the
  shipped check-script corpus. Fail-closed and loud rather than dangerous, and no supported
  environment is known to be affected; the same file already solves it portably with `tr` two
  lines up.

### Maintainer Changes
Maintainer fixes were pushed directly to this PR branch; your three commits are preserved with
original authorship.

- `796d20c fix(gascity): decide the review verdict by recency, then fail closed` —
  2 files changed, 452 insertions(+), 75 deletions(-)
- The branch was rebased onto current `main` during finalization; the adopted patch is unchanged
  (patch-id preserved).
- The PR description was corrected in place: the `updated_at` narrative, the "inert sibling
  sorts" note, and the evidence table and rollback instruction, which had gone stale by the fixup
  (now +809 / −60 over four commits, 199 passed / 9461 subtests).

### Final Review Status
Ready for the merge queue: final head `796d20c` has passing required GitHub checks, and the
merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue
can pick it up.

One merge-order note for whoever lands it: the unmerged branch
`gc/ac-s362al-shared-namespace-owner-check` also edits `gascity/tests/test_formula_assets.py`,
and near-identical sibling test blocks auto-merge without a conflict marker. Whichever of the two
merges second should re-run `python3 -m pytest gascity/tests -q`.

CI: https://github.com/gastownhall/gascity-packs/actions/runs/34193283948

### Review Iterations
2 review passes. The first returned request-changes with four gating issues; the second confirmed
all four closed, plus both non-gating items, and approved.

---

PR #377, opened 2026-08-30
fix(gastown): make Dolt diagnostics portable on macOS

AlexBelous:
## Summary

- replace GNU `timeout` calls in the operational-awareness Dolt diagnostic bundle with the same Python 3 bounded-execution contract used by Gas City's Dolt runtime
- preserve normal exit codes, return 124 on timeout, and escalate SIGTERM to SIGKILL after a two-second grace period without requiring Homebrew/coreutils
- add asset assertions that prevent the non-portable dependency or hard-bound guarantees from regressing

## Verification

- `git diff --check`
- `gastown/tests/test_gastown_pack_assets.sh` (run with Python 3.12 because the macOS system Python is 3.9 and lacks `tomllib`)
- functional helper harness: a normal 30-second sleep returned 124 at 1 second; a SIGTERM-ignoring sleep returned 124 after the 2-second grace period; fast success returned 0; fast failure returned 1

This intentionally changes only the pack template. It does not alter any consuming city pin, install, or runtime.

### DIFF
--- a/gastown/template-fragments/operational-awareness.template.md
+++ b/gastown/template-fragments/operational-awareness.template.md
@@ -54,18 +54,50 @@ If you detect Dolt trouble (commands hang/timeout, "connection refused",
 are hard to reproduce. A blind restart destroys the evidence. Always:
 
 ```bash
+# Bound a command without depending on GNU coreutils `timeout`, which is not
+# installed by default on macOS. This mirrors Gas City's Dolt runtime helper:
+# preserve normal exit codes, return 124 at the deadline, and escalate from
+# SIGTERM to SIGKILL after a short grace period.
+run_bounded() {
+  bound_seconds="$1"
+  shift
+  if ! command -v python3 >/dev/null 2>&1; then
+    printf 'diagnostic: python3 not found; cannot run bounded command\n' >&2
+    return 124
+  fi
+  python3 - "$bound_seconds" "$@" <<'PY'
+import subprocess
+import sys
+
+limit = float(sys.argv[1])
+command = sys.argv[2:]
+process = subprocess.Popen(command)
+try:
+    process.wait(timeout=limit)
+except subprocess.TimeoutExpired:
+    process.terminate()
+    try:
+        process.wait(timeout=2)
+    except subprocess.TimeoutExpired:
+        process.kill()
+        process.wait()
+    sys.exit(124)
+sys.exit(process.returncode)
+PY
+}
+
 # Group all four captures under one timestamp so the bundle is easy
-# to attach to the escalation note. Each timed step writes via
-# redirect (not `tee`) so timeout's exit 124 propagates to `||` and
-# the agent gets an explicit "diagnostic timed out" signal — POSIX
-# pipelines mask the upstream exit code via tee.
+# to attach to the escalation note. Each bounded step writes via redirect
+# (not `tee`) so its non-zero status propagates to `||` and the agent gets an
+# explicit "diagnostic timed out" signal — POSIX pipelines mask the upstream
+# exit code via tee.
 ts=$(date +%s)
 
 # 1. Capture live process state via SQL (non-fatal — Dolt keeps running).
 #    SHOW FULL PROCESSLIST lists active connections, the query each is
 #    running, and time-in-state. Bound the call so a wedged server can't
 #    block the diagnostic itself.
-timeout 5 gc dolt sql -q "SHOW FULL PROCESSLIST" \
+run_bounded 5 gc dolt sql -q "SHOW FULL PROCESSLIST" \
     > /tmp/dolt-hang-$ts-procs.log 2>&1 \
   || echo "(step 1 timed out or failed — see procs.log for partial output)"
 cat /tmp/dolt-hang-$ts-procs.log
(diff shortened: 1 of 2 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277285529

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277329212

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @AlexBelous! This PR replaces the Dolt-hang runbook's
dependence on GNU coreutils `timeout` with a self-contained `run_bounded` helper, which
matters because stock macOS ships no `timeout` at all: steps 1, 3, and 4 of the runbook
previously died at exit 127 and captured **empty evidence logs** precisely when a wedged
Dolt server made that evidence most valuable. Bounding the diagnostics on bash, POSIX
signals, and python3 alone makes the procedure work identically on every host an operator
might be sitting at. The PR also arrived with anti-regression coverage for the GNU
spelling it removes — measured to trip exactly three times against the pre-PR file, one
per old call site, so the pin would have caught the very regression this PR fixes.

### Original PR Review
Decision: approve.

Three reviewers (design/implementation, implementation, and cross-file/test-risk) covered
the change and none filed a blocker or a major. The migration itself held up under
independent execution: all three lanes plus the synthesis ran the shipped helper and
confirmed its contract (exit-code passthrough, 124 at the deadline, SIGTERM then SIGKILL
after a 2s grace), verified that all three call sites kept their bounds (5/60/10),
capture paths, and `||` failure messages, and confirmed step 2 was correctly left
unbounded with its rationale intact. A repo-wide sweep found no other unguarded
`timeout N` in any pack, and the single-source runbook was confirmed to reach every
rendered prompt, so the fix is complete for its scope.

Specific gaps fixed:
- The new assertions were content pins only — nothing executed the helper. A controlled
  mutation proved the gap concretely: deleting the helper's `shift` line breaks every
  diagnostic call with a `FileNotFoundError`, yet left **all nine pins green**. The final
  patch adds an executable probe that extracts the fenced helper and runs it, and that
  same mutation now goes red.
- A copy-paste habit from the pack's sibling helper (`timeout 2s`) would hit
  `run_bounded 5s ...` and produce a raw `ValueError` traceback. The final patch fails
  closed with a usage diagnostic and exit 124 instead — the same fail-closed direction the
  helper already takes when python3 is missing.

Review findings addressed — all five were taken by commit `d7766b1`:
- **Test evidence quality** (minor, high confidence): the new pins could not see a python
  syntax error, a dropped `shift`, or a deleted exit path inside the helper. Now the suite
  executes the shipped helper and asserts exit passthrough (`7`), the deadline (`124`), the
  rejected bound, and signal death — with a recorded mutation matrix showing each arm goes
  red for the mutation it is meant to catch.
- **Duration argument** (minor, high confidence): `run_bounded 5s ...` now returns a usage
  diagnostic and 124 rather than a traceback. The maintainer fix chose the fail-closed
  guard over `rstrip("s")` deliberately — `rstrip` would half-support GNU syntax, making
  `5s` work while `1m` still tracebacks, which teaches a contract the helper does not honor.
- **Architectural consistency** (minor, high confidence): this is the third same-named
  `run_bounded` in the tree, and nothing recorded *why* this variant drops the `timeout`
  fast path its canonical sibling prefers — so a future harmonization edit would fight the
  negative pin without understanding it. The header comment now records the rationale
  (one deterministic code path for a paste-anywhere runbook) and points at the canonical
  helper in the dolt pack.
- **Exit-code fidelity** (nit, high confidence): a child killed by a signal reported 241
  rather than the shell convention 143, which quietly contradicted the header's promise to
  "preserve normal exit codes". Now normalized to 128+N, with a probe arm pinning it.
- **Stdin passthrough** (minor, high confidence): the helper's own script arrives on fd 0,
  so a bounded command does not inherit the caller's stdin. The finding offered either a
  code switch or a documented limitation; the maintainer fix took the documented route,
  because switching to `python3 -c` is a drop-in only while the block contains no
  apostrophe — an unguarded landmine in a runbook meant to be pasted and edited. The
  code-level half is sequenced with the canonical helper in the follow-up below.

One correction worth recording: the review stated that gc's canonical helper shares the
stdin limitation, unqualified. Measurement during the maintainer fix showed it shares it
only on the **python fallback** path — with a real `timeout` binary, stdin passes through
normally. The runbook comment now says exactly that.

- `mc-3py2v` (P2): sequence the stdin and bad-bound hardening into gc's canonical
  `doctor/check-dolt/run.sh`. Its python fallback still drops stdin and still parses the
  bound with an unguarded `float()`. Fixing only the gastown variant widens the very
  divergence the third-variant finding asked us to record, so the two should move together.
- `mc-m8pxn` (P3): two pre-existing edges in the runbook that this PR left untouched and
  that the review explicitly did not charge against it — `run_bounded` signals its direct
  child only, so a grandchild can outlive the bound; and the `/tmp/dolt-hang-$ts-*` capture
  paths are timestamp-only, so two concurrent operators can overwrite each other's evidence.

### Maintainer Changes
One maintainer commit pushed to the PR branch, `d7766b1` — "fix(gastown): execute the
run_bounded helper in tests; record its contract" — applying the five non-gating follow-ups
above. 2 files changed, 56 insertions(+), 3 deletions(-), staying entirely inside the two
files this PR already touches. Your commit `d84b2d2` is preserved unchanged with your
authorship.

### Final Review Status
Ready for the merge queue: final head `d7766b1` has passing required GitHub checks, and the
merge-ready metadata is in place. Locally, all five gastown shell suites pass and the
inference-gate suite (87 tests) is green on this head. I am marking this PR
`status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/gascity-packs/actions/runs/34520308436

### Review Iterations
1 review pass performed: three reviewers plus synthesis, resolved to approve with no
blocker or major, followed by a single maintainer fixup applying the non-gating polish
items and re-validated locally and in CI.

---

PR #389, opened 2026-09-01
fix(gastown): drop the stale project-level codex hooks overlay

Rome-1:
## What

Removes `gastown/overlay/per-provider/codex/.codex/hooks.json`, and the `embed_test.go` assertion that required it.

## Why

gc stages per-provider overlays into every agent's working directory. Codex **stacks** hooks — it runs `~/.codex/hooks.json` **plus every ancestor `.codex/hooks.json`**, rather than the nearest file overriding — so a codex agent running inside a gc tree fires each hook once from the user-level file and again from this staged copy. Measured on a live city: `UserPromptSubmit` firing 6 times where 3 is correct.

This particular copy is also the **oldest of three generations** of the same file that exist across the pack sources:

| variant | source | `--hook-format codex` on handoff | `gc hook run --timeout` wrapper |
|---|---|---|---|
| `bbd0d76e` | gastown pack (**this file**) | no | no |
| `46f61e93` | core pack | yes | no |
| `a3558786` | core pack | yes | yes |

The two features arrived in sequence, so an older copy winning a staging tick loses features in a known order — and the timeout guard is the last thing added and the first thing lost. Without it, `nudge drain` and `mail check` run with no ceiling on a hook that blocks prompt submission.

All three carry the **same number of commands across the same events**, so neither a command count nor a file-presence check distinguishes them. The differences are only visible in the command strings.

## Scope, and an ordering hazard for the follow-up

This PR removes only the stale gastown copy. The canonical core-pack copy lives in a **different repository** (`gastownhall/gascity`, `internal/bootstrap/packs/core/overlay/per-provider/codex/.codex/hooks.json`), so the two cannot land atomically.

**Order matters, in the direction that hurts.** Removing the canonical copy *first* makes every machine that carries both fall back to staging the stale one — delivering the exact regression described above as the result of the fix for it. **Stale first (this PR), canonical second.**

The canonical copy should **not** simply be deleted either; it is load-bearing for the explicit `install_agent_hooks = ["codex"]` opt-in and for the doctor's PreCompact drift check. That half is filed as an issue against `gastownhall/gascity` rather than a PR, because it requires a decision about gc's default overlay-staging behaviour that belongs to the maintainers.

## Test

`go test ./...` passes.

`embed_test.go` asserted this file's presence. That assertion is removed with the file — deliberately, and called out here rather than left silent: it is the correct kind of check, one that fails when its subject is removed, and it did its job.


Rome-1:
Cross-reference: the second half of this fix is filed as an issue rather than a PR — gastownhall/gascity#5872 — because the canonical copy cannot simply be deleted (it is load-bearing for the explicit `install_agent_hooks = ["codex"]` opt-in and for the doctor's PreCompact drift check; removing it fails 9 tests in that repo's own suite).

**Merge order matters, and this PR is the one that goes first.** This removes the *stale* copy. If the canonical copy in `gastownhall/gascity` were removed first instead, every machine carrying both would fall back to staging this stale variant — losing `--hook-format codex` on handoff and the `gc hook run --timeout` guard on the prompt-submit hooks. Merging this one first cannot cause that.

julianknutsen:
```fires each hook once from the user-level file ```

This is intended. If we remove the auto-installed one, anybody who doesn't have it in their user location won't get hooks at all. A better solution would be some sort of idempotent mechanism so they could exist in multiple places, but only fire once.

PR #395, opened 2026-09-04
fix(gastown): make prefix collision guidance proxy-safe

julianknutsen:
Updates Gastown mayor prompt to avoid unsupported gc bd rename-prefix in proxied mode and documents safe direct-mode migration prerequisites. Adds prompt inventory regression test.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -61,7 +61,7 @@ jobs:
                       pr-pipeline slack-channel slack-full slack-mini; do
             "$GC_BIN" lint "$pack"
           done
-          GC_TEST_BIN="$GC_BIN" python3 -m pytest tests/test_gc_role_prompt_integration.py -q
+          GC_TEST_BIN="$GC_BIN" python3 -m pytest tests/test_gc_role_prompt_integration.py tests/test_pack_composition_warnings.py -q
 
       # Installed, not parsed. `gc lint` above reads each pack on its own and
       # the per-pack suites read their own TOML; neither can say what gc does
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,16 +41,46 @@ jobs:
         run: python3 validate_registry.py --require-git
 
       - name: Run Python pack tests
-        run: python3 -m pytest tests contributing/tests gascity/tests discord/tests github/tests slack-full/tests slack-channel/tests pr-pipeline/tests profiler/tests -q
+        run: python3 -m pytest tests contributing/tests gascity/tests discord/tests github/tests oversight-rig/tests slack-full/tests slack-channel/tests pr-pipeline/tests profiler/tests -q
 
       - name: Lint and exercise shared role prompt composition
         run: |
           go install github.com/gastownhall/gascity/cmd/gc@latest
           GC_BIN="$(go env GOPATH)/bin/gc"
-          for pack in gascity gascity/roles bmad compound-engineering gstack superpowers profiler; do
+          # oversight-rig is deliberately absent. Under the gc this step
+          # installs it reports four bd-unknown-flag findings for `gc bd list
+          # --rig`, which is valid: `gc bd` strips --rig in extractBdScopeFlags
+          # before forwarding, and the scanner validated against bare bd's
+          # manifest alone. Same upstream defect as the `pre-5220` section of
+          # tests/gastown_lint_upstream_defects.txt, fixed by gascity 7724983de
+          # and not in v1.4.1. Add oversight-rig here in the same change that
+          # deletes that section. A dev gc already reports zero for it, so
+          # pinning them would waive findings most contributors never see.
+          # Refute: gc lint oversight-rig  (0 findings on main, 4 on v1.4.1)
+          for pack in gascity gascity/roles bmad compound-engineering gstack superpowers profiler \
+                      pr-pipeline slack-channel slack-full slack-mini; do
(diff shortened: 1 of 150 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277308207

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR replaces the Mayor prompt's one-line prefix-collision advice — which told the agent to run the Beads `rename-prefix` subcommand — with a runbook that actually works in proxied mode, and backs it with a regression test that inventories the bundled packs for refused subcommands. That matters because the old advice failed at precisely the worst moment: a prefix collision makes a rig's beads unroutable, and the remedy the Mayor was told to reach for is refused by the proxy, so the guidance dead-ended exactly when someone needed it. The new text is executable end to end — stop writes, record the affected rig IDs, fix the prefix in the owning city's config, and, only if IDs must be rewritten, take a verified backup and switch that scope to direct/server mode for the supported migration — and the new test keeps the packs from drifting back.

### Original PR Review

Decision: approve.

Specific gaps fixed:

- **The prohibition line tripped the packs' own scanner.** The first draft named the refused subcommand adjacent to `bd` inside the guidance text itself, so a mirror scan over the packs flagged the very sentence that was supposed to prevent the mistake. The wording now breaks that adjacency while keeping every semantic pin intact — the mirror scan is at zero hits at head.
- **The test asserted the fix instead of inventorying for it.** It originally checked that a specific string was present in one file. It was inverted into the acceptance criteria's inventory form: mirror the two scanner regexes per line across `gastown/` and `gascity/` (skipping comments, handling leading delimiters) and assert zero matches. It also deliberately stays at the repository root `tests/` rather than moving under `gastown/tests/`, where it would end up scanning itself.
- **The remediation pointed at a generated file.** The guidance told the reader to change `.beads/routes.jsonl`, which is rebuilt wholesale from configuration and would silently lose a hand edit. It now points at the authoritative surface — giving the rig a unique `prefix` in its `city.toml` entry — and describes `routes.jsonl` as regenerated from that. This was confirmed against the generating code and against a real, populated city config, and the whole-pack census now shows every mention of `routes.jsonl` in the packs is descriptive, with no instruction to hand-edit it surviving anywhere.

Review findings addressed:

- **Documentation accuracy of the test's scope** — the test's docstring explained why it scans two packs rather than all five; the reasoning was checked against the pack registry and corrected in `4c26685`.
- **Undocumented working-tree walk** — the test walks the working tree rather than tracked files, which is now stated as deliberate and grounded in `go:embed` semantics: the embed directive reads the filesystem at build time and never consults git, so an untracked pack file really would be embedded. Also fixed in `4c26685`, and it keeps the test runnable from a plain copy of the tree.

Non-blocking follow-ups (shared for visibility — **NOT** completed in this PR):

- `mc-adkh8` — **Test Evidence Quality, minor:** the new remedy pin checks that the first file token after the anchor phrase is `city.toml`, which is a positional check, so it stays green if someone *adds* a second, prescriptive "then hand-edit `routes.jsonl`" sentence alongside the correct one. A measured, purely additive assertion that pins the generated-artifact framing (rather than token position) is recorded on the bead, along with a note that two other tempting variants were measured and do not close the gap.
- `mc-ropg3` — **Release Safety, minor:** the downstream scanner carries a single tracked exception for this exact file and subcommand. Once this rewording flows downstream, that entry becomes unreachable dead config, and the ledger logs rather than asserts on matched exceptions, so nothing would flag it. The close-out is to delete the entry when the downstream pin next moves past this PR — deliberately out of this repository's diff.
- `mc-fy42v` — **Release hygiene, nit:** the downstream tracking note still cites the pre-adoption commit as having implemented the inventory test; it should cite the pushed head after finalize.
- `mc-n8u83` — **Maintainability, nit:** this test mirrors the upstream scanner construct-by-construct, but the upstream gate is not yet on the downstream default branch and this mirror deliberately carries no exception ledger of its own. Widening the scan or granting exceptions therefore has to be synced by hand. The direction is fail-safe (the mirror can only be stricter), so this is worth revisiting only when the upstream gate lands.

### Maintainer Changes

Maintainer fixes were pushed directly to the PR branch:

- `4c26685` — `fix(gastown): make the prefix-collision runbook proxy-safe and actionable` — 2 files changed, 207 insertions(+), 9 deletions(-). This carries the three review rounds' fixes: the non-adjacent prohibition wording, the inventory-form test, and the repointing of the remedy at `city.toml`.
- `25524dd` and `ec66824` — merge commits bringing the branch up to current `main`. They contribute no changes of their own; the branch's net diff against `main` is still exactly the two files this PR set out to change (2 files changed, 220 insertions(+), 1 deletion(-)).

Your original commit `f2fd383` is preserved in the branch history.

### Final Review Status

Ready for the merge queue: final head `ec66824` has passing required GitHub checks (CI and CodeQL both green), and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/gascity-packs/actions/runs/34211046762

### Review Iterations

3 review passes across a three-lane review, with 2 fix iterations. Rounds 1 and 2 requested changes (the scanner-tripping wording and its test shape, then the generated-file remedy); round 3 approved with no required changes, after independently re-deriving the fix claims against the generating source and a real city config rather than inheriting the previous round's verdict. The two remaining minors are recorded as follow-ups above and neither changes shipped behavior.

---

PR #411, opened 2026-09-09
fix(gastown): make every wisp resolution reach the wisps tier, and pin the pair

julianknutsen:
Post-merge review follow-up for #282 (landed range `d65abfa..5bffaf7`).

The post-merge review of the landed range found two actionable issues. Both are
fixed here in one commit.

## F1 — wisp resolution queries could not reach the wisps tier

Wisp roots are ephemeral, and `gc bd list` skips the wisps tier unless
`--include-infra` is passed. Ten executable `--type=molecule` resolution
queries lacked the flag, so they returned `[]` even with a wisp assigned:

- `gastown/formulas/mol-refinery-patrol.toml` (5 sites)
- `gastown/formulas/mol-deacon-patrol.toml` (1 site)
- `gastown/agents/refinery/prompt.template.md` (2 sites)
- `gastown/agents/deacon/prompt.template.md` (2 sites)

The consequences were role-specific. Under a provider that does not export
`$GC_BEAD_ID` the refinery had no working resolver at all, leaking one
`in_progress` wisp per affected cycle; the deacon prompt's
`[ -n "$CURRENT_WISP" ] && [ -z "$ASSIGNED_WISP" ]` guard poured a duplicate
whenever the blind query read empty. The witness already carried the flag, so
this brings the rest of the pack to the contract the witness already had.

The flag is deliberately **not** swept onto every `gc bd list`.
`mol-refinery-patrol.toml`'s find-work query is blind on purpose — its
correctness depends on staying blind — and the pack's other unflagged
occurrence is prose describing a closed-wisp lookup, not a resolution.

## F2 — the discriminator pin never measured what it claimed

The pin added in #282 compared forms only, while its failure message asserted
the resolution was "backed by the live assignee query". Deleting the entire
fallback query block at all nine non-witness sites left the suite green, so
nothing pinned that any refinery or deacon surface could resolve a wisp at all.

Replaced with a positive pin on the resolution **pair**: each
`CURRENT_WISP=${GC_BEAD_ID:-}` site must be followed within three lines by the
live `--include-infra` assignee query. Counting bare `gc bd list` occurrences
would have reproduced the same vacuity one layer down, so the pin matches the
full query literal.

`test_witness_wisp_queries_pin_include_infra` is widened to the whole census
and renamed `test_wisp_queries_pin_include_infra`. The pair pin reaches eight
of the ten newly-flagged sites — it scans only the four files that resolve into
`CURRENT_WISP` — so `mol-deacon-patrol.toml`'s query and the deacon prompt's
`ASSIGNED_WISP` query would otherwise be pinned by nothing. The census selects
on `--assignee=`, a property of the class rather than an incidental token,
which keeps the deliberately-blind prose line out without narrowing on `--json`
or `--limit`; narrowing on those would let an unflagged new site written
without them pass silently.

## Census header correction

The header said "Five role-surfaces" where no reading of the census yields
five. It now says six, enumerates the three prompt templates individually, and
drops the "correct today by form" exemption rationale, which was false while
the query half was blind. The exemption rests on the pair: a provider need not
export `$GC_BEAD_ID` at all, so the bare half can resolve nothing and the query
is the load-bearing half.

## Test evidence

Full gastown shell suite green (5/5 scripts, rc=0), and
`test_no_bare_bd_commands.py`, `test_gascity_pack_inference_gate.py`,
`test_ci_workflow.py`, `test_no_gc_sling_wrapper.py` — 68 passed.

The new pins were mutation-probed in a hermetic lab, 12 arms, all red with the
intended pin firing and the unmutated tree green before each:

- fallback query deleted at one site, and at all non-witness sites
- `--include-infra` stripped at the refinery formula, the deacon formula, the
  deacon `ASSIGNED_WISP` query, and the witness prompt
- a novel unpaired bare resolution site added
- a whole query site deleted (caught by the census floor, which the ratio alone
  cannot catch)
- a bare resolution converted to the spawn-trigger form
- the query pushed outside the pairing window

The causal control for F2: the "delete all non-witness fallbacks" arm is
**green** on pre-fix `main` and **red** here, reproducing the reported vacuity
and confirming the new pin is the delta. Non-vacuity was measured directly at
this head — refinery formula 5/5, refinery prompt 2/2, witness prompt 1/1,
deacon prompt 1/1, census 15/15.

The inference gate needed no repin: its `mol-refinery-patrol` tuple pins the
find-work query, which this change does not touch.

## Deferred

These were raised as non-gating and are deliberately not in this PR: splitting
the trigger-reference budget into separate code and prose checks; excluding
comment lines from the negative trigger pins; capturing `gc hook current`'s
exit status and stating the inherited-`$GC_BEAD_ID` assumption; parsing all
pack formulas in the suite.

## Known local-only failure

`tests/test_gastown_lint_findings.py::test_every_universal_waiver_entry_is_still_reported`
fails in this environment. It fails identically on a pristine `git archive` of
`origin/main` with the same finding sets, so it is dev-binary divergence and
not caused by this change.

### DIFF
--- a/gastown/agents/deacon/prompt.template.md
+++ b/gastown/agents/deacon/prompt.template.md
@@ -89,9 +89,9 @@ If `next-iteration` already ran, do not pour again; run `gc hook`.
 ```bash
 CURRENT_WISP=${GC_BEAD_ID:-}
 if [ -z "$CURRENT_WISP" ]; then
-  CURRENT_WISP=$(gc bd list --assignee="$GC_AGENT" --status=in_progress --type=molecule --limit=1 --json | jq -r '.[0].id // empty')
+  CURRENT_WISP=$(gc bd list --assignee="$GC_AGENT" --status=in_progress --type=molecule --include-infra --limit=1 --json | jq -r '.[0].id // empty')
 fi
-ASSIGNED_WISP=$(gc bd list --assignee="$GC_AGENT" --status=open --type=molecule --limit=1 --json | jq -r '.[0].id // empty')
+ASSIGNED_WISP=$(gc bd list --assignee="$GC_AGENT" --status=open --type=molecule --include-infra --limit=1 --json | jq -r '.[0].id // empty')
 if [ -n "$CURRENT_WISP" ] && [ -z "$ASSIGNED_WISP" ]; then
   NEXT=$(gc bd mol wisp mol-deacon-patrol --root-only --var binding_prefix={{ .BindingPrefix }} --json | jq -r '.new_epic_id // empty')
   if [ -z "$NEXT" ]; then
--- a/gastown/agents/refinery/prompt.template.md
+++ b/gastown/agents/refinery/prompt.template.md
@@ -69,7 +69,7 @@ external observers (witness, mayor) only catch on a slow patrol cycle.
 ```bash
 CURRENT_WISP=${GC_BEAD_ID:-}
 if [ -z "$CURRENT_WISP" ]; then
-  CURRENT_WISP=$(gc bd list --assignee="$GC_AGENT" --status=in_progress --type=molecule --limit=1 --json | jq -r '.[0].id // empty')
+  CURRENT_WISP=$(gc bd list --assignee="$GC_AGENT" --status=in_progress --type=molecule --include-infra --limit=1 --json | jq -r '.[0].id // empty')
 fi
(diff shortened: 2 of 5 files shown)


boylec:
Works for us. We hit #252 in production, and this PR covers the part that's still open on main.

- It applies cleanly to today's main (05031f2). With it applied, no `gc bd list --type=molecule` lookup in `gastown/agents` or `gastown/formulas` lacks `--include-infra`. That's 11 lines across the deacon and refinery prompts and patrol formulas.
- One leftover, prose only: `mol-refinery-patrol.toml` (around line 1194 on main) still tells the reader to find the wisp with `gc bd list --type=molecule`, without the flag.
- We run the formula half as a local override until this merges.

(Posted by my Gas City agent.)

PR #451, opened 2026-09-18
fix(gascity): accept live session claim identity

jamelt:
## Problem

`gc gc claim` currently chooses `BEADS_ACTOR` before `GC_SESSION_ID` when verifying the claimed bead. Pool workers export `BEADS_ACTOR` as the reusable pool name, while `gc hook --claim` assigns the bead to the live `ci-*` session ID. The hook succeeds, then the wrapper rejects its own claim with `CLAIM_REJECTED assignee mismatch`, leaving work in progress until an operator manually reruns the raw hook.

This reproduced on live Fab Pack attempts `fp-07e2fn` and `fp-jrkiyg`.

## Fix

Accept either authoritative identity exported for the worker:

- the live `GC_SESSION_ID`; or
- the configured actor fallback used by legacy/explicit-run environments.

Unrelated assignees remain rejected. A regression test covers the exact pool-name/live-session split and the rejection case.

## Verification

- `python3 -m unittest` focused claim command cases: 6 passed
- `uv run --with PyYAML --with pytest python -m pytest gascity/tests/test_formula_assets.py -q`: 95 passed, 5,966 subtests passed
- `uv run --with PyYAML --with jsonschema python validate_registry.py --require-git`: passed
- `go test .`: passed

The broader Python suite has one unrelated local failure because the installed development `gc` no longer reports five Gastown named-session lint findings that the pinned waiver fixture expects.


jamelt:
Superseded by upstream commit 9f98ea4 / PR #360, which fixes the same live-session versus pool-actor claim verification defect with the broader canonical identity set. Closing this duplicate; the remaining action is to update the deployed city pack pin.

